PR #2, opened 2026-08-24
Make the app deployable to GitHub Pages and the top chrome responsive

achma-learning:
The fork could not be published as a static site and the wordmark overran narrow viewports. Both are fixed here.

Responsive top row

  #title-bar and #style-indicator were the only fixed top-row chrome with no
  fluid sizing: 26px type at 8px tracking, a 52px logo and hard 32/36px insets.
  At 1440px that is 395px wide, so from ~721px down the wordmark ran straight
  through the centred action buttons and off the right edge. A single hard
  step-down existed at 720px, which left the 721-1090px range uncovered and
  still collided under ~500px.

  Type, tracking, logo, gap and insets are now clamp()-scaled, so the full-size
  design is reached at ~1090px and degrades continuously below it — 1440px
  renders byte-identical to before. Under 640px the action row un-centres to
  the right rail and the style readout stacks beneath it, which is the only
  way three items fit one row on a phone. The now-redundant 720px overrides
  are removed; leaving them would freeze the fluid scale.

  .hud-summary-wrap had the same problem: a flat 420px cap overran a 390px
  viewport by ~70px, and since it is the widest child it dragged the whole
  intel readout with it. Capped against the viewport too.

  Verified 320-1440px: no horizontal overflow and no overlapping chrome.

Static hosting

  Every live feed is brokered by configureServer middleware in vite.config.js,
  so it runs under `vite dev` only. A static host cannot run it, and /api/*
  does not exist there. That is a property of static hosting, not a bug — and
  the globe, styles, share links, bundled datasets and the CORS-open feeds
  (USGS, GBFS) never needed a backend at all.

  - Builds use a relative base, so one artifact is valid at any URL prefix. This also avoids vite-plugin-cesium copying its runtime to dist/<base>/ when a non-root absolute base is set. Dev keeps '/'.
  - Runtime-assembled asset paths (aircraft GLBs, logo, mic) resolve against document.baseURI via a new appAssetUrl helper. The source strings stay put since several are identity keys pinned by tests.
  - A missing GOOGLE_MAPS_API_KEY no longer aborts init. MapStackController already fell back to the keyless OSM stack; the app now boots on it and explains itself instead of showing an error on a black screen.
  - Static builds answer their own /api calls with an explanatory 503, so each layer's existing error surface reads "needs the local server" rather than choking on the host's 404 HTML. GEV_STATIC_BUILD=0 opts out.
  - A one-time notice states which half of the app is live and accepts the two browser-side keys into localStorage, so a keyless deployment is still useful — visitors can point it at their own quota without a rebuild.

Adds the Pages workflow (keys optional, via repo secrets), docs/GITHUB-PAGES.md and README/.env.example pointers. Full suite passes (2587).

Claude-Session: https://claude.ai/code/session_01FhXFJjdiwFFMcdou5QbeAQ

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -60,6 +60,15 @@ OPENSKY_CLIENT_SECRET=
 # Optional: OpenSky credentials JSON file path (alternative to above)
 # OPENSKY_CREDENTIALS_FILE=/path/to/credentials.json
 
+# Optional: static-hosting flag. A BUILD has no /api middleware — every proxy
+# is registered through Vite's configureServer hook, so it runs under
+# `vite dev` only (`vite preview` has none either). Builds therefore default to
+# static mode: /api calls are answered locally with an explanatory 503 and the
+# app shows a one-time notice about which layers need the dev server.
+# Set to 0 ONLY when serving dist/ behind something that actually implements
+# the same /api surface. Ignored by `npm run dev`. See docs/GITHUB-PAGES.md.
+# GEV_STATIC_BUILD=0
+
 # Optional: dev server config
 PORT=4173
 # Network binding. Default is localhost — the server is reachable ONLY from
--- a/.github/workflows/deploy-pages.yml
+++ b/.github/workflows/deploy-pages.yml
@@ -1,0 +1,90 @@
+name: Deploy to GitHub Pages
+
+# Publishes the static build to GitHub Pages.
+#
+# What you get: the 3D globe, every visual style, the camera/scene director,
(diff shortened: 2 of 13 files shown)


bilawalsidhu:
GitHub Pages can't host the app because every layer goes through the server proxies, so that half can't land. The responsive top-row chrome is worth having on its own; open it as a separate PR and I'll review it. Closing this one.

PR #11, opened 2026-08-25
Issue 4 wigle wifi

Tom-Neverwinter:


bilawalsidhu:

Tom-Neverwinter:
Rebuilt on current main as a single commit. Independent of the tap-contract work in #382 — WiGLE is a fixed public API behind a token, not a user-addressed local device, so it needs none of that.

`vite.config.js` went from 7,995 lines to 3 while this sat open, so the proxy is re-authored as `server/providers/wigle.js` registered in `local.js`. It was also stacked on the OpenSeaMap branch; that's untangled and this is one layer on its own.

## Licensing — the part I'd want you to decide on

You confirm data terms yourself before a source lands (#256), so here's the text rather than my summary. WiGLE's [data licence](https://wigle.net/eula.html):

> **LICENSE GRANT.** …a non-exclusive right to use the maps and access point database (hereinafter, "the Data") **on a single computer (i.e., with a single CPU)**. You, or another person with your express permission and supervision, may also use the Data on any computer system owned by you at a single location provided you can: (1) **guarantee that there is no possibility of its use by more than one person at a time**; and, (2) that you do not receive any remuneration for the other person's use of the Data.

> **RESTRICTIONS.** You may not rent, lease, sublicense, transfer, **distribute, telecommunicate, copy**, or sell the Data in its entirety or in any part **for any commercial consideration, gain, or benefit**.

WiGLE's FAQ adds that money-making use requires a separate commercial licence with its own unthrottled tokens.

For the local open-source app this is fine, and it's what the layer is built for: each user brings their own account, so each user is the licensee under their own acceptance, and nothing is redistributed — no WiGLE data is committed, it's fetched at runtime and cached only on the machine that queried it under git-ignored `.gev-cache/`.

**Where it bites is the hosted version.** README line 502 says one is coming at Halfpixel. Single-computer, one-person-at-a-time, and no distribution for commercial benefit all point the same way: **this layer needs a commercial licence from WiGLE before it can ship in a hosted product, or it has to stay disabled there.** I've written that into DATA_SOURCES.md in the same shape as the Google News restriction already on the regional briefing, since that's the established precedent in that file. Happy to reword if you read the clauses differently — I'd rather you saw the quotes than my paraphrase.

## Hardening beyond the port

Re-reading the original turned up three things worth fixing:

- **Upstream message text was being relayed to the browser.** WiGLE answers HTTP 200 with `{success:false}` for several real failures, so status alone can't separate them. `wigleClientError()` now maps status to our own wording and keeps the three distinctions that change what the user should actually do — credentials rejected / allowance exhausted / everything else — and the upstream message is logged locally only. There's a test asserting a planted upstream string never reaches the client.
- **No rate limit, timeout, or body cap.** Now 60/min per client (300 global), 15s timeout, 4 MB cap, `redirect: 'manual'`.
- **A non-JSON body threw inside the handler.** Now a clean 502.

## Spending the allowance slowly

The per-account daily allowance is small, and that shapes everything: 0.4° viewport cap, 24h memory+disk cache keyed on the rounded bbox (access points don't move), in-flight coalescing so a burst of pans costs one query, and a soft daily counter defaulting to 50.

The counter reuses `normalizeBudget`/`isOverBudget` from `tomtomTiles.js` rather than a second hand-rolled copy, but rolls on `wiglePacificDayKey()` — WiGLE resets at US/Pacific midnight, not UTC. It's in-memory and resets on server restart; TomTom's persisted `budget.json` is the upgrade path if that proves insufficient.

## Fixture

Tests run offline against `src/data/fixtures/wigle-network-search.json`. **It's synthetic and labelled as such — deliberately containing no real observed network.** Real rows are other people's crowdsourced observations under the licence above, so committing any would be exactly the redistribution DATA_SOURCES says we don't do. It covers what the normalizer must handle: a current network, an aged one, blank `ssid`/`channel` (`Number('')` is `0`, so blank has to read as missing, not zero), and an untriangulated `0,0` row that must be dropped rather than plotted in the Gulf of Guinea.

## Verification

`npm test` 2990 pass / 0 fail with `GEV_REQUIRE_ALLOCATION_GATE=1` on Node 24.19.0, `npm run build` clean, `npm run test:track` 108 passed / 0 failed, `format:check` and `check:boundaries` clean. 11 new proxy tests plus the existing pure-helper tests.

`DATA_SOURCES.md`, `docs/CURRENT-STATE.md`, `SECURITY.md` and `.env.example` are updated in the same commit.

> 🤖 Generated with [Claude Code](https://claude.com/claude-code)
>
> https://claude.ai/code/session_01KGKBNNjPh4iQe8sviGZNPb

PR #13, opened 2026-08-25
Add NIFC wildfire incidents layer

Tom-Neverwinter:
New layer showing confirmed wildfire incidents from NIFC's public WFIGS ArcGIS Feature Service — deliberately distinct from the existing FIRMS layer, which shows raw VIIRS satellite thermal-anomaly detections. NIFC data is human-reported incident tracking (name, cause, acreage, percent contained), same complementary relationship as ALPR sits next to CCTV.

Public, CORS-open, keyless ArcGIS endpoint — fetched directly from the browser exactly like earthquakes.js fetches USGS, no server proxy needed. Verified live: real endpoint returns 602 current incidents matching the exact query/fields the layer uses. Scoped to IncidentTypeCategory IN ('WF','CX') — wildfires and complexes; prescribed burns (RX) are excluded at the query level as out of scope for a wildfire-incidents layer.

Marker size scales with acreage (sqrt, clamped); color reflects containment (uncontained/partial/contained, unset reads as uncontained). Click shows cause, discovery date, acreage, percent contained, state.

Registered in main.js + layerState.js (token 'v' — 'z' was already a hardcoded "unknown token" test sentinel in two test files; caught by the full suite before committing), voice/text aliases, attribution in dataCredits.js and DATA_SOURCES.md.

Claude-Session: https://claude.ai/code/session_01GZFzxaNKiGGkq9V64Y9omb


bilawalsidhu:
On hold until the NIFC data terms are confirmed. Full status for your open PRs: https://github.com/bilawalsidhu/gods-eye-view/pull/171#issuecomment-5612644797

PR #14, opened 2026-08-25
Add public ALPR / Flock Safety camera layer

### LINKED ISSUE #5: [Feature Request] Public ALPR / Flock Safety Camera Layer
Summary

Add a layer showing publicly mapped automated license plate reader (ALPR) cameras, including Flock Safety cameras.

A good starting source appears to be DeFlock / OpenStreetMap, which already maintains crowdsourced locations of ALPR infrastructure.

This would fit very naturally beside the existing CCTV and infrastructure layers.

Suggested behavior
New ALPR Cameras layer.
Optional filter for:
All ALPR cameras
Flock Safety
Other manufacturers
Use publicly documented/crowdsourced camera locations rather than private Flock systems.
Possible sources:
DeFlock data
OpenStreetMap / Overpass
DeFlock PMTiles if appropriate
Cluster cameras at wide zoom levels.
Render individual cameras when zoomed in.
If direction information exists, render a small camera-facing cone/heading indicator.
Clicking a camera could show:
Manufacturer
Camera/system type
Operator, when publicly documented
Direction
Source
Last verification/update
Clearly distinguish:
confirmed/tagged Flock cameras
generic ALPR cameras
uncertain/crowdsourced reports
Label this as mapped surveillance infrastructure, not a live camera feed.
Preserve OpenStreetMap/DeFlock provenance and required attribution.
Scope / privacy boundary

This request is only for the physical locations and public metadata of surveillance infrastructure.

It would not involve:

license plate records
vehicle histories
Flock customer accounts
private Flock APIs
camera imagery unavailable to the public

The goal is simply to visualize publicly documented surveillance infrastructure as another OSINT layer.

Nice-to-have

Because God's Eye View already has a click-to-track/contact concept, it could eventually support things such as:

"Show ALPR cameras within 10 km"

or

"Show surveillance infrastructure near this location."

That would make the layer especially useful without needing access to any private surveillance data.

Tom-Neverwinter:
Closes #5.

New optional layer showing crowdsourced ALPR camera locations from OpenStreetMap's DeFlock tagging (`man_made=surveillance` + `surveillance:type=ALPR`), fetched viewport-bounded through the existing generic `/api/overpass` proxy — no new server route needed.

- `src/data/alprCameras.js`: viewport-bounded fetch/render, patterned on `militaryInstallations.js` (moveEnd-debounced, unavailable-retry backoff, click-to-select entity context). Flock-operated cameras render in a distinct color from other tagged ALPR vendors; a `camera:direction`/`direction` tag draws a short facing-direction line.
- Registered in `main.js` + `layerState.js` for the auto-built toggle panel and share-link serialization.
- Voice/text aliases in `gevActions.js`.
- Attribution in `dataCredits.js` and `DATA_SOURCES.md`.
- `src/data/alprCameras.test.mjs`: unit tests for the pure helpers (tag normalization, Flock classification, destination-point math, retry backoff).

Mapped surveillance infrastructure only — no plate records, vendor accounts, or private Flock/OSM data.

Includes a follow-up commit fixing a bug a code-review pass found: `numTag()` treated a present-but-blank OSM tag value as a fabricated `0` instead of missing (`Number('') === 0`), which would have rendered a camera with no real facing data as pointing due north.


d8ahazard:
+1 - This was the first thing I want to see if it supported, and if not, add it.

bilawalsidhu:
Reviewed the diff — this is a quality contribution: registry + credits + DATA_SOURCES row, backoff, saturation reporting at the Overpass cap, abort-on-supersede, tests on the pure helpers. Thank you. I want this as the base for the ALPR layer.

Two things before merge:

1. Strip the `ponytail:` markers from the code comments (they show up across your PRs — looks like a tooling artifact leaking through).
2. Fold in bbox quantization along the lines of #63 — snapping fetch boxes to a coarse grid so repeated pans hit Overpass's query cache instead of 502ing on metro-wide boxes. That insight is @d8ahazard's; credit it in the commit.

Sequencing heads-up: a release lands this week (zero-key boot + in-app provider settings), and I'm speccing a country-scale aggregate view + context-mode integration to sit on top of this layer afterwards. Your PR stays scoped exactly where it is — viewport-level detail — which is the right v1.

bilawalsidhu:
Three more concrete items from a deeper pass on the data (all verified against live Overpass/taginfo today), to fold into the fix list:

1. **The exact-match filter misses a variant.** `["surveillance:type"="ALPR"]` skips the `camera;ALPR` semicolon-combo tagging (small but real — 61 nodes globally). Match it with a regex clause or a second selector.
2. **The 600-node cap saturates on a single city.** A half-degree box over Austin alone returns **669** ALPR nodes — above both `QUERY_LIMIT` (600) and `MAX_RENDERED` (500) — so a metro view silently truncates today. The saturation *reporting* you built is right; the ceilings need to rise (or density-decimate) so one city fits.
3. **The "336K+ worldwide" figure in the `DATA_SOURCES.md` row is wrong.** taginfo (data through 2026-08-28) puts `surveillance:type=ALPR` at **147,207** total. The 336K number appears to count `man_made=surveillance` broadly. Quote "~147K mapped worldwide (Aug 2026)" or, better, don't hardcode a number that ages.

Also for the record since I'm asking for changes: your `numTag('') === 0` fix — refusing to render a tagless camera as due north — is exactly right and should survive any refactor. `direction` coverage is 93.6% in the data, so the cones are the layer's best feature.

bilawalsidhu:
This PR is the base of the ALPR layer about to ship, with your commits and authorship intact. Full status for your open PRs: https://github.com/bilawalsidhu/gods-eye-view/pull/171#issuecomment-5612644797

Tom-Neverwinter:
All five items from your 2026-08-29 review are in, pushed as `40a9667`. Also rebased on main — `src/main.js` became the thin bootstrap since this branch was opened, so registration moved to `src/standalone/data.js`. PR is mergeable again.

**1. `ponytail:` markers.** Gone. You were right that it was a tooling artifact leaking through; I've stripped them and they won't be in future PRs.

**2. bbox quantization (#63, @d8ahazard).** Fetch boxes now snap to a 0.125° grid via `quantizeBox()`, credited to @d8ahazard in both the code and the commit. Three details worth flagging since they're judgement calls:

- **Snaps outward** (floor the low edges, ceil the high edges) rather than to nearest, so the fetch box always contains the viewport. Rounding to nearest would sometimes fetch a box smaller than what's on screen and leave a visible margin unpopulated.
- **Snaps after the viewport-size gate**, so the "zoom in" prompt still reflects the real viewport rather than the padded fetch box.
- **Panning inside a cell now skips the request entirely**, not just reuses the cache key. That's the bigger win, but it makes staleness a real risk, so the cached cell is cleared on a failed fetch and on `disable()` — otherwise an error or a disable/re-enable could serve a stale cell. `destroy()` routes through `disable()`.

I used `0.125` rather than #63's `0.14`: a power-of-two fraction is exactly representable, so grid lines never drift.

I did **not** bring over the bundled Lannon/US-45 snapshot from #63, per your note that it can't ship.

**3. `camera;ALPR` variant.** `surveillance:type` is multi-value, so the exact match was dropping the combos. Now anchored on the value separators — `~"(^|;)ALPR(;|$)"` — so `ALPR`, `camera;ALPR` and `ALPR;camera` all match, but a hypothetical `NOTALPR` wouldn't.

**4. Ceilings.** `QUERY_LIMIT` 600 → 2500, `MAX_RENDERED` 500 → 2000, so a dense metro fits instead of truncating before the saturation warning can fire. Austin's ~669 now has real headroom.

**5. The 336K figure.** Took your "better" option and dropped the number instead of swapping it for 147K — the tag is actively being mapped, so any figure baked into a comment ages the same way. Removed from both `alprCameras.js` and the `DATA_SOURCES.md` row; the row now describes the quantization and the new cap instead.

**Verification.** `npm test` 2982 pass / 0 fail, `npm run build` clean, `npm run test:track` 108 passed / 0 failed, `format:check` and `check:boundaries` clean. Four new tests on the pure helpers cover outward snapping, cell reuse across a small pan, pole/antimeridian clamping, and non-finite rejection.

One note: I ran the suite on Node 22.19.0, which skips the two allocation microbenchmarks (they're calibrated for 24). CI will exercise those.

> 🤖 Claude Code helped

samehkhamis:
Thank you, @Tom-Neverwinter, for the contribution! The adapted ALPR camera layer is now merged in #434, preserving your authorship and Bilawal’s contribution credit.

Issue #44, opened 2026-08-25
test:track: ground-3d probe expects 4 sampleHeight calls but only makes 3 (fails on clean main)


prajwal-raj:
Running npm run test:track on a fresh clone fails one check, and it isn't tied
to any local change. It reproduces on a clean main (release commit 880a672).

The check that fails:

  ground-3d: ground snap + mesh-floor probes are one-shot/per-poll bounded
  (no per-frame sampling). sampleHeight is called 3 times after the models
  come up; the check wants at least 4 (a snap plus a mesh cell for each of the
  two grounded planes). The per-frame guard itself is fine: growth over ~60
  frames is 1.

So it's 99 passed, 1 failed with a valid Google Maps key, and it isn't a
browser artifact: it reproduces identically on the pinned Chrome-for-Testing
145.0.7632.77 the harness targets, and on a newer system Chrome (151). Without
a key you also get three "no console errors" failures from the tiles 400, but
those clear once a real key is in place.

A few things I noticed while digging in:
- It's consistent, not flaky: always 3, on both the pinned v145 and system v151.
- Only the initial count is short. The "no per-frame sampling" half passes.
- The terrain upstream (terrain.reearth.land) is up and quick for single
  points (200 in about 2s), so it isn't an outage. On a cold cache, though, a
  batch can run past the 30s window on first load.

Environment: macOS arm64, Node 26.4.0, pinned Chrome-for-Testing 145.0.7632.77.

Is the "at least 4" right for two grounded planes, or can one of them
legitimately produce a single sampleHeight call depending on timing? Happy to
help chase it if it's a real gap.

prajwal-raj:
One more piece of evidence: the dev-server log during the failing run shows the cold-cache terrain path timing out, which lines up with the short sampleHeight count.

`[terrain-heights-proxy] refresh incomplete (The operation was aborted due to timeout)`

onionviolet:
[...]

PR #45, opened 2026-08-25
fix: measure GBFS response cap in bytes, not string length

### LINKED ISSUE #32: Use byte counts, not JS string length, for GBFS body limits

Problem:
- `GBFS_MAX_BODY_BYTES` is a byte limit, but the proxy compares it to JavaScript string length after decoding.
- Character count is not equivalent to byte count for non-ASCII payloads.

Evidence:
- `vite.config.js:3317-3330` calls `await upstream.text()` and compares `body.length` to `GBFS_MAX_BODY_BYTES`.

Suggested fix:
- Enforce the byte cap before text decoding.
- Use `Buffer.byteLength` only after a bounded read if a full string is still required.
- Rename variables so byte limits and character limits cannot be confused.

prajwal-raj:
The GBFS proxy is supposed to cap upstream responses at 5 MB, but it was
checking body.length after decoding. That's a character count, not bytes, so
for multi-byte (non-ASCII) responses a body could go over the byte limit and
still pass the check.

Fix is one line: compare Buffer.byteLength(body, 'utf8') against the cap
instead. ASCII traffic under the cap behaves exactly as before; this only
tightens the edge case for multi-byte bodies.

Closes #32

### Testing
- npm run test:track: 99/100 on the pinned Chrome-for-Testing v145. The one
  failure (a ground-3d probe that expects 4 sampleHeight calls but sees 3)
  isn't from this change. It reproduces the same way on a clean main checkout
  and lives in the terrain/ground-snap code, which this PR doesn't touch.
  I've opened #44 for it.
- Checked the actual path by hand too: hitting /api/gbfs with Austin's
  station_information.json still returns 200 and valid JSON.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,10 @@ of current runtime behavior, see [`docs/CURRENT-STATE.md`](docs/CURRENT-STATE.md
 
 ### Security
 
+- GBFS proxy body-size cap now measures the response in bytes
+  (`Buffer.byteLength`) instead of JavaScript string length, so the
+  `GBFS_MAX_BODY_BYTES` limit holds for multi-byte payloads and cannot be
+  overrun by non-ASCII upstream responses.
 - Production transitive dependencies resolve to patched DOMPurify and
   protobufjs releases without changing the Cesium version or application APIs.
 - Production dependency audit reports no known advisories; remaining audit
--- a/vite.config.js
+++ b/vite.config.js
@@ -3323,7 +3323,7 @@ function gbfsProxy() {
             return;
           }
           const body = await upstream.text();
-          if (body.length > GBFS_MAX_BODY_BYTES) {
+          if (Buffer.byteLength(body, 'utf8') > GBFS_MAX_BODY_BYTES) {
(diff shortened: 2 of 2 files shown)


prajwal-raj:
Hi @bilawalsidhu, thanks for open-sourcing this. This is a small one-line fix for the GBFS byte-cap check (#32) whenever you have a moment, no rush. I also opened #44 separately for an unrelated test:track failure I ran into on a clean checkout.

bilawalsidhu:
Merged — thank you. What shipped: the GBFS proxy's response-size cap now counts actual UTF-8 bytes rather than JavaScript string length, so `GBFS_MAX_BODY_BYTES` genuinely bounds multi-byte payloads instead of undercounting them. Small diff, real correctness — exactly the kind of fix that's a pleasure to take. Ships in the next release.

bilawalsidhu:
@prajwal-raj — closing the loop: your GBFS byte-limit fix is now included in [v0.1.1](https://github.com/bilawalsidhu/gods-eye-view/releases/tag/v0.1.1). The proxy now correctly bounds multi-byte payloads, with your contributor credit preserved on `main`. Thanks again for the careful catch.

Issue #19, opened 2026-08-25
Protect /api/google/text-search from quota spend and invalid coordinates


jsawyerdev:
Reviewed commit: 880a672b5e16ad3e41d318801d3a5203f9201923

Problem:
- `/api/google/text-search` proxies Google Places text search with no auth and optional-only throttling.
- It only checks that `lat` and `lon` are finite, not that they are valid latitude/longitude ranges.

Evidence:
- `vite.config.js:5390-5499` implements the proxy.
- `vite.config.js:5398-5407` applies only the optional limiter.
- `vite.config.js:5418-5423` validates presence/finite values but not coordinate bounds.

Suggested fix:
- Require auth and mandatory rate limits for exposed use.
- Validate latitude in `[-90, 90]` and longitude in `[-180, 180]` before forwarding upstream.
- Fail before contacting Google on invalid input.

This was found during an adversarial repository review; impact depends on how the dev/preview server is exposed.

bilawalsidhu:
Thanks for the report. Consolidating this into #16; its remaining requirements are preserved there. Current main validates WGS84 bounds before Google text search; opt-in quota policy remains. This closure does not mean the remaining work is fixed.

Issue #18, opened 2026-08-25
Protect /api/google/nearby-places from unauthenticated quota spend


jsawyerdev:
Reviewed commit: 880a672b5e16ad3e41d318801d3a5203f9201923

Problem:
- `/api/google/nearby-places` proxies Google Places requests with the server Google key.
- The route is unauthenticated and the limiter defaults to unlimited.

Evidence:
- `vite.config.js:5274-5384` implements the nearby places proxy.
- `vite.config.js:5284-5293` applies only the optional limiter.
- `vite.config.js:5296` reads `GOOGLE_MAPS_API_KEY`.
- `.env.example:13-22` documents Google proxy rate limiting as optional.

Suggested fix:
- Require auth or restrict the route to trusted local use.
- Make request budgets mandatory before enabling the proxy on a reachable interface.
- Add telemetry for rejected and accepted quota-bearing requests.

This was found during an adversarial repository review; impact depends on how the dev/preview server is exposed.

bilawalsidhu:
Thanks for the report. Consolidating this into #16; its remaining requirements are preserved there. Coordinate validation exists, but the limiter remains opt-in and the exposure policy is shared. This closure does not mean the remaining work is fixed.

Issue #17, opened 2026-08-25
Require auth and hard rate limits for /api/openai/hud-summary


jsawyerdev:
Reviewed commit: 880a672b5e16ad3e41d318801d3a5203f9201923

Problem:
- `/api/openai/hud-summary` sends requests to OpenAI Responses using the server key.
- The route is unauthenticated and the limiter defaults to unlimited.

Evidence:
- `vite.config.js:4960-5020` implements the route and calls OpenAI.
- `vite.config.js:4970-4971` only applies the opt-in limiter.
- `vite.config.js:4973` reads `OPENAI_API_KEY` server-side.

Suggested fix:
- Require authentication/authorization for cost-bearing OpenAI calls.
- Make server-side quota controls mandatory for any non-local bind.
- Return a hard configuration error instead of silently running unlimited.

This was found during an adversarial repository review; impact depends on how the dev/preview server is exposed.

bilawalsidhu:
Thanks for the report. Consolidating this into #16; its remaining requirements are preserved there. Same OpenAI key-spend boundary and opt-in limiter; one policy issue should govern both routes. This closure does not mean the remaining work is fixed.

Issue #16, opened 2026-08-25
Require auth and hard rate limits for /api/realtime/token


jsawyerdev:
Reviewed commit: 880a672b5e16ad3e41d318801d3a5203f9201923

Problem:
- `/api/realtime/token` mints OpenAI realtime client secrets using the server `OPENAI_API_KEY`.
- The only limiter is opt-in; the default is unlimited.
- If the Vite dev or preview server is reachable by other clients, they can spend the project OpenAI quota.

Evidence:
- `vite.config.js:5045-5213` implements the endpoint and calls `/v1/realtime/client_secrets`.
- `vite.config.js:427-479` makes the limiter optional/unlimited by default.
- `.env.example:41-49` documents the OpenAI limiter as optional.

Suggested fix:
- Require an explicit auth boundary before minting a token.
- Make rate limiting default-on and fail closed when not configured for exposed deployments.
- Consider disabling this endpoint outside localhost unless the deployment has real authentication.

This was found during an adversarial repository review; impact depends on how the dev/preview server is exposed.

bilawalsidhu:
Consolidating #17, #18 and #19 here so the cost-bearing provider boundary has one owner. Preserve all remaining scope: OpenAI session/token and HUD-summary routes; Google geocoding and Places text-search routes; server-held-key spend; explicit local/LAN/remote access policy; safe defaults for rate/concurrency/quota controls; and tests for disabled, enabled and bypass/error paths. Current main fixes #19's coordinate-bounds validation, but that does not settle access control or spend policy. Preserve intentional local development/preview behavior while deciding non-local exposure.

PR #46, opened 2026-08-25
fix: reject out-of-range coordinates in Google text-search proxy

prajwal-raj:
/api/google/text-search validated that lat and lon were finite, but not that
they were actual coordinates. Anything finite got forwarded to Google Places,
so an out-of-range value spent quota on a request that could never return a
useful result.

This adds a bounds check right after the existing finite check: latitude has
to be within [-90, 90] and longitude within [-180, 180], otherwise the route
returns 400 before contacting Google. Valid coordinates are unaffected.

Scope note: issue #19 also mentions auth and rate limits for exposed use.
That is a broader concern shared across the other Google and OpenAI proxy
routes (#16, #17, #18), so I kept this PR to the coordinate-validation half,
which is the self-contained correctness fix. Using Refs rather than Closes so
the auth part can stay tracked.

Refs #19

### Testing
- npm run test:track: 99/100 on the pinned Chrome-for-Testing v145. The single
  failure is the pre-existing ground-3d probe tracked in #44, unrelated to this
  change and reproducible on a clean main.
- Verified by hand against a running dev server:
  - lat=999 and lon=-200 both return 400 with "lat must be within [-90, 90]
    and lon within [-180, 180]", with no upstream call made.
  - valid coordinates (lat=12.97, lon=77.59) pass the check and reach Google
    as before.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,10 @@ of current runtime behavior, see [`docs/CURRENT-STATE.md`](docs/CURRENT-STATE.md
 
 ### Security
 
+- The `/api/google/text-search` proxy now rejects out-of-range coordinates
+  (latitude outside [-90, 90] or longitude outside [-180, 180]) with a 400
+  before contacting Google, so invalid input can no longer reach the
+  quota-bearing Places request.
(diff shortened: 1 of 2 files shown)


bilawalsidhu:

Issue #26, opened 2026-08-25
Enforce streaming byte and duration caps for CCTV media responses


jsawyerdev:
Reviewed commit: 880a672b5e16ad3e41d318801d3a5203f9201923

Problem:
- CCTV media responses are only rejected when upstream declares an oversized `Content-Length`.
- Chunked or missing-length responses can stream indefinitely through the proxy.

Evidence:
- `vite.config.js:4320-4359` implements `proxyMediaResponse`.
- `vite.config.js:4335-4343` checks declared `Content-Length > 64 MB`.
- The subsequent stream piping has no accumulated byte cap, idle timeout, or duration cap.

Suggested fix:
- Count bytes while streaming and destroy the upstream/response once the cap is exceeded.
- Add idle and total-duration timeouts.
- Prefer a smaller default cap for image/video preview routes unless larger media is required.

This was found during an adversarial repository review; impact depends on how the dev/preview server is exposed.

bilawalsidhu:
Consolidating the remaining work from #25 here. Current main has a deadline for opening the upstream through response headers. It does not establish an idle-body deadline, total stream lifetime, or cancellation/backpressure contract. Preserve stalled-after-headers tests, downstream disconnect cleanup, timely live-stream behavior, and the intended duration/resource policy. The header fix is not proof that this issue is resolved; #581 excludes CCTV media.

ethanstoner:
Taking this one if nobody else is on it. Scoping to the idle-body deadline and the backpressure contract from your comment above, and leaving total stream lifetime out. The cap value reads like a policy call you should make rather than something I should guess, so I will raise it as an open question in the PR.

When pipe() pauses the source for backpressure, no bytes arrive from upstream, so a naive idle timer would tear down a healthy feed that is only being served to a slow viewer. I reschedule the deadline while the response is still waiting to drain, so a slow client cannot be mistaken for a stalled upstream.

The stalled-after-headers test, the downstream disconnect cleanup and unbounded live MJPEG/HLS all stay as they are.

PR #53, opened 2026-08-25
fix: add an upstream connect timeout to the CCTV media proxy

### LINKED ISSUE #25: Add upstream timeout handling to the CCTV media proxy

Problem:
- `/api/cctv/media/:id` fetches configured upstream media without an abort timeout.
- Slow or hanging upstreams can hold server resources indefinitely.

Evidence:
- `vite.config.js:4543-4611` implements the media proxy.
- `vite.config.js:4565-4567` calls `fetch(mediaUrl, { headers })` without an `AbortSignal` timeout.

Suggested fix:
- Add a bounded connection/read timeout with `AbortController`.
- Return a 504-style response for upstream timeout.
- Log source ID and timeout class without logging sensitive headers.

prajwal-raj:
The CCTV media proxy (/api/cctv/media/:id) fetched the configured upstream
with no timeout. A slow or hung source could keep a proxy request open
indefinitely and tie up resources.

This wraps the upstream fetch in an AbortController with a bounded connect
timeout (15s) and returns a 504 when the upstream stalls, instead of a
generic 502. The important detail for streaming: the timer is cleared as soon
as response headers arrive, so a healthy live video stream is never cut off.
It follows the same pattern the GBFS proxy already uses in this file.

Scope note: this covers the connect timeout the issue calls out (the media
fetch had no AbortSignal). Idle and total-duration limits while a stream is in
flight are a separate concern tracked in #26.

Closes #25

### Testing
- npm run test:track: 99/100 on the pinned Chrome-for-Testing v145. The one
  failure is the pre-existing ground-3d probe tracked in #44, unrelated to
  this change and reproducible on a clean main.
- Verified both paths by hand against a running dev server:
  - A healthy camera returns 200 image/jpeg in well under a second, so normal
    proxying is unaffected.
  - A source pointed at a deliberately hung upstream now returns
    504 "Upstream media timeout" right at the 15s bound, instead of hanging.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,10 @@ of current runtime behavior, see [`docs/CURRENT-STATE.md`](docs/CURRENT-STATE.md
 
 ### Security
 
+- The CCTV media proxy now applies a bounded connect timeout to the upstream
+  fetch and returns a 504 when an upstream stalls, so a slow or hung source can
+  no longer hold a proxy request open indefinitely. The timer is cleared once
+  response headers arrive, so healthy live streams are unaffected.
 - Production transitive dependencies resolve to patched DOMPurify and
   protobufjs releases without changing the Cesium version or application APIs.
 - Production dependency audit reports no known advisories; remaining audit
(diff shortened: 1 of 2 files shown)


bilawalsidhu:

PR #58, opened 2026-08-25
feat(data): add SLF Alpine measurement stations layer

darrendignam:
Adds the WSL/SLF IMIS network as a keyless CC BY 4.0 layer: ~207 automatic stations across the Swiss Alps, mostly above the tree line, reporting snow depth, wind, and temperature every 30 minutes.

Fetched through a new /api/slf/stations proxy because measurement-api.slf.ch sends no CORS headers, unlike USGS. The proxy also joins stations to measurements so the browser receives ~207 ready-to-render records rather than the ~9,400 raw rows upstream returns, and caches for five minutes against a 30-minute feed.

Snow depth is drawn as a vertical stem, following the dam-stem recipe in localGeojson.js: polyline and point on one entity that carries a position, explicit graphics constructors, anchored on the sampled ground floor rather than the station's published elevation, and scaled to camera altitude so it stays legible at any zoom while preserving the ratio between stations.

Units come from the published OpenAPI schema, not guessed. An absent reading stays null rather than becoming zero, because 0 cm is a real August measurement and the two must not be conflated.

Station detail renders as world-overlay cards; the viewer runs with infoBox: false, so Cesium entity descriptions would render nowhere.


bilawalsidhu:
Hi @darrendignam, thanks for this, and sorry it sat so long. SLF's high-altitude snow depth, wind and temperature is a source we'd like in God's Eye View, and your `slfStations.js` normaliser is careful work. Main has moved a long way since August, though, so rather than rebase this branch we'd love a fresh one. Here's the shape we'd merge:

1. Grow it from one Swiss network into an Alpine weather-stations layer with an adapter per network. SLF IMIS is the first; other open networks worth a look are MeteoSwiss SwissMetNet (open government data), GeoSphere Austria's station data (Data Hub API) and the regional services in Italy, France, Slovenia and Germany. Please check each one's terms before adding it.
2. Put the proxy in `server/providers/` (the old `vite.config.js` middleware block is gone). Use fixed upstream URLs, capped reads (`readResponseTextCapped` / `readResponseJsonCapped` from `server/providers/common/http.js`), one in-flight request per upstream, and both `configureServer` and `configurePreviewServer` (see `server/providers/cyclones.js`).
3. Put the layer under `src/layers/<name>/` and add it to the Weather group. Its readings, a legend and the nearest stations in view go in a card in the WEATHER panel in the right rail (`readout: true`; see "Adding a right-rail readout panel" in `docs/CODE-BOUNDARIES.md`), rather than only on-map cards.
4. Draw stations as simple markers coloured by the chosen value, and don't rebuild all entities or hold continuous rendering on every camera move. Please also drop the imports from the fires layer.
5. Leave the share-link token to us; every letter is taken now (#658).

If that sounds good, open the new PR whenever you're ready and we'll review it.

PR #60, opened 2026-08-25
Make the Google Maps key optional

ekl254:
### The problem

`src/main.js` throws when `GOOGLE_MAPS_API_KEY` is absent, before the viewer is constructed. The app therefore cannot start at all without a billing-enabled Google Cloud account.

That is a hard gate in front of everything that needs no credentials. Ten of the thirteen layers are keyless — flights, military ADS-B, satellites, earthquakes, CCTV, radio, bikeshare, launches, mapped installations, and every bundled dataset — and none of them are reachable without first setting up Google billing.

I hit this as a new user: enabling Map Tiles requires a billing account, and the signup flow is easy to get stuck in. The README's promise that most of the globe is 🟢 isn't reachable until that's resolved.

### The change

`MapStackController` already treats a null Google tileset as the ion/OSM regime, so very little needed to move.

- **`main.js`** — the throw becomes a warning. With no key the photoreal tileset fetch is skipped and the Cesium globe stays visible.
- **`main.js`** — the initial map stack now resolves `photoreal` → `bing-aerial` → `osm` by what's actually configured. Previously `'osm'` was hardcoded whenever the Google tileset was absent, so an available Cesium ion token was ignored and users with ion got flat OSM tiles instead of Bing aerial.
- **`locations.js`** — `searchAndFlyTo` threw `No Google Maps API key available for geocoding`, breaking search entirely. It now falls back to OpenStreetMap Nominatim, reusing the existing `geocodeNavigationMode` / `flyToViewportBounds` / `flyToLandmark` framing rather than introducing a second policy.

**With a key present, every path is byte-for-byte unchanged.** The fallbacks are reached only when the key is absent.

### Limitations, stated honestly

The keyless geocoder has no viewport bias and no Places-near-view recovery, so ambiguous names ("Sixth Street") can land in the wrong city. That's documented in the JSDoc rather than papered over. Nominatim place classes are mapped onto the Google geocode types `geocodeNavigationMode` already understands, with a two-step lookup on `addresstype` then `type` — Nominatim reports airports as `addresstype: aeroway`, `type: aerodrome`, and only the latter maps cleanly.

Users without Google lose photorealistic 3D building meshes. Everything data-driven — tracking, orbits, trails, detection, cockpit — is unaffected.

### Testing

- `npm test` — **2601 passing, 0 failing**, on this branch in isolation.
- Verified in-browser keyless: app boots, map stack reports `Aerial` with an ion token and `OSM` without one, no console errors, all live layers populate (OpenSky, adsb.lol, AISStream, FIRMS, TomTom).
- Verified with a Google key present: photoreal path unchanged.

### DIFF
--- a/src/locations.js
+++ b/src/locations.js
@@ -341,6 +341,95 @@ export function findPoiByName(query) {
 /** Distinguishes an authority veto from a genuine not-found result. */
 export const CANCELLED_SEARCH = Object.freeze({ cancelled: true });
 
+// Nominatim place classes mapped onto the Google geocode types that
+// geocodeNavigationMode() already understands, so the keyless path reuses the
+// same framing policy rather than inventing a second one.
+const NOMINATIM_TYPE_MAP = {
+  country: ['country'],
+  state: ['administrative_area_level_1'],
+  region: ['administrative_area_level_1'],
+  province: ['administrative_area_level_1'],
+  county: ['administrative_area_level_2'],
+  city: ['locality'],
+  town: ['locality'],
+  village: ['locality'],
+  municipality: ['locality'],
+  suburb: ['sublocality'],
+  neighbourhood: ['neighborhood'],
+  postcode: ['postal_code'],
+  road: ['route'],
+  park: ['park'],
+  aerodrome: ['airport'],
+  stadium: ['stadium'],
+  university: ['university'],
+  peak: ['natural_feature'],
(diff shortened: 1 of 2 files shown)


bilawalsidhu:
The keyless boot fallback shipped in v0.1.0, and keyless geocoding is landing through #166. Closing with thanks; this PR flagged the geocoding gap first.

PR #62, opened 2026-08-25
fix: make Windows Precision Touchpad pinch-zoom the globe

d8ahazard:
## Summary

- Windows Precision Touchpad pinch arrives as Ctrl+wheel. Cesium recorded that event but only zoomed unmodified wheel, so laptop pinch did nothing.
- Two-finger scroll could tilt/orbit when the pad also emitted touch. On a fine pointer this PR maps pinch and continuous two-finger motion to a height-scaled zoom (`0.002 × altitude × wheel Δ`, about 2.5–2.7× Cesium's wheel) and drops unmodified PINCH from tilt.
- Click-drag orbit, Ctrl+drag tilt, discrete mouse-wheel notches, and touch-screen pinch-zoom stay on Cesium.

## Test plan

- [ ] On a Windows Precision Touchpad laptop, pinch-zoom the globe (should zoom, not page-zoom the UI).
- [ ] Two-finger scroll zooms; it should not tilt.
- [ ] Mouse wheel notches still zoom at Cesium's existing feel.
- [ ] Click-drag orbit and Ctrl+drag tilt are unchanged.
- [ ] `npm test` (includes `src/cameraTouchpad.test.mjs`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,12 @@ of current runtime behavior, see [`docs/CURRENT-STATE.md`](docs/CURRENT-STATE.md
 
 ### Fixed
 
+- Windows laptop touchpads can pinch-zoom and two-finger-scroll the globe.
+  Precision Touchpad pinch arrives as Ctrl+wheel, which Cesium recorded but
+  never bound to zoom; two-finger motion no longer tilts on a fine pointer.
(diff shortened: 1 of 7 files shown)


bilawalsidhu:
Thanks! Pinch zoom through the Ctrl+wheel path landed in #284. If Windows Precision Touchpad pinch still misbehaves on current main, please reopen with the details.

PR #63, opened 2026-08-25
feat: load Flock/ALPR cameras when Overpass returns 502

d8ahazard:
## Summary

- Official `main` does not yet have the DeFlock OSM ALPR layer. This PR adds that layer **and** the client load path that makes Flock cameras actually appear when public Overpass mirrors 502 on new/wide bboxes.
- Related to #14 (same layer, different head). This is not a re-submit of that PR's history; it is the smallest change against `main` that can show mapped ALPR/Flock locations. Prefer #14 if you want the original layer-only patch, or this one if you want the layer plus 502 resilience.
- Queries snap to a small quantized box (~0.14°) and, when the Lannon / US-45 corridor is in view, reuse the known-good cached bbox `43.10,-88.28,43.24,-88.08`. If Overpass is still down, a bundled ODbL snapshot of mapped nodes for that corridor is shown as stale fallback.
- Locations and public OSM metadata only — no plate records, vendor accounts, or private Flock data.

## Test plan

- [ ] Enable the ALPR / Flock layer over the Lannon / US-45 corridor. Cameras should appear even if Overpass returns 502 (stale snapshot / cached bbox).
- [ ] Pan nearby: queries should reuse a snapped bbox instead of aborting every frame.
- [ ] Flock-tagged cameras render distinctly from other ALPR tags; a `camera:direction` tag draws a short facing line.
- [ ] `npm test` (includes `src/data/alprCameras.test.mjs`).


bilawalsidhu:
The bbox quantization insight here is real and it's the keeper: wide metro boxes do 502 on Overpass, and snapping fetch boxes to a coarse grid so repeats hit its query cache is the right fix. I've asked for it to be folded into #14 with credit to you — #14 landed first and this branch shares its code, so that's the base.

What can't ship is the bundled snapshot: a hardcoded dump of one neighborhood (the Lannon/US-45 corridor) as the global offline fallback. The right resilience story when Overpass is down is DeFlock's own published data (API/PMTiles) — that would make a great focused follow-up if you want to build it.

Closing in favor of #14 + quantization. If you'd rather land the quantization diff yourself on top of #14, even better. (Separately: your touchpad PR #62 is queued for a hands-on pass — the macOS two-finger feel change is the part I need to test personally.)

PR #78, opened 2026-08-26
feat: boot diagnostics on the loading screen

juanatsap:
Hi Bilawal, first of all thanks for opening this. I put it on a VPS this week and it is a lot of fun.

Small one, and **I checked the open PRs and issues first**: nothing touches the loading screen. When I opened it on my phones the loading screen stayed on "Initializing photorealistic world..." forever and there was no way to know why (no console on a phone). It turned out it was just slow (the dev server sends a 10 MB `cesium.js` and my nginx was not compressing it), but I only found that out reading the nginx access log.

So this adds `src/bootDiagnostics.js`, a small module with no imports, loaded from `index.html` right before `main.js`. It does **two things**. If any error or unhandled rejection happens during boot, the status line shows "Boot error: message @ file:line" in red. And if after 20 seconds `main.js` has still not replaced the initial text, the line grows a seconds counter and a "still downloading modules; slow network?" hint. Both stop as soon as `main.js` takes over or the loading screen goes away, so on a normal boot nothing changes.

**It is loaded before `main.js` on purpose:** module scripts run in document order, so the listeners exist before the app graph starts evaluating. A failed import inside that graph never reaches the try/catch in `init()`, and that is exactly the case that used to look like a hang; the error listener uses the capture phase so a module that fails to load is reported too, not only one that throws.

**Verified** on an iPhone (Edge iOS, WebKit) and a Galaxy S24 (Chrome): when the boot is slow the counter shows up, and then the app loads as usual. Desktop unchanged. `npm test`, `npm run build` and `npm run test:track` are green. `CHANGELOG.md` and `docs/CURRENT-STATE.md` updated.

One more thing for whoever runs this behind nginx, since it is what made my phones slow in the first place. Vite serves the modules as `text/javascript`, and the usual nginx gzip config only lists `application/javascript`, so the dev `cesium.js` (10 MB, not minified) is sent uncompressed.

With this in the server block it drops to **about 2 MB**:

    gzip on;
    gzip_vary on;
    gzip_proxied any;
    gzip_comp_level 6;
    gzip_min_length 1024;
    gzip_types text/plain text/css text/javascript application/javascript application/json application/wasm image/svg+xml;

To be clear, **this is an nginx thing**. Caddy, Traefik, Apache with the default mod_deflate and Cloudflare all compress `text/javascript` by default, and Vite itself never compresses, it leaves that to whatever is in front. It only happens with nginx and the usual copy-paste gzip block, which is probably the most common setup for this. Not part of the code change, just leaving it here in case it helps. If you want it in the README "Sharing an instance" section I can add it.

Let me know if you prefer the text or the threshold different. Have a great day!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ of current runtime behavior, see [`docs/CURRENT-STATE.md`](docs/CURRENT-STATE.md
 
 ### Added
 
+- Added boot diagnostics to the loading screen: a module-load failure now
+  shows `Boot error: <message> @ <file>:<line>` instead of an eternal
+  "Initializing photorealistic world…", and a boot still waiting on modules
+  after 20 s says so with a running counter. Lives in `src/bootDiagnostics.js`,
+  loaded from `index.html` before `main.js` so it also covers a module that
+  fails to load, not only one that throws.
 - Added honest aircraft identity narration: callsign, operator, registration,
   type, and route come only from selected-contact context, and missing operator,
   route, or type enrichment is named explicitly.
--- a/docs/CURRENT-STATE.md
+++ b/docs/CURRENT-STATE.md
@@ -1,6 +1,15 @@
 # God's Eye View Current State
 
-Updated: August 24, 2026
+Updated: August 27, 2026
+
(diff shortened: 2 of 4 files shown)


bilawalsidhu:

PR #79, opened 2026-08-26
feat: lighter rendering on phones (mobile profile)

juanatsap:
Hi Bilawal, this PR makes the globe render lighter on phones, automatically. **Desktop does not change at all.**

First, so nobody sends the same thing twice: I went through the open PRs and issues before writing this and nothing covers phones. The closest are #2 (the top bar responsive) and #72 (the desktop perf items from #8). #72 also drops `preserveDrawingBuffer`, so the two touch the same line in `main.js`; I can rebase on whichever lands first.

The short version of why: the desktop settings ask the GPU to draw every frame several times over (the 3x pixel density of a phone screen, and each visual style, sharpen, bloom and scope mask being one more full-screen pass), and the photoreal tile cache has no limit. A laptop GPU does not care. A phone GPU cannot keep up, and on iOS the tab reloads itself once WebGL memory goes past **1 to 1.5 GB**.

**What changes on a phone:** the viewer now starts with no preserved drawing buffer (only recording needs it), the screen's own resolution instead of 3x, a 256 MB tile cache with a bit less detail (`maximumScreenSpaceError` 24), and the post-process passes off. The picture is slightly softer and the globe keeps running. **Anti-aliasing stays at 4x** on purpose: at the screen's own resolution it is cheap, and edges need it more there. All of it is one small file, `src/mobileProfile.js`, with the desktop and phone values side by side.

**How it decides:** touch screen and the shorter side under **900 px**. That is the whole rule, on purpose. It means a regular iPad counts as a phone here and gets the lighter profile; a 12.9 inch iPad Pro does not. Nobody has to do anything, but there are two manual switches, documented in `docs/KNOWN-ISSUES.md` under "Phones and small tablets": `?gevMobile=0` if a tablet user wants desktop quality, and `?gevMobile=1` to try the phone path on a desktop.

**One more small thing** in the same file: `?gevFlags=diag` paints a green box with browser, GPU, pixel density and canvas sizes. When someone says "it looks wrong on my phone" you can ask for a screenshot of that box instead of guessing. Without the flag nothing is added to the page. I needed exactly that this week with a Galaxy S24 and Edge; screenshots and the story in the comment below.

**Verified** on an iPhone (Edge iOS, WebKit) and a Galaxy S24 (Chrome), both load and render with the profile active. Desktop checked with the profile forced off: stages, cache and resolution exactly as before. `npm test` (4 new tests in `src/mobileProfile.test.mjs`), `npm run build` and `npm run test:track` are green. `CHANGELOG.md` and `docs/CURRENT-STATE.md` updated.

To be clear, this **does not touch the HUD layout**, which is still a desktop layout on a narrow screen.

And honestly, thanks for opening this. I have spent more hours than I should this week flying over my own island with it, and it is the first map app in a long time that made me want to work on it. Have a good one!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,13 @@ of current runtime behavior, see [`docs/CURRENT-STATE.md`](docs/CURRENT-STATE.md
 
 ### Added
 
+- Added a lighter rendering profile for phones and small tablets
+  (`src/mobileProfile.js`): no preserved drawing buffer, CSS-pixel
+  resolution, a 256 MB photoreal tile cache with coarser detail,
+  and the post-process passes (scope mask, sharpen, bloom, visual styles)
+  disabled. Desktop rendering is unchanged. `?gevMobile=1|0` forces the
+  profile; `?gevFlags=diag` shows an on-screen GPU/canvas readout for bug
+  reports. See `docs/KNOWN-ISSUES.md`, "Phones and small tablets".
 - Added honest aircraft identity narration: callsign, operator, registration,
   type, and route come only from selected-contact context, and missing operator,
   route, or type enrichment is named explicitly.
--- a/docs/CURRENT-STATE.md
+++ b/docs/CURRENT-STATE.md
@@ -1,6 +1,21 @@
 # God's Eye View Current State
 
-Updated: August 24, 2026
+Updated: August 27, 2026
+
+> **2026-08-27 — mobile rendering profile** (`src/mobileProfile.js`, applied
+> in `src/main.js`). `isMobileDevice()` is true for touch-first devices whose
+> shorter screen side is under 900 px (`?gevMobile=1|0` overrides). On those
+> the viewer is created with `preserveDrawingBuffer: false` (MSAA stays 4x),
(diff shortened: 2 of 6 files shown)


juanatsap:
**The Galaxy S24 case.** Edge 151 for Android with default settings draws vertical streaks over the whole map:

Vertical streaks over the map in Edge 151 on a Galaxy S24

Same phone with `?gevFlags=diag`: Samsung Xclipse 940 on Vulkan, WebGL2, canvas and drawing buffer the same size. Chrome on this phone reports the same GPU and backend and renders fine, which is what pointed at Edge's default ANGLE backend and not at Cesium or the app. **This box is what I would ask for in an issue.**

?gevFlags=diag readout on the same Galaxy S24

**The fix is on the phone**, and it is in `docs/KNOWN-ISSUES.md` too: open `edge://flags` in Edge, search "ANGLE", set "Choose ANGLE graphics backend" to **Vulkan** (OpenGL ES stays broken), tap Restart and reload the site. Back to Default and the streaks return. I mention it because with 1.4k forks somebody else with a Galaxy and Edge will hit it.

demolerbi-cmyk (approved):
Open

PR #82, opened 2026-08-27
fix(cctv): cap the buffered snapshot body in the frame proxy (#28)

### LINKED ISSUE #28: Cap CCTV image response bodies before buffering them

Problem:
- CCTV image fetches buffer the full upstream body with `arrayBuffer()` after only checking content type.
- A malicious or compromised configured source can return a large image body and exhaust memory within the timeout window.

Evidence:
- `vite.config.js:4374-4400` implements `fetchCctvImageFromUpstream`.
- `vite.config.js:4392` performs `Buffer.from(await upstream.arrayBuffer())` with no response-size cap.

Suggested fix:
- Enforce `Content-Length` before reading when present.
- Stream with an accumulated byte cap when length is absent or untrusted.
- Return a clear upstream-too-large error instead of buffering unbounded input.

ethanstoner:
Closes #28.

`fetchCctvImageFromUpstream` validated content-type and then buffered the whole upstream body:

```js
if (!upstream.ok || !contentType.startsWith('image/')) return null;
return { ok: true, body: Buffer.from(await upstream.arrayBuffer()), contentType };
```

There's no cap on that read. SECURITY.md says "high-volume or attacker-influenced upstream responses are capped where that boundary matters" — this path is the gap.

## Confirming it's real, not theoretical

Before writing the fix I ran the current function against a source that streams 64 chunks of 64 KB with no `Content-Length`:

```
{ isNull: false, len: 4194304, pulled: 64, cancelled: false }
```

Every chunk pulled, whole body buffered, stream never cancelled. The 8s frame timeout doesn't help — it bounds *time*, not bytes, and a fast upstream can deliver a lot in 8 seconds. Frames refresh on a 10s cadence per active camera, so this repeats.

The catalogs are server-registered, so this needs a configured source to turn hostile or be misconfigured — not a client-supplied URL. That's why I'd call it hardening rather than an exploitable hole.

## The fix

`readCappedResponseBytes` — the binary counterpart of the existing `readCappedResponseText`, placed next to it and following the same shape:

- a declared `Content-Length` over the cap is refused **before a single body byte is pulled**
- a missing or lying one is caught as chunks arrive, and the stream is cancelled
- overflow returns `null`

`null` is what makes this cheap: the frame route already treats it as an upstream miss, so an oversize body falls through the existing Street View → synthetic-SVG chain instead of erroring. No new failure mode, no new response shape.

Cap is 16 MB (`CCTV_FRAME_MAX_BODY_BYTES`), exported so the test can pin it. Public CCTV stills run well under a megabyte; 16 MB clears even a 4K PNG. It sits under the media route's existing 64 MB declared cap, and the test asserts that ordering so the two can't drift.

## Tests

Four added to `src/data/cctvProxy.test.mjs`:

| Test | Pins |
|---|---|
| declared oversize | cancelled, ≤1 chunk pulled (the stream's own prefetch) |
| undeclared oversize | cancelled mid-stream, ≤6 of 64 chunks pulled |
| chunked under cap | every chunk reassembled, correct byte length |
| existing valid image | unchanged |

Verified red→green: with the old `arrayBuffer()` line restored, the two oversize tests fail; with the fix, all five pass.

One deliberate detail — the oversize tests assert on `result?.body?.length` rather than on `result` itself. Asserting object equality made a regressed run hang: `assert.equal` builds a diff of the 4 MB Buffer, which takes longer than the check it's reporting on. Comment in the test says so.

## Gates

```
npm test    ℹ tests 2590  ℹ pass 2590  ℹ fail 0
npm run build   ✓ built (only the pre-existing chunk-size warning, #40)
```

Node 24.14.0. I could not run `test:track` or the `qa:*` gates — they need a Google Maps key with the Map Tiles API enabled.

CHANGELOG entry added under Security. I left `docs/CURRENT-STATE.md` alone since no product-visible behavior changes — the fallback chain it documents is unchanged. Happy to add a note there if you'd rather it be recorded.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -115,6 +115,10 @@ represent previously published GitHub Releases.
 
 ### Security
 
+- The CCTV frame proxy caps one buffered snapshot at 16 MB while it reads. A
+  configured source that declares or streams a larger body is cancelled and
+  treated as an ordinary upstream miss, so the Street View and synthetic
+  fallbacks still run.
 - Production transitive dependencies resolve to patched DOMPurify and
   protobufjs releases without changing the Cesium version or application APIs.
 - Production dependency audit reports no known advisories; remaining audit
--- a/src/data/cctvProxy.test.mjs
+++ b/src/data/cctvProxy.test.mjs
@@ -2,8 +2,26 @@ import { test } from 'node:test';
 import assert from 'node:assert/strict';
 import {
   CCTV_FRAME_FETCH_TIMEOUT_MS,
+  CCTV_FRAME_MAX_BODY_BYTES,
   fetchCctvImageFromUpstream,
 } from '../../vite.config.js';
+
+/** A body that arrives in chunks and never declares a Content-Length. */
+function chunkedImageResponse(chunkBytes, chunkCount, { onChunk = () => {}, onCancel = () => {} } = {}) {
+  const stream = new ReadableStream({
+    async pull(controller) {
+      if (chunkCount <= 0) {
+        controller.close();
+        return;
+      }
+      chunkCount -= 1;
+      onChunk();
+      controller.enqueue(new Uint8Array(chunkBytes));
+    },
+    cancel: onCancel,
+  });
+  return new Response(stream, { status: 200, headers: { 'Content-Type': 'image/jpeg' } });
+}
 
 test('CCTV upstream frame fetch supplies a bounded abort signal', async () => {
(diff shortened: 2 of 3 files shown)


rajpratham1 (approved):
LGTM — the fix correctly removes the browser-side node:fs dependency and adds a useful regression test for the module boundary.

bilawalsidhu:

PR #83, opened 2026-08-27
fix(data): keep node:fs out of the browser-built data modules (#34)

### LINKED ISSUE #34: Remove node:fs imports from browser-built data modules

Problem:
- Production browser build emits warnings that `node:fs` is externalized for browser compatibility.
- The affected modules are browser data loaders, so the runtime boundary is fragile and easy to break with future bundler/env changes.

Evidence:
- `src/data/naturalEarthRegions.js:116-128` dynamically imports `node:fs` behind an environment guard.
- `src/data/neighborhoodPolygons.js:97-107` has the same pattern.
- `npm run build` warns that `node:fs` was externalized for browser compatibility for both modules.

Suggested fix:
- Move Node-only file loading into a Node-only module or build-time script.
- Keep browser modules free of `node:` imports.
- Add a build check that fails on Node core module externalization warnings in browser chunks.

ethanstoner:
Closes #34.

`npm run build` emits these on every production build:

```
[plugin vite:resolve] Module "node:fs" has been externalized for browser
compatibility, imported by ".../src/data/naturalEarthRegions.js".
[plugin vite:resolve] Module "node:fs" has been externalized for browser
compatibility, imported by ".../src/data/neighborhoodPolygons.js".
```

Both modules carried the same dual path — the comment in `neighborhoodPolygons.js` explains why it was there:

```js
if (isNode) {
  // node:test path — plain dynamic JSON import needs an import attribute in Node,
  // so read the file directly (same dual-path pattern as naturalEarthRegions.js).
  const { readFileSync } = await import(/* @vite-ignore */ 'node:fs');
  ...
} else {
  const mod = await city.loader();   // Vite bundles the JSON file as a module
}
```

Vite only *warns* about externalizing `node:*`, so the import survives into the browser build and the boundary rests entirely on the `isNode` guard being right.

## The fix

The comment names the real constraint: Node needs an import attribute. So give it one, and the two paths collapse into one that both runtimes serve:

```js
loader: () => import('./local_data/neighborhoods/san-francisco.json', { with: { type: 'json' } })
```

Vite bundles the JSON exactly as before, Node loads the same file under `node:test`, and `isNode` plus `node:fs` disappear from both modules. Net −27/+22 lines in the two data modules.

## Verification

**Build.** Zero `externalized` warnings. The emitted JSON chunks are byte-identical to before — same content hashes:

```
regions-RPMKg9pq.js  marine-BJ61ZZ9E.js  san-francisco-B-TUYeaR.js
```

**Browser, not just build.** A build passing doesn't prove a dynamic import with an attribute still resolves at runtime, so I built a probe entry against these two modules and loaded it in headless Chrome:

```
{"neighborhoodName":"Chinatown","neighborhoodRingPoints":12,
 "saharaFound":true,"regionCount":1338}
```

Identical to the same calls under Node. No console errors (the one 404 was the probe page's own missing favicon). Probe was scaffolding and isn't in the diff.

**Regression guard.** `src/browserModuleBoundary.test.mjs` scans every non-test `.js` under `src/` for static and dynamic `node:` imports. Verified it fails on the pre-fix tree, naming exactly the two modules from the issue:

```
AssertionError: Node core imports reached the browser build:
  data/naturalEarthRegions.js, data/neighborhoodPolygons.js
```

This is the "build check that fails on Node core module externalization" the issue asks for, done as a unit test so it runs in `npm test` alongside the other source-shape checks (`cockpitMarkup.test.mjs`, `director.test.mjs`) rather than adding a build step.

## Gates

```
npm test    ℹ tests 2588  ℹ pass 2588  ℹ fail 0
npm run build   ✓ built, no externalization warnings
```

Node 24.14.0. `test:track` and the `qa:*` gates need a Google Maps key with the Map Tiles API enabled, which I don't have — but the browser probe above covers the runtime path these two modules own.

## One thing to check

`with { type: 'json' }` needs Node ≥ 20.10 and Rollup 4 / Vite 5+. `package.json` requires Node ≥ 24.14 and Vite ^6, so both are comfortably clear — flagging it in case you support an older toolchain somewhere I can't see.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,9 @@ of current runtime behavior, see [`docs/CURRENT-STATE.md`](docs/CURRENT-STATE.md
   outages and wait for measured photoreal-surface evidence before a 3D model
   takes over from its billboard.
 - Cockpit altitude uses aviation MSL data rather than Cesium render height.
+- The bundled Natural Earth region and neighborhood-polygon packs load through a
+  single import path in both the browser and `node:test`, so the production
+  build no longer externalizes `node:fs` for two browser data modules.
 
 ### Security
 
--- a/src/browserModuleBoundary.test.mjs
+++ b/src/browserModuleBoundary.test.mjs
@@ -1,0 +1,36 @@
+import { test } from 'node:test';
+import assert from 'node:assert/strict';
+import { readFileSync, readdirSync } from 'node:fs';
+import path from 'node:path';
+import { fileURLToPath } from 'node:url';
+
+const SRC_ROOT = fileURLToPath(new URL('.', import.meta.url));
+
+/** Every browser-built module under src/ — the test files are Node-only. */
+function browserModules(directory = SRC_ROOT) {
+  const files = [];
+  for (const entry of readdirSync(directory, { withFileTypes: true })) {
+    const absolute = path.join(directory, entry.name);
+    if (entry.isDirectory()) files.push(...browserModules(absolute));
+    else if (entry.isFile() && entry.name.endsWith('.js')) files.push(absolute);
+  }
+  return files.sort();
+}
+
+test('no browser-built module imports a Node core module', () => {
+  // Vite externalizes `node:*` for the browser and only WARNS, so a stray
+  // import survives the build and turns into a runtime failure the moment the
+  // guard around it is wrong. src/data/naturalEarthRegions.js and
+  // src/data/neighborhoodPolygons.js both carried one to read their bundled
+  // JSON packs under node:test; an import attribute serves both runtimes.
+  const offenders = [];
+  for (const file of browserModules()) {
+    const source = readFileSync(file, 'utf8');
+    // Static `from 'node:fs'` and dynamic `import('node:fs')`, quoted either way.
+    if (/\bfrom\s*['"]node:|\bimport\s*\(\s*(?:\/\*[^*]*\*\/\s*)?['"]node:/.test(source)) {
+      offenders.push(path.relative(SRC_ROOT, file).split(path.sep).join('/'));
+    }
+  }
+
+  assert.deepEqual(offenders, [], `Node core imports reached the browser build: ${offenders.join(', ')}`);
+});
--- a/src/data/naturalEarthRegions.js
+++ b/src/data/naturalEarthRegions.js
@@ -113,19 +113,13 @@ function suffixVariants(norm) {
 /** @type {Array|null} flat entry list for listRegions() */
 let _entries = null;
 
-const isNode = typeof process !== 'undefined' && !!process.versions?.node
-  && typeof window === 'undefined';
-
 async function loadPackFile(base) {
-  if (isNode) {
-    const { readFileSync } = await import(/* @vite-ignore */ 'node:fs');
-    const url = new URL(`./local_data/natural_earth/${base}.json`, import.meta.url);
-    return JSON.parse(readFileSync(url, 'utf8'));
-  }
-  // Vite bundles these JSON files as modules (same pattern as neighborhoodPolygons.js)
+  // Vite bundles these JSON files as modules; the import attribute is what Node
+  // needs to load the same files under node:test (same pattern as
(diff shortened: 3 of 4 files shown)


rajpratham1 (changes requested):
There is a significant concern in the change: the PR removes the Node-specific fs fallback and makes both browser and Node use JSON imports with { with: { type: 'json' } }. That may work with the stated Node/Vite setup, but the PR's new test only checks for textual node:* imports; it does not actually verify the browser build or that the JSON loading works correctly in the supported environments.

Also, the PR shows no visible checks/results while the change affects the browser build boundary.

ethanstoner:
Two of those are fair and one isn't.

Fair: `browserModuleBoundary.test.mjs` is a textual scan by design — it guards the boundary (no `node:*` specifier in a browser-built module), not the loading. And the repo runs no CI, so there are no check results to show on any PR here.

**The Node path is covered**, just not by the file added here. Two test files already exercise these loaders, and this PR deletes the `fs` branch they used to take — `naturalEarthRegions.test.mjs` (`findNaturalRegion`, `lookupNaturalRegionOutline`, `listRegions` → `loadPackFile`) and `neighborhoodPolygons.test.mjs` (`lookupNeighborhoodRing` → `cityLoader`). On this branch, with `grep -c "node:fs"` returning 0 in both modules:

```
$ node --test src/data/naturalEarthRegions.test.mjs src/data/neighborhoodPolygons.test.mjs
# tests 20
# pass 20
# fail 0
```

With the fallback gone those 20 assertions can only pass through the import attribute.

**The browser build — I built both branches to check.** Both succeed (exit 0). The difference is exactly two warnings, on exactly the two files this PR touches.

`main`:

```
[plugin vite:resolve] Module "node:fs" has been externalized for browser compatibility,
imported by ".../src/data/naturalEarthRegions.js".
[plugin vite:resolve] Module "node:fs" has been externalized for browser compatibility,
imported by ".../src/data/neighborhoodPolygons.js".
```

This branch: no externalization warning at all.

That warning is not cosmetic. In `main`'s shipped bundle, `node:fs` becomes an empty frozen module:

```js
const i3=Object.freeze(Object.defineProperty({__proto__:null},Symbol.toStringTag,{value:"Module"}))
```

so `const { readFileSync } = await import('node:fs')` yields `undefined` and the next line calls it. Only the `isNode` guard stops that being a `TypeError` in production — which is the failure mode this PR removes rather than keeps guarding. This branch's bundle has 0 occurrences of `readFileSync` and 0 of `node:fs`.

The packs are bundled as lazy chunks and reachable from the index chunk: `assets/regions-*.js` (1.99 MB, contains `Pentecost` and `ne_10m_geography_regions_polys`), `assets/marine-*.js` (633 kB), `assets/san-francisco-*.js` (222 kB, contains `Bayview Hunters Point`).

One thing I hadn't noticed before building. On `main`, each pack ships **twice** — once as a JS chunk for the browser branch, and once as a raw `.json` asset, because the Node branch's `new URL('./local_data/.../x.json', import.meta.url)` makes Vite emit a URL asset too. Both sets are referenced from `main`'s index chunk. Removing the dual path drops the duplicates:

```
dist total:   30,536,517 B (main)  →  27,692,834 B (this branch)     −2.84 MB
dist/assets:  14,373,781 B         →  11,530,098 B
modules:      156                  →  152
```

So this also removes ~9% of the build output, which wasn't in the PR description.

Caveat on the numbers: I built on Node 22.15.1, below the declared `engines` floor of 24.14.0. Vite 6.4.3 built both branches fine and the comparison is apples-to-apples, but it isn't the declared runtime.

bilawalsidhu:
Merged — thank you, including the boundary scan test, which now guards the whole tree upstream too. Good catch, cleanly fixed, and your review-thread reasoning made this an easy take.

bilawalsidhu:
Shipped in [v0.1.0](https://github.com/bilawalsidhu/gods-eye-view/releases/tag/v0.1.0) — the browser module boundary fix and your scan test are in the release, and that test now guards the whole tree upstream as well. Thanks @ethanstoner; your authored commit is intact on main and credited in the release notes.

Issue #85, opened 2026-08-27
Feature: Live weather radar and satellite imagery overlay layer


chiang21fcb:
Problem

God's Eye View already has environmental layers for fires (FIRMS), earthquakes (USGS), and an opt-in cockpit WX cloud effect. But there is no real-time weather data layer for the globe itself. A weather radar or satellite imagery layer would provide critical context for the other layers: flight diversion patterns, storm formation, and current conditions anywhere on the planet.

Proposed solution

Add a weather radar and satellite imagery overlay as a new toggleable layer:

- Global satellite cloud imagery (GOES, Himawari, Meteosat composite)
- Regional weather radar reflectivity where available (NOAA MRMS for US, local equivalents)
- Optional temperature, wind, or precipitation contour overlays

Data sources (all public and free tier)

- NOAA GOES satellite imagery (public, no API key)
- OpenWeatherMap (free tier, optional API key for more data)
- NOAA MRMS radar (public, US only)
- EUMETSAT for European coverage (public)

How it would work

Similar to the existing map stack switching, the weather layer would be a toggleable overlay. Tiles would be fetched through a lightweight proxy (already the pattern used for FIRMS, TomTom, etc.) and rendered as a semi-transparent overlay on the 3D globe.

The layer would:
- Show current cloud cover as a semi-transparent overlay
- Animate the last few hours of radar and satellite data with a time slider
- Respect the existing sensor modes (FLIR, NVG, etc.)

Why this fits God's Eye View

The project is about making the world's signals visible. Weather is one of the most fundamental signals. It provides context for flights (diversions, holds), ships (storm avoidance), fires (wind direction), and the cockpit WX mode already hints at this capability. A proper weather layer would make the globe feel even more alive.

bilawalsidhu:
I do love this idea. Let's see if there's a way to do it performantly. The goal should be to make it easy for people to set this up as well, rather than introducing new paid API sources, so I like your approach. If anyone would like to tackle this as a focused PR, please let me know!

chiang21fcb:
@bilawalsidhu thanks for the response. Glad the public-data approach aligns with the project philosophy.

[...]

PR #89, opened 2026-08-28
feat: add weather radar layer using RainViewer free tile service (PoC for #85)

chiang21fcb:
WIP proof of concept for the weather radar layer proposed in #85.

## What this does

Adds a weather radar overlay using RainViewer's free tile service (no API key required). The layer fetches the latest radar frame every 10 minutes and renders it as a semi-transparent overlay on the Cesium globe.

## How it works

- `src/data/weatherRadar.js` — new layer module following the standard DataLayerManager contract (enable, disable, getStats, etc.)
- Uses `Cesium.UrlTemplateImageryProvider` with RainViewer's tilecache
- No proxy needed: Cesium loads tiles via `` tags, no CORS issues
- Registered in `localLayers.js` and `layerState.js` with token `z`

## Status

This is a proof of concept. I have not been able to run it locally (requires Google Maps API key and Cesium runtime). The layer module follows the standard contract, but the tile layer behavior needs visual verification.

## What to verify

- The RainViewer API returns valid radar timestamps
- The tile URL template works correctly with Cesium's coordinate system
- The semi-transparent overlay looks correct on the 3D globe
- The 10-minute refresh cycle works without memory leaks

## Changes

- `src/data/weatherRadar.js` (new, 100 lines)
- `src/data/localLayers.js` (+4 lines)
- `src/data/layerState.js` (+1 line)


bilawalsidhu:
Thanks for the PoC — the direction works: keyless source, single imagery layer, registry token, opt-in. Reviewed the diff (not run yet). Five things, four blocking:

1. **The tile URL is wrong — the layer can't render.** RainViewer's `weather-maps.json` returns per-frame `path` values that are opaque hashes now (e.g. `/v2/radar/b8ed988a56a2`), not timestamp-composed paths, and tile URLs also need the tile-size segment. Compose `host + frame.path + '/256/{z}/{x}/{y}/2/1_1.png'` from the API response instead of hardcoding a template around `{nowcastTime}`. (Checked against the live API today — substituting the timestamp 404s.)
2. **`destroy` throws.** `destroy: (viewer) => { this.disable(viewer) }` — an arrow function doesn't get the object's `this`, so `this.disable` is undefined and destroy raises a TypeError. Share a closure-level helper between `disable` and `destroy` instead.
3. **Registry tests.** The new `weather-radar` entry in `LAYER_STATE_REGISTRY` needs `src/data/layerState.test.mjs` updated to match — `npm test` fails as-is.
4. **Credits.** Every source registers in `src/data/dataCredits.js` and gets a row in `DATA_SOURCES.md` — house rule, and RainViewer's terms want visible attribution.
5. *(non-blocking)* Unit tests for the pure parts (latest-frame selection: happy/empty/error) in the style of `earthquakes.test.mjs`; bump `REFRESH_MS` to 30 min per the issue thread; add the missing trailing newline.

On not being able to run it: the next release boots fully keyless, so you'll be able to verify visually with zero setup. I'll put eyes on the rendering from this side once the fixes land.

bilawalsidhu:
Follow-up with corroboration from the wild: @onionviolet merged this into their fork and had to revert it (`f7b6412` on their fork) — three tests fail: the production-registry exactness test and both unknown-layer-token rejection assertions. The second failure is a subtle one: `layerState.test.mjs` uses `l=z` as its *unknown token* fixture, and this PR registers `z` as a real token — so along with the registry expectations, the fixture needs to move to a deliberately still-unused letter. Pick the layer token with the test file open.

Thanks @onionviolet — a merge-test-revert on a live fork is exactly the verification this PoC needed.

onionviolet:
Took the diagnosis and fixed all four blocking points on a branch off @chiang21fcb's head, so it should apply cleanly to this PR:

https://github.com/onionviolet/gods-eye-view/tree/fix/weather-radar-review

**Tile URL.** Confirmed against the live API. `frame.path` is `/v2/radar/3ae628836a15` and the composed tile returns HTTP 200, 4,753 bytes. Now `host + frame.path + '/256/{z}/{x}/{y}/2/1_1.png'`. Worth flagging how this failed: Cesium renders an empty imagery layer instead of raising on 404, so the layer looks enabled and draws nothing. Nothing in the test suite or the console would have caught it.

**destroy.** Both paths share a closure-level `teardown()`.

**The `z` fixture.** Your read was right, and there is a second instance: `src/sharelink.celestial.test.mjs:61` uses `l=z` the same way. Rather than move the fixture to another free letter, both now derive the first unregistered letter from `LAYER_STATE_REGISTRY`. Picking a letter reschedules the same failure for layer 18.

**Credits.** Registered in `dataCredits.js` with a `DATA_SOURCES.md` row.

Also took the non-blocking items: `REFRESH_MS` to 30 minutes, trailing newline, CHANGELOG entry, and six unit tests on `latestRadarTileUrl`, which is exported and pure so the frame-selection contract is testable without a viewer.

`npm test` is 2,593 passing, `npm run build` is clean, and `npm run test:track` is 96 passed, 4 failed.

All four failures are the same assertion family (`no console errors` / `no HTTP 5xx`) and the same single cause: `/api/agent/models?provider=ollama` returns 502 because I have no Ollama running. That endpoint comes from #51, which is not on this branch, so it cannot fire on the diff itself. Every `display-floor` and terrain test passes.

Two things I hit on the way that may be worth their own issues, both unrelated to this PR:

- `scripts/dev-fresh.sh` exits on a missing `GOOGLE_MAPS_API_KEY` at line 76, which is before `CESIUM_ION_TOKEN` is resolved at line 201. With #60 or #51 merged the app supports an ion-only route, but the launcher still refuses to start it, and the error names the wrong credential.
- `/api/terrain/heights` returned 502 for a while on Node 26.4.0 even though `terrain.reearth.land` answered 200 for `layer.json` and a `.terrain` tile from the same machine. It cleared once a valid ion token was in place, so I cannot cleanly separate it from my own misconfiguration, but the 502 body was `terrain heights fetch failed and no cache available for every point` rather than anything about credentials.

Happy for @chiang21fcb to cherry-pick from that branch, or I can open it as a PR against their fork, whichever is less disruptive.

chiang21fcb:
﻿@onionviolet thanks for the thorough fix and for running the full test suite against it. The composed tile URL and the closure-level teardown both look correct to me. I will cherry-pick from your fix/weather-radar-review branch and push the update to this PR. Thanks also for catching the second z fixture in sharelink.celestial.test.mjs and for the heads-up about the dev-fresh.sh launcher issue.

bilawalsidhu:
The weather radar layer is landing through the next maintainer batch. Closing this proof of concept with thanks for opening the design.

PR #55, opened 2026-08-25
feat: add Docker image and GHCR publishing

DarshSinghKushwaha:
## Summary

Implements the Docker/GHCR portion of Issue #48.

### Changes

- Add Dockerfile using Node.js 24.14.x
- Add `.dockerignore`
- Add GitHub Actions workflow for GHCR publishing
- Add Docker usage and local build instructions to README

### Docker usage

```bash
docker run -d \
  --name gods-eye-view \
  -p 4173:4173 \
  --env-file .env \
  ghcr.io/bilawalsidhu/gods-eye-view:latest
```
Resolves #48

PR #94, opened 2026-08-28
Add Docker support for local development

nitsuah:
## Summary
- Adds a `Dockerfile` (node:24-slim, runs the Vite dev server) and `docker-compose.yml` so the app can be built and run in a container via `docker compose up`
- Adds `.dockerignore` to keep `node_modules`, `.env`, `dist`, and local files out of the build context
- `PORT`/`HOST` are wired through to match `vite.config.js`'s existing `env.PORT`/`env.HOST` handling, so the container exposes port 4173 correctly

## Test plan
- [ ] `docker compose up --build` starts the app and it's reachable at `http://localhost:4173`

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,4 @@
+node_modules
+.env
+*.local
+dist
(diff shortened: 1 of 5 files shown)


upchui:
👍

bilawalsidhu:
Docker support is being consolidated into a single PR (#208 or #55). Closing this one so there's one Docker change to review; thanks for the compose file.

Issue #41, opened 2026-08-25
Split security-sensitive middleware out of the 7k-line Vite config


jsawyerdev:
Reviewed commit: 880a672b5e16ad3e41d318801d3a5203f9201923

Problem:
- `vite.config.js` is serving as Vite config, API server, proxy layer, cache manager, budget gate, env parser, and security boundary.
- The file is 7,383 lines, which makes security review and regression testing of key-bearing routes unnecessarily difficult.
- `src/ui.js` is also over 10,000 lines, compounding review and ownership risk on the client side.

Evidence:
- `vite.config.js` is 7,383 lines and registers many API/proxy plugins in `vite.config.js:7342-7363`.
- `src/ui.js` is 10,293 lines.
- Multiple findings in this audit come from unrelated security-sensitive paths sharing one config file.

Suggested fix:
- Move server middleware into small modules grouped by trust boundary/provider.
- Add focused tests for each proxy module instead of relying on one giant config import.
- Keep `vite.config.js` responsible for composing plugins, not implementing every API route inline.

This was found during an adversarial repository review; impact depends on how the dev/preview server is exposed.

bilawalsidhu:
Good feedback it has definitely ballooned in size.

bilawalsidhu:
Thanks for the report. Checked against public main `2746ec7`: Current root Vite config is a small composition facade and server providers are split into focused modules.

[Current implementation](https://github.com/bilawalsidhu/gods-eye-view/blob/2746ec753ce4d6ebafee881a212c3f327907c673/vite.config.js). Closing the original report as resolved/superseded. If the same behavior reproduces on current main, please reopen with the commit, environment and reproduction.

PR #100, opened 2026-08-29
feat(ocean): ocean observations, GSHHG land/sea mask, animated current field, drift Monte Carlo

MarioDanielPanuco:
> **This PR has grown a lot since it was opened.** It was an ocean-observation
> layer plus a drift simulator; it is now those plus a bundled land/sea mask and
> an animated surface-current field. If the size is a problem I am happy to split
> it — the four pieces are separable and I've noted the seams at the bottom.

## What

Four pieces, stacked, each usable without the ones above it.

### 🌊 Ocean Conditions — NDBC buoy observations (`ocean-conditions`, token `o`)

~900 NOAA NDBC stations from **one** server-cached bulk fetch (`/api/ocean/obs`:
10 min TTL, 60 min stale-serving, non-fatal station-name join from
`activestations.xml`) — never per-station polling. Rendered as static-geometry
entities color-banded by significant wave height.

Cards show **observed values sparsely**: a wave-only station gets wave lines,
missing fields are omitted rather than rendered as placeholders. Open-Meteo
**Marine forecast lines are appended `FC`-prefixed** so observation and forecast
are never conflated. Direction fields name their convention — NDBC `MWD`/`WDIR`
and Open-Meteo `wind_direction_10m` read `from N°` (meteorological),
`ocean_current_direction` reads `toward N°` (oceanographic) — because a bare
arrow shared one glyph between two opposite meanings.

Full analyst + voice integration ("turn on buoys", "sea state", "waves"), with
analyst queries over `waveHeightM, wavePeriodS, windSpeedMs, gustMs, sstC,
pressureHpa`.

### 🗺️ Bundled land/sea mask (GSHHG-derived)

A global **1/8° three-state raster** — water / land / coastal-mixed, 2 bits per
cell, 2880×1440, exactly 1,036,816 bytes — built offline from GSHHG v2.3.7 by a
committed regeneration script, with a zero-dependency native `.b` parser, an
even-odd scanline rasterizer (per-ring longitude unwrap for the dateline,
artificial south-pole seam edges for the globe-circling Antarctic ice front), and
a `GEVM`-header codec shared by the build script, the browser loader and the
drift worker.

Three states rather than two, and the third is the point: a cell a shoreline
passes through is *partly wet*, so it is never trusted either way. Ocean clicks
on a coastal cell fall back to the live probe; drift particles never beach there.

Measured composition of the committed artifact: 64.963% water, 32.113% land,
2.925% coastal by cell count (69.948% / 27.321% / 2.731% area-weighted).

**Licence:** GSHHG is LGPL, so the derived raster is too. `LICENSE` gains a
carve-out alongside the existing TeleGeography and OSM ones, and
`DATA_SOURCES.md` records the source URL, fetch date, input SHA-256 and
regeneration command.

### 🌀 Ocean Currents — the field itself, animated (`ocean-field`, token `n`)

`/api/ocean/field` serves a **two-tier ladder, and every tier that contributes is
named on screen with its own age, resolution and physics.**

- **Nearshore**: IOOS HF-radar total vectors (nine CoastWatch ERDDAP datasets,
  0.5–6 km, hourly) through QC gates — `hdop ≤ 1.6`, `|u|,|v| ≤ 2.4 m/s` per
  component, ≥ 40 vectors in a **single** served hour (never a multi-hour pool:
  M₂ is 12.42 h), served hour ≤ 12 h old — then a **two-pass Barnes objective
  analysis** with holdout cross-validation. `L = 2.05·Δ` floored at 2.05 km,
  solved so the two-pass response reaches half amplitude at exactly 4Δ.
- **Everywhere else**: **HYCOM ESPC-D-V02** (0.04°×0.08°, 3-hourly, −10 d…+5 d) —
  a primitive-equation forecast carrying wind-driven flow and eight tidal
  constituents, so it is physically comparable to the radar tier rather than
  merely adjacent to it. NOAA CoastWatch's 0.25° blended altimetry is retained as
  an **automatic fallback**; it is absolute geostrophic velocity and nothing else
  (no Ekman, no tides), and the legend says which one served and what it carries.
- **Both**: a view served by both is `composite` and states the split — "11% of
  drawn cells are 1 km HF radar (4 h old); the other 89% is 4.4 km HYCOM global
  forecast currents (40 min old)". Each drawn cell keeps exactly one provenance:
  radar, then fill, then `null`. Never averaged, never zero.

Client side: 4,000 particles advected through a 16 px screen-space velocity
raster rebuilt on camera settle, so the projection's meridian convergence, tilt
and foreshortening are measured per node rather than approximated. Time
compression is solved per rebuild to hold a median current at 1.1 px/frame,
keeping the field legible across the ~400× range in screen speed between a
100 km view and the whole Earth. The colour ramp is strictly increasing in WCAG
relative luminance (0.169 → 0.636) with hue falling 192° → 41°, interpolated in
linear light.

**Coverage is measured over water cells**, using the mask above — counting land
against a coastal analysis demotes exactly the views HF radar exists to serve.
The same radar hour scored 20% over a 1°×1° box centred on Monterey Bay and
99.9% over a 0.6°×0.6° box moved offshore; the difference was land, not data.
A cell with no data is drawn as nothing, never as slack water.

Tier logic and ingest live in `src/server/ocean/*`, not in `vite.config.js`
(issue #41) — the route handler is 65 lines over 7 modules.

### 🧍‍🌊 Person-overboard drift Monte Carlo

A `▶ DRIFT` chip on buoy and ocean-point cards runs a stochastic **leeway drift
ensemble** — Breivik & Allen (2008) formulation, USCG taxonomy **PIW-1**
coefficients (Allen & Plourde 1999 / Allen 2005), transcribed from OpenDrift's
`OBJECTPROP` table with citations in `src/sim/leeway.js` and verified against it
verbatim. Classical RK4 with step-held stochastic terms (measured: forward-Euler
global error 0.052 m vs RK4 8.0 × 10⁻⁶ m on the shipped configuration), a
zero-noise deterministic control particle, backward/hindcast integration, a
turbulent-diffusion term, and beaching against ETOPO bathymetry with the bundled
mask as offline fallback.

Labelled `SIMULATED DRIFT ENSEMBLE — NOT A SAR PRODUCT` everywhere, following the
repo's `RECONSTRUCTED ESTIMATE` honesty grammar. **What the ensemble actually
is**, measured and stated in the panel rather than implied: at the defaults
(n = 10⁴, 24 h, dt = 10 min, σ_pos = 1 km, σ_turb = 0.05 m/s, 8 m/s wind), the
variance decomposition is leeway residual + crosswind sign **153.70 km² (98.1%)**,
initial position scatter 2.00 km² (1.3%), turbulence 0.26 km² (0.2%). The spread
is a leeway-uncertainty cloud, not a resolved-flow prediction, and the panel says
so.

## Fixes found while auditing this branch

I ran a review over the whole branch before updating the PR. Seven defects came
out of it; all are fixed here, each with a test that fails without the fix.

1. **The HYCOM tier never served.** `parseHycomAscii` required both velocity
   components to carry a coordinate map named exactly `time`, but the THREDDS
   FMRC aggregation numbers the two axes apart — the `.dds` declares
   `water_u[time]` beside `water_v[time1]` — so the parser rejected *every*
   well-formed response as shape drift and fell through to the altimetry
   fallback on 100% of requests. Measured across five disjoint boxes (Monterey,
   mid-Pacific, North Sea, antimeridian, equator): 5/5 failed against a healthy
   upstream. The test fixture reproduced the server's variable *ordering* quirk
   but not its *naming* quirk, so all 52 of the module's tests passed against a
   body the server never sends. Now matched by family — safe because the two
   axes are the same axis under two names (max |time − time1| = 0 across all 129
   steps) — and the fixture emits the real shape, which turns 10 of its tests
   into a regression guard.
2. **A forecast field rendered as `just now`.** `resolveAgeMs` clamped with
   `Math.max(0, …)`, discarding the negative `ageMs` HYCOM reports on purpose, so
   a field valid five days out looked like a just-published analysis with
   `stale: false`. The sign is kept on the fetcher-reported branch (measured
   against *now*) and still clamped on the derived branch (measured against the
   *requested* instant, which may be historical); `formatAge` renders a negative
   age as `4 days ahead`.
3. **The Ocean Currents layer reported no status.** It exposed `getStatus()`
   while `DataLayerManager` reads `getStats()`, so its DATA panel row showed a
   blank count and "never" permanently — including while `refresh()` was failing.
   The contract test asserted the wrong name too; it now drives
   `DataLayerManager._moduleStats` itself.
4. **`/api/ocean/field` answered 500 for every malformed request.**
   `normalizeBox` refuses by throwing, so the handler's `if (!box)` 400 branch
   was unreachable. Verified live against five invalid inputs. Four route-level
   tests now drive the middleware — previously all 26 tests in
   `oceanProxy.test.mjs` called pure helpers, leaving dispatch, method checks and
   status codes covered by nothing.
5. **`node:fs` was back in a browser module.** The mask loader branched on an
   `isNode` check, reintroducing what #83 removed. Now two modules over one pure
   codec, and `browserModuleBoundary.test.mjs` gains a second assertion — no
   browser module imports from `src/server/` — which is what makes excluding that
   tree from the first assertion sound rather than a hole.
6. **Voice could not turn the current field on.** `ocean-field` was in the
   registry and `main.js` but in no tool enum and had no aliases, so "show me the
   ocean currents" resolved to the buoys.
7. **Documentation described a design the code could not reach.** With HYCOM
   actually serving, README / `DATA_SOURCES` / `CURRENT-STATE` / the credits /
   the `@file` block / the composite provenance strings are reconciled. HYCOM had
   no credit entry at all while it was the primary tier; the CoastWatch entry
   said "U.S. public domain" where the dataset's own licence text requires
   Copernicus and AVISO+ acknowledgements. A Monterey coverage figure recorded as
   20% in two places and 25% in a third is reconciled to 20%.

## Verification

All three gates green on the merged tree:

| Check | Result |
|---|---|
| `npm run build` | ✅ exit 0 |
| `npm test` | ✅ **3,182 pass / 0 fail**, across 24 new test files |
| `npm run test:track` | ✅ **108/108**, no console errors, no HTTP 5xx |

Live endpoint smoke against a fresh dev server (2026-09-02):

- `/api/ocean/obs` → 200, real NDBC bulk feed.
- `/api/ocean/marine`, `/marine-grid` → 200, served-coordinate georeferencing in
  the payload.
- `/api/ocean/etopo` → 200.
- `/api/ocean/field` → 200. Global-only box: HYCOM, `kind: modeled`, 17.72 km,
  40 min old. Monterey: `composite`, holdout RMSE 0.031 m/s, 11% radar / 89%
  HYCOM, 69% water coverage.
- Malformed input → **400** with the failing bound named, on all five refusal
  classes and on an omitted bound.

All five `/api/ocean` routes register in **both** `configureServer` and
`configurePreviewServer` — they share one `install(server.middlewares)`, so
preview parity is structural rather than per-route.

`upstream/main` (v0.1.1) is merged in. Three conflicts: `CHANGELOG.md` folded
under the new bare `## [Unreleased]`; `README.md`'s layer count combined and
verified against the table itself (15 rows, 13 keyless); and the frozen
`GEV_REALTIME_TOOLS` pin **re-cut** rather than resolved to a side — both sides
recorded a length and sha over the same 31,104-byte baseline and the two edits
are disjoint (upstream's `esri-imagery` +85 B, this branch's ocean layers
+562 B), so neither value could survive.

## Known limits

- **Neither tier resolves Stokes drift**, and only the HYCOM tier carries tides —
  on the altimetry fallback there is no tidal signal at all. The legend states
  the physics of whichever source served rather than describing the tier once.
- **The drift model is not coupled to the current field**, deliberately. A
  single-hour observed snapshot cannot honestly force a 24–48 h integration, and
  I measured that even perfect HF forcing roughly doubles the structured spread
  against a cloud that is 98.1% leeway uncertainty. That is a research decision,
  not missing wiring.
- **The mask is 1/8°** (~14 km at the equator). That is why the third state
  exists and why coastal cells are never trusted either way.
- **One leeway class** (PIW-1). The USCG taxonomy has many more — life raft,
  kayak — and the coefficient table is already structured for them.

## If the size is the problem

The seams, in dependency order: (1) the NDBC layer and its proxies; (2) the GSHHG
mask; (3) the drift Monte Carlo, which needs (2) for beaching; (4) the current
field, which needs (2) for water-relative coverage. (4) is the largest and the
most novel and would get the better review on its own — happy to peel it off into
a follow-up PR against this one, or to restage entirely. Just say which.


bilawalsidhu:
Really like the animated ocean currents direction. Could you split this into smaller PRs along the boundaries you suggested and rebase against current main? We'd like to review the current-field visualization with its necessary land/sea mask and data dependencies, with buoy observations and the drift simulation kept separately reviewable. Stacked PRs are fine where there are dependencies.

A short screen recording of the currents at globe, regional, and coastal scales would help us assess the visual experience too. Thanks for putting all this work into it.

bilawalsidhu:
Following up on the split @MarioDanielPanuco: the current field is the part we most want. Since this opened, main has an animated wind layer (GFS/IFS particles in `src/layers/wind/`), so the shortest path is:

1. A first PR with only the GSHHG land/sea mask: the data under `src/data/local_data/gshhg_mask/` with its README, plus the codec, loader and build script.
2. A second PR that serves the HYCOM/CoastWatch current field in the same manifest + `.bin` grid shape as wind and draws it through the wind renderer rather than a separate canvas. The renderer would need a no-data value, regional (non-global) grids and a per-field speed scale; your median-based time compression is a good basis for the last one.
3. Buoys and drift after that.

Please also drop the `o`/`n` share tokens (every letter is taken, #658) and use a streaming capped read for the three field fetchers. If we don't hear back in a couple of weeks, we'll take the mask ourselves, with credit.

PR #102, opened 2026-08-29
fix(installations): derive a point from bounds when Overpass omits center

Schnix84:
## The bug

`militaryInstallationsProxy` asks Overpass for `out center tags geom`. Overpass honours only the **last** geometry mode, so `geom` wins and `center` is never emitted. Every way and relation therefore arrives without `lat`/`lon`, and `pointFrom` drops it.

The layer has been rendering nodes and nothing else:

| Area | Kept |
|---|---|
| San Diego | 149 of 228 elements — every way and relation gone |
| Warendorf (DE) | 0 — the town has no surveyed nodes, so the screen is empty over a mapped Bundeswehr barracks |

Silent: no error, no warning, just missing installations.

## The fix

`out geom` ships `bounds` alongside exactly those elements — relations included, which carry no `geometry` at all — so `pointFrom` falls back to the midpoint of that box. No change to the query, no change to the footprint path.

## Tests

Four new cases in `src/data/militaryInstallationData.test.mjs`. Two fail on the unfixed normalizer (way and relation carrying only `bounds`); two are regression guards — an explicit `center` must keep winning, and a half-filled or out-of-range box must be dropped rather than averaged into a plausible-looking lie.

## Verification

`npm run build`, `npm test` (2591 passing) and `npm run test:track` (100/100) green.

### DIFF
--- a/src/data/militaryInstallationData.js
+++ b/src/data/militaryInstallationData.js
@@ -54,7 +54,27 @@ function finiteLongitude(value) {
 function pointFrom(element) {
   const lat = Number(element?.lat ?? element?.center?.lat);
   const longitude = Number(element?.lon ?? element?.center?.lon);
-  return finiteLatitude(lat) && finiteLongitude(longitude) ? { latitude: lat, longitude } : null;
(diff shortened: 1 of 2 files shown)


bilawalsidhu:

PR #103, opened 2026-08-29
fix(qa): poll waits on a timer so qa-cctv-v2 can boot the app under SwiftShader

Schnix84:
## The bug

Every `waitForFunction` in `scripts/qa-cctv-v2.mjs` used Puppeteer's default `raf` polling. The app idles behind the explicit render governor, and under SwiftShader the rAF polling loop competes with the render loop the app needs to finish initialising.

Measured on an M-series Mac, headless:

| Polling | Result |
|---|---|
| `raf` (default) | boot predicate still `false` after 45 s |
| `polling: 200` | ready in 0.2 s |

So the harness timed out at its 60 s boot wait and never ran a single check — it was not reporting failures, it was not running.

## The fix

`polling: 200` on all four waits, plus the `protocolTimeout: 300_000` hardening `scripts/track-regression.mjs` already carries with the same reasoning:

> This machine intermittently stalls single CDP calls past the 180 s default under SwiftShader load.

Four lines of behaviour change, no test logic touched.

## Verification

The harness now boots and runs. It still cannot complete headless on this hardware (a later `Runtime.callFunctionOn` stalls out), but it reaches 30+ checks instead of zero. Headful it runs to completion.

### DIFF
--- a/scripts/qa-cctv-v2.mjs
+++ b/scripts/qa-cctv-v2.mjs
@@ -282,7 +282,7 @@ function waitForTilesLoaded(page, timeoutMs = 15000) {
       }
       return true; // no tileset in the scene — nothing to wait for
     },
-    { timeout: timeoutMs }
+    { timeout: timeoutMs, polling: 200 }
   ).then(() => true).catch(() => false);
 }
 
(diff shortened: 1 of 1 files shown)


bilawalsidhu:

PR #104, opened 2026-08-29
feat(overpass): allow extra mirrors via OVERPASS_EXTRA_UPSTREAMS

Schnix84:
## Why

The built-in mirror list is not reachable from everywhere. From one German consumer connection every entry failed at once — `overpass-api.de` and `lz4` refused TCP outright, and `overpass.kumi.systems` is a CNAME onto `private.coffee`, so the list holds fewer distinct hosts than it looks like.

Traffic, mapped installations and every other Overpass-backed layer were dead, with no fix short of editing `vite.config.js`. A network-local mirror belongs in the operator's environment, not in the repo's default list.

## What

`OVERPASS_EXTRA_UPSTREAMS` (whitespace- or comma-separated) is tried **before** the built-ins: a mirror someone configured deliberately is a better first try than one that just timed out for them, and each dead built-in costs a full `OVERPASS_TIMEOUT_MS` before the chain moves on.

Resolved lazily and re-resolved when the value changes, because module evaluation runs before Vite's `loadEnv()` copies `.env` into `process.env` — the file already documents that trap elsewhere.

Validation is **structural only**. Unlike a world-editable OSM tag, this value comes from the operator's own env, so a private or localhost address is a legitimate self-hosted instance rather than an SSRF target, and is deliberately allowed. The JSDoc says so, so it does not get "hardened" later by mistake.

One warning carried over from the field: **only add planet-wide instances.** A regional extract such as `overpass.osm.ch` answers fast with HTTP 200 but serves Switzerland only, so every query outside CH returns an empty element list — which this proxy then caches as a valid result for 24 h. It fails silently, not loudly.

## Tests

`src/data/overpassUpstreams.test.mjs` — separator forms and order, private/localhost allowed, unusable entries dropped rather than passed to the fetch loop, duplicates collapsed.

## Verification

`npm run build`, `npm test` (2591 passing) and `npm run test:track` (100/100) green. Confirmed in the running dev server: `[Overpass] 1 extra upstream(s) from OVERPASS_EXTRA_UPSTREAMS, tried first`.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -117,6 +117,13 @@ AISSTREAM_API_KEY=
 # Over the cap the proxy serves cached/stale tiles instead of hitting upstream.
 # TOMTOM_DAILY_TILE_BUDGET=40000
 
+# Optional: extra Overpass mirrors, tried BEFORE the built-in list. Set this if
+# the built-in mirrors are blocked or rate-limited from your network — every
+# Overpass-backed layer (traffic, mapped installations, OSM webcams) depends on
+# them. Whitespace- or comma-separated. Planet-wide instances only: a regional
+# extract answers HTTP 200 with an empty result, which gets cached for 24 h.
+# OVERPASS_EXTRA_UPSTREAMS=
+
 # Optional: CCTV layer tuning (advanced). Defaults are sensible — leave unset
 # unless you're customizing the camera source pack.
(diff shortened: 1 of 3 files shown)


bilawalsidhu:

PR #105, opened 2026-08-29
feat(cctv): add a keyless OSM webcam source pack

Schnix84:
## What

A fourth live pack alongside Austin, Caltrans and TfL, reachable in any country via `CCTV_OSM_AREA` (ISO 3166-1 alpha-2, default `DE`). `man_made=surveillance` + `contact:webcam` is the operator's **own published frame URL**, so fetching it is the same act as fetching an Austin or a TfL frame. No camera data is committed — the catalog is fetched at runtime, per CONTRIBUTING rule 4.

## Why it earns a place next to three traffic-camera packs

`camera:direction`. Austin and TfL have no heading signal at all and fall back to `fallbackHeadingFromId()` — an id hash, a number unrelated to where the camera looks. Measured for `DE` on 2026-08-28:

| | Nodes |
|---|---|
| `man_made=surveillance` | 46,129 |
| …with `camera:direction` | 20,771 |
| …with `contact:webcam` | 851 |
| …surviving the still-image filter | 279 |
| …of the 150 served, with a real mapped heading | 136 |

## What survives the filter, and why

- **Still images only.** Most `contact:webcam` values are landing pages. Reading a frame off those would be scraping.
- **No MJPEG.** `axis-cgi/mjpg/video.cgi` answers with an unbounded multipart stream; declared as an `image` feed it would not fail fast, it would hold the frame proxy to its own timeout every refresh tick.
- **Public hosts only.** This catalog is world-editable — anyone can point `contact:webcam` at a link-local metadata address — so the same non-global refusal `publicRadioHttpsUrl` applies to Radio Browser streams applies here. Plain `http:` is allowed, because municipal webcams are routinely TLS-less.

Roughly half the surviving URLs are dead at any moment; municipal webcams rot and OSM keeps the old URL. That is health's job, not the loader's — the first failed frame marks the camera `degraded`, exactly as for an offline Caltrans camera. No liveness sweep at load, which would be a per-refresh cost no other pack pays.

## Two supporting changes this pack forces

- **`fetchOverpassPayload` takes a per-caller `timeoutMs`.** The country-wide scan measured ~51 s on the fastest reachable mirror, so under the shared 22 s *interactive* budget it could never succeed. The pack also refreshes on its own 6 h clock and **never on the request path** — awaiting it inside `refreshCctvSources` would hang the first `/api/cctv/sources` after every TTL expiry for a minute, for every client.
- **The catalog cap goes 900 → 1100.** Four packs default to 950, and the cap slices from the *front* of the merged list, whose tail is the file/env pack — so 900 silently dropped hand-authored cameras rather than trimming a live pack. Stays under `HEALTH_MAX_ENTRIES` (1200).

## Licensing

ODbL covers the **catalog** (position, direction, the URL). It never covers the picture — each frame stays the publishing operator's under their own terms. `DATA_SOURCES.md` says exactly that.

## Known limitation

`OSM_WEBCAM_ANCHORS` is a fixed list of five German metros used to pick which cameras survive the per-pack cap. Set `CCTV_OSM_AREA=FR` and find more than 150, and you get the ones nearest Germany. Happy to make the anchors derive from the area, or to skip distance prioritisation entirely for a non-default area — say which you prefer.

The deeper version of the same point: like the other three packs this one is region-pinned at boot. Making it viewport-driven (as the mapped-installations layer already is) would remove the question entirely, but that changes the CCTV registry contract for all four packs and does not belong in this PR.

## Tests

`src/data/osmWebcamSources.test.mjs`, 13 cases: URL filter accept/reject including MJPEG and landing pages, the SSRF refusals, `camera:direction` wrapping (`-195 → 165` is live OSM data), mount priors, the drop rules, and that the licence names catalog and frame separately.

## Verification

`npm run build`, `npm test` (2600 passing) and `npm run test:track` (100/100) green. In the running app: catalog 803 → 953, 150 OSM cameras of which 136 carry a mapped heading, frames served as `X-CCTV-Source: upstream-image`.


Schnix84:
Follow-up context on the known limitation above (the five German metro anchors), so it does not hold up review of this PR.

I measured what a worldwide scan would actually yield — chunked by region, since a single global query gets a 504 from the mirror gateway at ~46 s regardless of the declared `[timeout:270]`:

| Region | tagged | usable |
|---|---|---|
| Europe West | 2000 (hit my own per-query cap — real number higher) | 920 |
| Europe East | 928 | 336 |
| Nordics | 468 | 294 |
| Oceania | 535 | 458 |
| North America East | 1092 | 228 |
| Africa + Asia | 146 | 9 |
| **Total** | **5169** | **2245** (1986 unique) |

66% carry a mapped `camera:direction`. Three regions returned 504 (North America West, South America, polar north), so worldwide is on the order of **~3000 usable cameras** — roughly 3× the current catalog, not a different order of magnitude.

**Why that is not this PR:** more cameras only help once the client stops draining the whole catalog. Measured in the running app with 924 cameras:

- 71 Cesium entities for the entire catalog — coverage geometry is already visible-set-only via `materializeCctvVisibleCoverageEntities`, so that part is fine
- `loading.loaded` reaches `total` at t=40 s, but `loading.active` only goes false at t=120 s — the tiles-ready pass re-enqueues the catalog while the counter is capped at `total`
- the real per-camera cost in the drain is `warmGroundFloor([point])` inside `updateRecordGeometry`, i.e. ~924 terrain resolutions

That second point is also, I believe, why `qa-cctv-v2`'s "geometry-load queue drains within Ns" assertion fails on a large catalog: the harness is right and the app's loading flag is wrong.

So the sequence I would suggest is: this PR as it stands → a viewport-scoped `/api/cctv/sources` (server keeps the full registry, client receives and drains only what is near the camera, so the frame proxy's registry contract and its SSRF hardening stay untouched) → worldwide chunks, which then also **deletes `OSM_WEBCAM_ANCHORS` entirely**, because with viewport scoping the per-pack cap no longer decides what a viewer sees.

Happy to open the viewport-scoping PR first if you would rather review them in that order — it stands on its own, since it cuts today's 803-camera drain just as much.

PR #108, opened 2026-08-29
fix: Windows CRLF test portability in reasonableDefaults.test.mjs (#88)

### LINKED ISSUE #88: test: reasonable Defaults test fails on Windows CRLF checkouts
Hi @bilawalsidhu,

  #### Reproduction

  #### Current result

  - Tests: 9
  - Passed: 8
  - Failed: 1

  Failing test:

  #### Environment

  - OS: Windows amd64
  - Node.js: 24.14.1
  - Git core.autocrlf: true

  #### Expected behavior

  #### Possible fix

Tom-Neverwinter:
Closes #88.

`node --test src/reasonableDefaults.test.mjs` reads `src/ui.js` with a hardcoded LF-only source anchor. On a Windows checkout with `core.autocrlf=true`, `ui.js` uses CRLF line endings, so the anchor is never found — 8/9 tests pass instead of 9/9.

## Fix

Normalize CRLF to LF before the anchor search, so the result doesn't depend on git's line-ending config.

## Verification

Ran on an actual Windows/CRLF checkout: 9/9 pass (was 8/9).

### DIFF
--- a/src/reasonableDefaults.test.mjs
+++ b/src/reasonableDefaults.test.mjs
@@ -38,9 +38,14 @@ import {
 } from './scopeMask.js';
 import { ShareLinkManager } from './sharelink.js';
 
-const uiSource = fs.readFileSync(new URL('./ui.js', import.meta.url), 'utf8');
-const indexHtml = fs.readFileSync(new URL('../index.html', import.meta.url), 'utf8');
-const shareSource = fs.readFileSync(new URL('./sharelink.js', import.meta.url), 'utf8');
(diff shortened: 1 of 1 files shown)


bilawalsidhu:
Superseded by the .gitattributes line-ending normalization already on main (4275f12). Full status for your open PRs: https://github.com/bilawalsidhu/gods-eye-view/pull/171#issuecomment-5612644797

bilawalsidhu:
Line endings are normalized repo-wide by .gitattributes on main. Closing as superseded.

PR #113, opened 2026-08-29
feat: live weather radar / satellite overlay layer via RainViewer (#85)

Tom-Neverwinter:
Addresses #85. **Note:** aware of #89, an independent WIP PoC for this same issue also using RainViewer — didn't coordinate in advance since we were working in parallel. Happy to close whichever the maintainer doesn't prefer; leaving both open for now so you can compare.

## What this adds

New toggleable "Weather Radar" data layer, two independently-toggled overlays (radar precip reflectivity, IR satellite cloud cover), drawn as semi-transparent `Cesium.ImageryLayer`s on the globe surface.

**Source:** RainViewer's public API (`api.rainviewer.com/public/weather-maps.json` for the frame list, tiles from `tilecache.rainviewer.com`) — keyless, CORS-open (verified live: both send `access-control-allow-origin: *`), fetched directly from the browser, no server proxy needed. Polls every 10 min (RainViewer's own frame cadence); a same-frame poll is a no-op rather than tearing down and rebuilding a live imagery layer for nothing.

**Scope:** just the latest radar and satellite frames from this one free endpoint. The full #85 proposal also named region-specific sources (NOAA MRMS, EUMETSAT) and an animated multi-frame time slider — deliberately out of scope for this pass. RainViewer's `radar.past` array already carries ~2h of frame history if a future pass wants to add animation on the same tile-URL-building code. Satellite defaults off (RainViewer's free satellite feed was observed empty during testing) so the chip doesn't default to showing nothing; radar (reliably populated) defaults on.

Wired into `layerState.js`'s share-link/persistence registry and `dataCredits.js` / `DATA_SOURCES.md` for attribution.

## Testing

Added `src/data/weatherOverlay.test.mjs` (8 tests: frame-picking, response parsing against a real captured RainViewer payload, tile URL building against a live-verified tile fetch, param/visibility helpers). Full test suite passes (no regressions vs. `main`), production build succeeds with no new warnings.


bilawalsidhu:
Superseded by the weather layer now in flight. Full status for your open PRs: https://github.com/bilawalsidhu/gods-eye-view/pull/171#issuecomment-5612644797

bilawalsidhu:
The weather radar layer is landing through the next maintainer batch. Closing as superseded.

PR #114, opened 2026-08-29
feat: OpenSeaMap / OpenSnowMap overlay layers (#10)

Tom-Neverwinter:
Addresses part of #10 (OpenRailwayMap/OpenInfraMap deliberately out of scope — see below).

Two independently toggleable open-data raster overlays — OpenSeaMap sea marks (buoys, lighthouses, ports) and OpenSnowMap ski pistes — drawn as `Cesium.ImageryLayer`s directly on the Cesium globe surface.

Visible on the `osm`/`bing-aerial`/`bing-labels` map stacks and invisible (not broken — the layer's stats say so in plain text) on the default `photoreal` stack, since Google Photorealistic 3D Tiles render with the globe hidden.

**OpenRailwayMap and OpenInfraMap, also requested in #10, are deliberately NOT included in this pass:**
- OpenRailwayMap's public tile server enforces a Referer allowlist scoped to its own site (confirmed by direct request — every other Referer, including none, 403s).
- OpenInfraMap serves MapLibre GL vector tiles, not raster XYZ — that would need a genuine vector-tile rendering pipeline, not a `Cesium.ImageryLayer` drop-in.

Both worth their own dedicated follow-up pass.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #115, opened 2026-08-29
feat: toggle to route Radio audio into the Voice Model (#52)

### LINKED ISSUE #52: Feature request: Toggle to enable sending any chosen radio output to the Voice Model so it can answer questions about it

Tom-Neverwinter:
Closes #52. Stacked on #114 (OpenSeaMap/OpenSnowMap overlays) for branch history reasons only — no functional dependency between the two.

A dock button (📻, next to the voice-tier button) lets the voice model hear whatever Radio station is currently tuned in ("what did they just say", "what's playing"). Off by default, opt-in only — this is the one voice setting that sends third-party broadcast audio to the realtime API, never inferred from Radio simply being on.

**Mechanism:** `radio.js` arms `crossOrigin='anonymous'` on newly-installed `<audio>` elements (never mutated on a live element, which can't retroactively fix already-buffered bytes); `gevRealtime.js` taps the element via `AudioContext.createMediaElementSource` and mixes it with the mic through a `MediaStreamAudioDestinationNode`, swapped onto the outbound WebRTC track via `RTCRtpSender.replaceTrack` (no renegotiation, instant toggle mid-session). CORS verified live against 8 real Radio Browser stations — 7/8 already send permissive headers; the toggle reports honest live per-station capture status rather than just "on".

**Includes a same-day follow-up fix:** the speaker-tap node for the previous station wasn't disconnected from `context.destination` on a station change, so per the Web Audio spec the audio graph kept each abandoned node+element pair alive indefinitely — one permanent leak per station change in a long session with the toggle on. Fixed with a regression test using a minimal fake Web Audio graph.


bilawalsidhu:

Issue #57, opened 2026-08-25
[Feature Request] Local RTL-SDR / dump1090 receiver tap layer


Tom-Neverwinter:
Add a local-only data layer that connects to a user's own local dump1090/readsb ADS-B receiver (RTL-SDR dongle), typically exposed at http://localhost:8080/data/aircraft.json (or a user-configurable host/port), and shows exactly what that receiver is hearing on the globe — independent of the existing public-feed Flights layer.

Why separate from the existing Flights layer: Flights already shows public aggregated ADS-B (one polling interval behind, aggregator coverage). This layer shows only what the user's own antenna actually receives right now — useful for verifying receiver performance/range and seeing genuinely first-hand local data with zero third-party aggregation.

Scope:
- User-configurable local receiver base URL (default to the common dump1090-fa/readsb default of http://localhost:8080).
- Poll aircraft.json (or tar1090-compatible equivalent) on a short interval.
- Plot received aircraft on the globe, visually distinguished from the public Flights layer.
- Off by default (opt-in local-network layer).

Filed to track implementation.

Lob26:
Separate layer is the right call, and for a reason worth writing down: the purpose is verifying what *your* antenna hears, and the moment that data is blended with an aggregator you can no longer answer "did my receiver get this?". So nothing below proposes merging the two by default.

What the issue leaves open is **overlap**: with both layers on, the same airframe is on the globe twice, at two positions, from two clocks. That's the one design decision here, and it's worth picking deliberately rather than finding it in review.

**Three defensible answers.** (1) *Draw both, always* — the issue's "visually distinguished". Honest and simplest; the cost is that in a dense receiver's coverage two markers for one plane at slightly different positions reads as a bug to anyone not thinking about provenance. (2) *Local wins where local has it* — Flights suppresses any ICAO the receiver is currently reporting with a fresh position, so one marker per airframe and the receiver's coverage becomes visible as the region where provenance switches. (3) *Local wins, and says so* — (2) plus a provenance field on the popup.

I'd suggest (1) as the default with (3) behind a toggle: keeps the layer independent as filed, and gives the dense-coverage case a way out. Worth flagging that (2)/(3) would be the first place in the tree where two sources are concurrently authoritative for the same object — OpenSky → adsb.lol is strict failover, one source at a time, and `src/data/adsbLolFallback.js` is a unit adapter rather than a merge. That's a precedent, not a pattern to follow.

[...]

PR #116, opened 2026-08-29
feat: Rayhunter (EFF cell-site-simulator detector) tap layer (#56)

### LINKED ISSUE #56: [Feature Request] Rayhunter (EFF) cellular IMSI-catcher tap layer
Add a local-only data layer that connects to a locally-running EFF Rayhunter instance (https://github.com/EFForg/rayhunter, typically flashed onto an Orbic mobile hotspot at its default local address/port) and surfaces its IMSI-catcher / cellular-anomaly warnings on the globe at the user's current location.

Rayhunter has no GPS of its own — it just flags suspicious cellular behavior seen by the hotspot it runs on. So this is inherently a **personal, opt-in, local-network** layer: the user points the app at their own Rayhunter's local API, and detected warnings get plotted at wherever the user currently is (or was) when the warning fired, not at any public/global endpoint.

Scope:
- User-configurable local Rayhunter base URL (default to the project's documented default address/port).
- Poll its API for warnings/analysis results; exact endpoint shape should come from the real EFForg/rayhunter API docs/source, not guessed.
- Plot warnings on the globe at the user's live location with severity/detail in the click card.
- Off by default (opt-in local-network layer), like other locally-sourced data.

Filed to track implementation.

Tom-Neverwinter:
Closes #56. Stacked on #115 for branch history reasons only — no functional dependency.

[Rayhunter](https://github.com/EFForg/rayhunter) runs on a device the user owns — typically a flashed Orbic mobile hotspot — and flags suspicious cellular behavior it observes. Personal, opt-in, local-network layer: polls the user's own Rayhunter over the same-origin `/api/rayhunter/*` proxy (the device's embedded web server sends no CORS headers, so the browser can't reach it directly) and plots newly-seen warnings at wherever the browser currently is when the warning is first observed — not the device's actual location at detection time, which this app has no way to know. That approximation is surfaced honestly via `getStats().error` rather than presented as precise.

API shape verified against EFForg/rayhunter's real source (route table + stats/diag modules), not guessed.

Includes two same-day code-review fixups (bugs found and fixed before this reached `main`, not after).


bilawalsidhu:

Tom-Neverwinter:
Rebuilt this on current main and on the shared tap contract, per your sequencing note. **Stacks on #382** — the first four commits here are that PR; review it first and this one shrinks to the last commit.

This replaces the original branch entirely: `vite.config.js` went from 7,995 lines to 3 while this sat open, so the proxy half had to be re-authored rather than rebased. It was also stacked on the OpenSeaMap and Radio→Voice work; that's untangled now and this is one tap on its own, as you asked.

## What it is

Rayhunter runs on a device the user owns — a flashed Orbic hotspot, reached over USB tether or its own Wi-Fi — and flags suspicious cellular behaviour. Its embedded web server sends no CORS headers, so the browser can't read it directly.

`server/providers/rayhunter.js` relays two read-only endpoints same-origin, registered in `local.js` like every other provider. GET only; nothing is ever written to the device.

## What changed versus the version you last saw

- **The address check is no longer ad-hoc.** The old `validBase()` accepted any host and blocklisted exactly one address, with a comment arguing the tradeoff. That's now `parseTapAddress()` from #382: loopback, RFC1918, or `.local`, and a public or malformed address is a 400 with **no upstream request made at all**. Both real access paths are private addresses, so this costs nothing in practice — I checked the branch's own default `192.168.1.1:8080` along with USB-tether, hotspot and iOS `172.20.10.x` ranges.
- **The client's hand-synced copy of the regex is gone.** It carried a comment saying it was kept in sync by hand across the trust boundary. Both sides now call the same function, which is why #382 puts the module in `src/data/` rather than under `server/providers/` — #274 is open about eight copies of haversine and I didn't want to add a ninth pattern.
- **Rate limit**, 120/min per client with a 600 global backstop. Not a quota guard, since the upstream is the user's own hardware — it stops a runaway client loop from wedging a small embedded web server.
- **8 MB streaming body cap, and redirects refused** rather than followed, so a device response can't walk the request outside the address that was just validated.
- **Upstream error text is no longer relayed to the browser** — logged locally, client is told only that the device is unreachable.

## Fixtures — the part I'd like you to look at

CI has no Rayhunter, so the proxy takes an injectable `fetchImpl` and the tests drive it from `src/data/fixtures/`. That matched an existing convention rather than needing a new harness, which is why #382 doesn't ship one.

**These fixtures are synthetic, written to the API shape in EFForg/rayhunter — not device captures, and they're labelled that way in the fixtures README.** I didn't want to put something in a file whose README documents provenance and quietly imply it came off hardware. They do exercise the cases the parser has to survive: the leading metadata line with no `events` array, `null` entries in the positional per-analyzer array, an `Informational` event that must be filtered, one of each real severity, and a trailing malformed line. A real capture would be strictly better and would need no test changes — I have the hardware and can swap one in if you'd rather.

## Honest limitation, unchanged

Warnings are plotted at **the browser's position when the warning is first observed, not the device's position at detection time** — the app has no way to know the latter. That's stated in `getStats().error`, `DATA_SOURCES.md` and the CURRENT-STATE section rather than presented as a precise fix.

## Also here

`DataLayerManager` grows a generic `chip.prompt` path so a layer row can carry a free-text setting instead of a static `params` object. The device address is its only consumer today, so I kept it here rather than in the contract PR; if #57 goes server-side it's already there.

## Verification

`npm test` 3000 pass / 0 fail with `GEV_REQUIRE_ALLOCATION_GATE=1` on Node 24.19.0, `npm run build` clean, `npm run test:track` 108 passed / 0 failed, `format:check` and `check:boundaries` clean.

11 new proxy tests cover the relay path, the SSRF refusals (public, link-local, hostname) asserting no upstream call is made, method gating, traversal-shaped recording names, unreachable devices, device HTTP errors, the rate limiter, redirect refusal, and the body cap. `DATA_SOURCES.md` and `docs/CURRENT-STATE.md` are updated in the same commit.

> 🤖 Generated with [Claude Code](https://claude.com/claude-code)
>
> https://claude.ai/code/session_01KGKBNNjPh4iQe8sviGZNPb

PR #118, opened 2026-08-29
fix: never frame an arrival below ground at elevation

Mkhanal04:
## Repro

Fly to a city well above sea level: Denver (~1,609 m) or Mexico City (~2,240 m). The camera lands under the surface and the frame renders black or from inside the terrain.

At Austin (~150 m) the same error is only a few metres, so it reads as a slightly low camera angle rather than a bug. That is probably why it hasn't come up.

## Cause

Arrival framing runs before the destination's tiles exist, so it has to predict the ground height. All three available predictions are unreliable at that point:

- `globe.getHeight()` is dead here, since the globe sits under the photorealistic tileset with no terrain provider installed.
- The `groundElevation` presets only exist for the cities in `CITY_POIS`.
- The DEM resolves async. Measured cold at one test site it came back about 134 m below the rendered surface, usually after the flight had finished.

So `flyToLandmark` fell back to `groundElevation`, which defaults to sea level. Anything outside the curated city list got framed as if it stood on the ocean.

## Fix

Prediction alone can't solve this, so this does two things.

**Better prediction.** Consult the ground-floor elevation service already in the tree (`src/data/groundFloor.js`, same datum) before falling back to zero. Applied on the coordinate path and the search path, and the cell is warmed so the next nearby arrival is already correct. The search path resolves elevation before the `mayFly()` gate so a cancellation still wins the race.

**Then measurement.** `src/cameraGroundGuard.js` samples the real surface after the flight lands, once tiles have streamed in and `scene.sampleHeight` can be trusted, and lifts the eye if it sits below `MIN_EYE_CLEARANCE_M` (120 m).

The guard stays out of the way. It stops on a clear view, on exhausted attempts, or as soon as the user touches the controls, since a low angle someone chose is better than a correction under their hands. The lift holds heading and pitch so it reads as the shot settling rather than a second flight. Sub-metre differences are ignored.

`groundClearanceDeficitM` is a pure function, so the clearance policy is unit-tested without a scene.

## Verification

- `npm test`: 2,593 pass, 0 fail. That is up 6, which are this PR's tests. The 2 allocation microbenchmarks self-skip on Node 26, same as on `main`.
- `npm run build`: clean.
- `npm run test:track`: 100 passed, 0 failed, 0 skipped, including the no-console-errors and no-5xx checks across the full run.
- Checked by hand at elevation: black frame before, arrives above the surface after.

Node 26.7.0.

## Scope

Limited to the ground-framing bug. No new dependencies: `probeMeshFloorM` and the ground-floor service are both already in the tree and unmodified. Other work from the same branch (time-of-day control, route duration caps) is held back for separate PRs.


Mkhanal04:
Before and after, same coordinates, same camera parameters, driven through `flyToLandmark` at Denver (39.7392, -104.9903). The only difference is the fix.

**Before** (current `main`). Black frame, `ALT: 470M` against ~1,609 m ground, so the eye is about 1,156 m underground. The HUD cannot even resolve the place, it just reads `UNKNOWN LOCATION`.

Denver arrival before the fix

**After** (this PR). `ALT: 1721M`, and the HUD resolves to `DENVER DELAWARE STREET NEAR HANK`.

Denver arrival after the fix

Both captured headless at 1600x1000 against a local dev server, one running `main` and one running this branch.

PR #121, opened 2026-08-29
fix(packaging): normalize line endings so Windows CRLF checkouts pass tests

### LINKED ISSUE #88: test: reasonable Defaults test fails on Windows CRLF checkouts

yunaremaia:
## Summary
Fixes #88. On machines with `core.autocrlf=true` (Git for Windows default), tracked text files materialize as CRLF on checkout while the index stores LF. This makes `git status` report a clean tree but breaks byte-exact tests such as the "reasonable Defaults" check.

- Add `* text=auto eol=lf` so Git normalizes line endings on checkout
- Working tree now matches the index regardless of host platform
- No source changes; `.gitattributes` only

## Verification
- `git check-attr text <file>` returns `auto`
- Build/test unaffected on Linux; resolves the Windows CRLF divergence

Fixes #88


yunaremaia:
Ready for review. Normalizes line endings via  so tests pass on Windows CRLF checkouts (fixes #88). Mergeable.

yunaremaia:
Resolved the merge conflict. The upstream already added a `.gitattributes` with more detailed comments, so I kept theirs. package.json and package-lock.json were resolved with the latest upstream changes.

Note: this PR is now redundant with upstream's `.gitattributes` — the upstream version already normalizes line endings identically. Should I close this PR in favor of upstream's fix, or keep it for the package.json/package-lock.json changes?

yunaremaia:
Hi @bilawalsidhu — this PR is mergeable (normalize line endings for Windows CRLF). Could you review when you have a moment?

yunaremaia:
Closing as redundant — upstream already applied the .gitattributes fix independently. No further action needed.

PR #128, opened 2026-08-30
Feature: Integrated Gemini Live Voice Agent & Dynamic Spatial Tools

gastondana627:
Added Gemini Multimodal Live bidirectional voice control.

Built a sleek bottom-left activity meter specifically for Gemini.

Implemented an RMS noise gate / audio buffer so the mic doesn't clip.

Upgraded the analyst_query to support satellites and added a custom adjust_camera_altitude tool so the agent can fly straight down to Earth

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -57,6 +57,11 @@ OPENAI_HUD_SUMMARY_MODEL=gpt-5-nano
 # provider-side budget: OpenAI platform -> Settings -> Limits (usage limits).
 # GEV_RATELIMIT_OPENAI_PER_MIN=30
 
+# Optional: Gemini Live voice control.
+GEMINI_API_KEY=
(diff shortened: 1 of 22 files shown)


OakSwingZZZ:
Yes please!

bilawalsidhu:
This sends the Gemini key to the browser, which the project can't ship. #157 shows the server-side shape; rework along those lines and reopen.

gastondana627:
Hey Bilawal, appreciate the review on #128! Honestly, I was just spinning this up as a quick 1-day local sandbox test to see how fast we could get the live voice agent and spatial tools talking, and completely treated it like a speed-run rather than a production build. 

Totally see why shipping client-side SDK initialization straight to the browser is a non-starter for the main repo. I'll take a look at the server-side proxy structure in #157 and refactor it before reopening. GG on the security gate!

PR #129, opened 2026-08-30
Add filterable AR experience data layer

bfrager:
## Summary

Adds a filterable AR Experiences data layer that turns Geoverse, MeshMap, ARpoise, and configured Open AR Cloud/OSCP operators into normalized map hotspots. Each source remains independently selectable, expired experiences are controlled by a separate Past filter, and launch links retain provider attribution.

The implementation follows the approved AR content-layer plan while keeping Niantic represented as capability coverage rather than inventing a public experience catalog.

## What Problem This Solves

- Gives users one map layer for discovering nearby mobile or headset AR content without collapsing provider identity.
- Adds a same-origin provider gateway with secret isolation, response limits, timeouts, exact-viewport caching, stale fallback, rate limiting, and bounded federation.
- Persists and shares provider selections through the existing layer-state contract, including voice and scene-policy support.
- Keeps ARpoise disabled by default until its lightweight maintainer/rate-limit gate is cleared, and treats each OSCP operator as its own source.
- Documents provider configuration, security boundaries, data provenance, and rollout constraints.

## Verification Steps

- `npm test` — 2,624 passed, 0 failed.
- `npm run build` — passed; only the existing chunk-size and browser externalization warnings remain.
- `npm run test:track` — 100 passed, 0 failed on the final run.
- Browser QA — verified distinct Geoverse, MeshMap, ARpoise, and OSCP chips; Past filtering; provider-by-provider filtering; disabled-provider messaging; overlay attribution; and a clean AR console path.
- `git diff --check` — passed.

## Post-Deploy Monitoring & Validation

- For the first 24 hours after enabling each provider, the deployment owner should monitor `/api/ar-content` latency and status, provider `error`/`stale` responses, unexpected `429` responses, timeout/size-limit errors, and request volume versus cache hits.
- Healthy behavior is a stable aggregate `200`, configured providers reporting `ok`, independent source filters returning only their selected markers, and Geoverse responses surfacing `coverageLimited` when the requested viewport exceeds its effective radius.
- Provider flags and OSCP entries are the first rollback lever. Revert this change if the gateway or map layer causes broader runtime regressions.
- Live provider schema compatibility still needs validation with production credentials. ARpoise remains intentionally gated until maintainer and rate-limit expectations are confirmed.

---

[Compound Engineering](https://github.com/EveryInc/compound-engineering-plugin)


bilawalsidhu:
Thanks for the proposal. An AR experience catalog is outside the project's scope, so closing this.

PR #147, opened 2026-09-01
fix: UI tooltips: explain each visual preset, and name the missing key on dis...

### LINKED ISSUE #143: UI tooltips: explain each visual preset, and name the missing key on disabled controls
Two related polish items, originally raised by @shahab-cl in #90:

1. **Preset tooltips** — hovering CRT / NVG / FLIR / NOIR (and friends) should say in one line what the preset does, so first-run users don't have to trial-and-error the grid.
2. **Disabled controls name their key** — when a control is unavailable because a provider key is missing or invalid, its tooltip should say exactly which one (e.g. "Needs CESIUM_ION_TOKEN — add it in Provider Settings"), turning a dead button into a next step.

Both are frontend-only. The provider registry in `src/keySetupCore.mjs` already knows every key, what it unlocks, and whether it is configured, so the second item is mostly wiring that knowledge into the disabled-state tooltips.

yhay81:
Fixes #143.

## Summary

UI tooltips: explain each visual preset, and name the missing key on disabled controls

## Validation

- Mechanical gate: +38/-19, tests passed
- Adversarial review: approved

---
🤖 AI-authored PR, operated by @yhay81.


bilawalsidhu:

PR #148, opened 2026-09-01
fix(radio): keep the stream playing when the Radio layer is turned off

byslovskyf:
Turning the Radio layer off called `stopRadioPlayback`, so a station you were listening to went silent the moment you hid its globe markers. Presentation and audio are now decoupled — turning the view off never stops what is already on air.

**radio.js.** `disable()` no longer stops playback or tears down the `<audio>` element; it only hides the globe presentation (markers, tuner, click-to-select, in-flight directory fetch) and keeps the selected station. `destroy()` now releases the stream explicitly. A new `radioTransportAllowed()` gate — layer presentable OR a live stream — replaces the presentation-only gate on `play` / `togglePlayback` / `setVolume` / `selectStation` / `cycleStation`. While detached, selection and cycle skip the camera flight and marker work. `getUIState()` gains `audioDetached` / `transportActive`.

**ui.js.** The transport row, mini-player, and cockpit strip (play/pause, stop, next/prev, volume) stay live while a stream is on air with the layer hidden; the tuner, filter, and Enable toggle stay layer-gated. The Radio panel no longer auto-collapses while audio is detached, and the playback line notes "Radio layer hidden". Re-enabling the layer brings presentation back; hitting STOP re-gates everything to inert. No behaviour change when the layer is on.

STOP (button, or voice "stop") is still how you silence it.

**Tests.** `npm test` — 2738 pass. `npm run build` — green. New test: "disabling the Radio layer leaves a live stream audible and its transport controllable". `npm run test:track` not run (needs a live dev server); this change does not touch flight-tracking invariants.

**Note.** This branch was cut from local `main`, which is 4 commits ahead of `bilawalsidhu/main`, so the PR carries those 4 commits plus the radio fix (`7a93bd3`).


bilawalsidhu:
The radio fix is carried in the next maintainer batch with your commits and authorship intact. Closing here; the unrelated commits this branch picked up stay out.

PR #152, opened 2026-09-01
fix: recognize completed Pinokio setup

bilawalsidhu:
## What changed

- Resolve the successful-install marker from the nested Pinokio launcher directory, so a completed clean install exposes **Start** instead of returning to **Install**.
- Replace the source-pattern assertion with behavioral coverage for the Install, Start, Update, Repair, Starting, and Open states.
- Correct the keyless startup summary to name Esri World Imagery with keyless terrain and OpenStreetMap as the fallback.

## Verification

- Focused launcher and startup tests: 15/15
- Unit suite: 2675 passed, 0 failed, 1 intentional Node-version skip
- Production build: passed
- Tracking regression: 108/108
- Private CI: Node 24.14, Node 26.x, and Windows onboarding all green

Thanks to @manjunath22466 for the fix and real-device Pinokio validation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ of current runtime behavior, see [`docs/CURRENT-STATE.md`](docs/CURRENT-STATE.md
 
 ### Fixed
 
+- Pinokio now recognizes its nested successful-install marker, so a completed
+  one-click install exposes Start instead of returning to Install.
+- The keyless `dev-fresh.sh` startup summary now names Esri World Imagery with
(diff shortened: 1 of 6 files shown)


bilawalsidhu:
@manjunath22466 — the Pinokio lifecycle fix and real-device validation are now included in [v0.1.1](https://github.com/bilawalsidhu/gods-eye-view/releases/tag/v0.1.1). Clean installs now land on **Start** as intended, with your authored commit preserved on `main`. Thank you.

PR #161, opened 2026-09-03
fix: isolate ACL verify PowerShell from pwsh7-polluted PSModulePath

michaelhan1208:
## Summary

On Windows machines with a side-by-side **PowerShell 7** install, the credential-file hardening step (`hardenCredentialFile` in `src/keySetupHardening.mjs`) fails closed and refuses to write provider keys.

### Root cause

PowerShell 7 prepends its own module trees to `PSModulePath` at startup:
```
C:\Users\<user>\Documents\PowerShell\Modules
C:\Program Files\PowerShell\Modules
c:\program files\powershell\7\Modules
...
```
The Windows hardening path spawns the **system PowerShell 5.1** (`powershell.exe`) to verify the exact DACL of the credential file. That process inherited the polluted `PSModulePath` from the caller. When 5.1 tried to autoload `Get-Acl` (from `Microsoft.PowerShell.Security`), it hit the incompatible **7.x module manifest** first and failed with `CouldNotAutoloadMatchingModule`:

```
Get-Acl : 找不到模块"Microsoft.PowerShell.Security"中的 Get-Acl 命令...
```

`hardenCredentialFile` returns `false` → provider settings refuse to save any key on such machines. This is also a **security gap**: the verify interpreter must never load modules from an environment an untrusted caller can steer.

### Fix

Reset `PSModulePath` for the **verify subprocess only** to the 5.1 system module tree, derived from the already-validated `powershell.exe` path (`<systemRoot>\System32\WindowsPowerShell\v1.0\Modules`). This matches the file's existing fail-closed posture toward inheritable environment steering (the same reasoning that refuses `PATH`-shadowed tools and redirected system roots). No other environment is touched, and the surrounding `environment` spread is preserved.

### Test

New regression test `Windows hardening isolates the verify PowerShell from a pwsh7-polluted PSModulePath` simulates a polluted `PSModulePath` and asserts the verify subprocess receives only the 5.1 system module tree.

Verified locally on Windows:
- `src/keySetupHardening.test.mjs`: **13/13 pass**, including the real-DACL integration test (`Windows production hardener applies its exact DACL with native tools`), which previously failed on this machine.
- Full suite: **exit 0** (2677 tests).

### DIFF
--- a/src/keySetupHardening.mjs
+++ b/src/keySetupHardening.mjs
@@ -143,6 +143,19 @@ export function hardenCredentialFile(filepath, {
     // accept only three explicit FullControl allow principals, with inheritance
     // disabled. Any unexpected rule, right, command error, or missing principal
     // fails closed before the secret reaches disk.
+    //
+    // The verify process must load Microsoft.PowerShell.Security (Get-Acl)
+    // from the Windows PowerShell system module tree ONLY. A side-by-side
+    // PowerShell 7 install prepends its own module trees to PSModulePath at
+    // startup; inherited into a 5.1 process, the incompatible 7.x manifest
+    // cannot be autoloaded and the verify step fails. Reset PSModulePath to
+    // the 5.1 system tree derived from the already-validated executable path —
+    // the same fail-closed posture that refuses PATH-shadowed tools above.
+    // Nothing an inheriting environment prepends may steer Get-Acl.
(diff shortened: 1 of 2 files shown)


bilawalsidhu:
Landed via #508 (rebased onto the current component layout, with fixes from review). Thank you @michaelhan1208!

PR #163, opened 2026-09-03
Add macOS app launcher and startup API health roster

jameslowuvsb:
## Summary
- **macOS app icon.** `npm run app:mac -- --install` builds `God's Eye View.app` (icon rendered from `public/logo.svg` via the existing `sharp` devDependency + `iconutil`) into `~/Applications`. Double-click starts the dev server through `scripts/dev-fresh.sh`, waits for the port, and opens a chromeless Chrome window; if the server is already up it just opens the window. Failures surface as a macOS dialog with the log tail.
- **Startup API health roster.** New `GET /api/health` (NDJSON stream, `?format=json` for scripts, 60 s report cache) probes all 24 upstreams. The loading cover paints one row per service as probes settle, holds with a `CONTINUE ▸ 10` countdown whenever anything is DOWN / KEY REJECTED / DEGRADED, shows the reason inline, and re-surfaces failures in the top banner after reveal. Non-OK rows are logged server- and browser-side.
- `npm run health` prints the same table from a running server (exit 1 if a required service is not live).

## Cost / quota rules baked into the probes
- No billable Google call: the Maps key is checked for presence only and the tile host with a keyless HEAD; Places is reported as `KEY SET · not probed`.
- Launch Library 2: site-root HEAD only (15 anonymous API calls/hour budget preserved).
- OpenSky anonymous/basic: one tiny bounding-box query; OAuth mode reuses the server's token helper.
- AISStream: reads the dev server's own feed snapshot instead of opening a second websocket.
- FIRMS: uses the free `mapkey_status` endpoint (also reports transaction usage).

## Test plan
- [x] `npm test` — 2603/2603 (16 new tests: probe state matrix, no-Google-key rule, timeouts, NDJSON parsing, roster copy)
- [x] Live `npm run health --fresh` against a fully keyed `.env`: 24 live in ~3 s
- [x] Forged DOWN / KEY REJECTED / DEGRADED rows via request interception: cover holds, countdown, inline reasons, banner after reveal (screenshots checked)
- [x] Built and launched the `.app`; reopen-when-running path verified

Linux/Windows launchers are not included.


jameslowuvsb:
Opened against the wrong repo; closing.

PR #166, opened 2026-09-03
feat(search): keyless place search at all three geocode call sites, and never cache an outage as a miss

daikaginza:
Location search, map annotations and Radio place lookup now support keyless Photon/OpenStreetMap geocoding, with Google preferred when configured.

The contribution is adapted to an explicit browser place-search service. Standalone setup composes the Google and Photon adapters; consumers retain camera framing, nearby landmark recovery, annotation footprint matching and station selection. The portable `./search` export reads no application configuration and Google transport is supplied by its caller.

Follow-up fixes keep malformed replies/outages out of negative caches, propagate cancellation through retries and late results, handle Google transport failures, and omit invalid/wrapped Photon bounds. Existing key setup and reverse geocoding remain unchanged. The Capitol identity regression is preserved.

Validation: full unit suite, setup doctor, formatting, package boundaries and production build; real-browser Google Capitol search/outline and keyless Hanoi lake search/annotation. All three GitHub CI jobs and 108 final browser tracking checks pass. Keyless Radio lookup for Hoi An selected a station through Photon; a cancelled action remained inert. Microphone audio is outside this change; voice actions are exercised directly.

Original implementation and place-matching work by @daikaginza, with provider integration and follow-up fixes by @samehkhamis. Photon/OpenStreetMap attribution is included in the data credits and source documentation.


samehkhamis:
Thank you @daikaginza for the contribution! Merged with the provider integration and cancellation/cache fixes.

PR #167, opened 2026-09-03
feat(layers): add Civic Offices, a viewport layer that states its own limits

daikaginza:
## What it adds

A viewport data layer for **police stations, town halls and government offices** from OpenStreetMap, through the existing `/api/overpass` proxy, debounced on camera `moveEnd`. One self-contained module implementing the layer interface, following the pattern CONTRIBUTING points at.

Classification is by declared order, so a building tagged both `amenity=police` and `office=government` is a police station rather than whichever tag was read first.

## The part worth reviewing

This layer's gaps do not look like gaps. A missing aircraft is obviously missing; an unmapped district renders **exactly** like a district with no office, so a reader takes the absence for a fact about the world rather than about the data. The ground rule about not misrepresenting public-data inference as authoritative intelligence lands, for this layer, almost entirely on one line of text — the row's meta line.

So that line is built from measurements the response actually carries, not from a disclaimer:

- **`out count`** is requested alongside the features, so the row reads `400 of 437 in view` when the query cap truncated the set. A silently truncated list reads as a complete one.
- **`osm3s.timestamp_osm_base`** is the snapshot date of whichever mirror answered. They lag — two mirrors gave `2026-05-06` and `2026-06-01` during development — so the row shows the date it got, not today's.
- A response **without** a count reports no total rather than reusing the drawn length, which would turn "unknown" into "this many".
- Above a **2°** viewport the layer clears and says `zoom in to load — too wide to map`. `0 in view` over a continent is a false statement about the world, and leaving the previous city's offices drawn over it is worse.

Nothing is accumulated across viewports. The claim is always "in view", and the tests assert the string never contains *total*, *all*, *complete* or *every*.

## Rendering

Names go through `worldOverlay`, not a Cesium entity `label`. Entity labels were the first attempt and rendered nothing — no layer here uses them, `infoBox` is off, and all on-globe text is placed by the overlay's collision arbiter, so a label outside it has nothing keeping it off its neighbours.

## Verification

- `npm test` — **2698 tests, 0 failures** (12 in `src/data/civicOffices.test.mjs`)
- `npm run build` — clean
- CI on this branch: Node 24.14.0, Node 26.x and Windows onboarding all green
- Live in Chromium over Hanoi (`0.077° × 0.144°`): **76 offices**, row reads `OpenStreetMap · 76 in view · mapped to 2026-06-01`, labels place correctly (`UBND phường Nhật Tân`, `Công an phường Bồ Đề`). Zooming to the globe (180° × 360°) clears the layer and the row reads `zoom in to load — too wide to map`. No console errors.

That capture predates #165, which has since merged. Re-checked on this branch rebased onto current `main`: the layer draws again — `25 in view · mapped to 2026-06-01` over central Hanoi at a wider `0.30° × 0.16°` box, `error: null`, nothing truncated. Before #165 landed it reported `UNAVAILABLE · OpenStreetMap · Overpass 406` and drew nothing, which was correct degradation on its side.

## Notes

- `DATA_SOURCES.md` and `dataCredits.js` gain the matching ODbL attribution.
- The registry assertion in `layerState.test.mjs` moves 16 → 17. It exists to make adding a layer deliberate, so it is updated here on purpose. The layer takes share-link token `o`.
- Nothing is bundled; all data is fetched at runtime.

This is a new layer, so it is a product call rather than a fix — happy to close it if Civic Offices is not a direction you want, or to split the meta-line honesty work out of it if the layer itself is the part in question.


daikaginza:
Closing this myself rather than leave a PR sitting in the queue that can't be merged any more.

It was written against the layer contract as documented at the time — CONTRIBUTING said each layer was one self-contained module in `src/data/<layer>.js`, and `src/layers/` didn't exist yet. I checked before deciding rather than assuming: it still rebases onto `5228ab9` with no conflicts at all. What took its integration point away was the 11 September refactor that cut `src/main.js` down to the `createStandaloneApplication` bootstrap, and then the layer extraction on the 13th. So there's nothing to argue about here — the ground moved, that's all.

Rebasing onto `main` today would be the wrong fix rather than just a painful one. The import block this PR edits is gone, and forcing those hunks through would register a layer through a mechanism you've since dismantled: it would look mergeable and be wrong.

The data side doesn't depend on any of that and still holds up — Overpass query construction, the civic-office classification, payload parsing, the too-wide-viewport guard, and the "N of M shown" label, with unit tests. Porting it into `src/layers/civicOffices/` against the current interface looks like roughly the shape of `src/layers/earthquakes`. Happy to do that and open a fresh PR if a civic offices layer is something you want. If it isn't, that's a perfectly good answer too and I'll leave this closed.

PR #168, opened 2026-09-03
test(locations): check the destination preset table as data

daikaginza:
## Why

`CITY_POIS` is the only place in the app where a landmark's position is written by hand, and nothing validates it. Every other test in the suite exercises code. The table's failure mode is different, and that is what makes it worth a file: a wrong entry **looks like working software** — the pill lights up, the camera flies, and it frames the wrong place.

## The invariant that pays for it

Every POI must lie inside its own destination's `viewBounds`. That rectangle is what "fly to \<city\>" opens on, so a landmark outside it is one the destination advertises but its own overview never shows.

**These pass clean on the current eight destinations.** This is a guard for the next entry, not a fix for an existing one — I'd rather say that plainly than imply it found something here.

It was written while adding destinations to a fork, where it caught one immediately: a landmark sitting 15 km outside its destination's western edge, added by hand a fortnight earlier and never noticed, because the camera still flew somewhere that looked reasonable.

## The rest

Structural rather than stylistic, so an entry framed unusually is still free to be added:

- coordinates finite and on Earth
- `viewBounds` corners the right way round (a swapped pair frames nothing, but the framing code still runs)
- pitch pointing downward, heading in `[0, 360)`, positive `alt`
- required keys present
- `LOCATIONS` still derived from `CITY_POIS` rather than kept alongside it

## Verification

- `npm test` — **2693 tests, 0 failures** (7 new)
- `npm run build` — clean
- CI on this branch: Node 24.14.0, Node 26.x and Windows onboarding all green
- The invariants bite on this table: moving `Golden Gate Bridge` 3° west fails with `sf · Golden Gate Bridge at 37.8199, -125.9 is outside sf's viewBounds`; swapping London's `viewBounds` corners fails with `london viewBounds latitudes are inverted`.

No runtime code changes. Happy to drop or trim this if it is more ceremony than the table warrants.


bilawalsidhu:
Landed via #506 (rebased onto the current component layout, with fixes from review). Thank you @daikaginza!

PR #112, opened 2026-08-29
fix: remove node:fs from browser-built data modules (#34)

Tom-Neverwinter:
Closes #34.

`npm run build` warns on every production build:

```
[plugin vite:resolve] Module "node:fs" has been externalized for browser compatibility, imported by "src/data/naturalEarthRegions.js".
[plugin vite:resolve] Module "node:fs" has been externalized for browser compatibility, imported by "src/data/neighborhoodPolygons.js".
```

Both modules carried a Node-only `fs` branch, only reachable under `node --test` (Node's loader requires an explicit import attribute for a JSON module that Vite doesn't need), alongside a Vite dynamic-JSON-import branch for the browser.

## Fix

A dynamic `import('./x.json', { with: { type: 'json' } })` works identically under Node's native loader and under Vite (dev + build, still its own lazy chunk) — Node 20.10+/22+ support the `with` import-attribute syntax natively. One loader for both environments, zero `node:fs` reference anywhere in either file, so there's nothing left for the browser build to externalize.

## Verification

`node --test` on both files' suites (20/20 pass), `npm run build` with the `node:fs externalized` warnings gone and the regions/marine/san-francisco JSON chunks still split out exactly as before (same sizes, still lazy — not folded into the eager bundle).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #111, opened 2026-08-29
fix: add tooltips for visual-style presets and disabled/unavailable data layers (#90)

Tom-Neverwinter:
Closes #90 (the tooltip requests — the Docker/onboarding half of that issue is the maintainer's own separate in-progress work, per their comment on the issue).

## Changes

- `index.html`: native `title` attributes on the 7 VISUAL PRESETS buttons spelling out what CRT/NVG/FLIR/Noir/Anime/Snow mean, not just the abbreviation on the button.
- `src/data/manager.js`: each data-layer toggle row now carries the same status text already shown on its meta line (source, feed state, and any error like "KEY REQUIRED") as a hover title too, so the reason a layer is greyed out doesn't require spotting a small line of text under a dense toggle grid.

Both native `<title>` tooltips — no new UI/JS framework code.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #110, opened 2026-08-29
fix: split server-only Places/Street View key from the browser-exposed Google Maps key (#33)

Tom-Neverwinter:
Separate optional Google server credentials for Places and Street View from the browser-exposed key. Existing single-key and keyless setups remain supported. Closes #33.

Both keys live in one ignored root `.env` for terminal development, or the ignored `pinokio/ENVIRONMENT` for Pinokio. The tracked examples document both entries. Provider Settings offers separate browser/server rows, and setup diagnostics recognize the server key. Only the browser key enters the client bundle.

The review follow-up fixes Pinokio app-value/blank precedence and makes the Street View tool resolve environment overrides per variable before preferring the server key. New tests exercise both Places routes in dev and preview, server-only/split/single-key/keyless configurations, CLI precedence, setup redaction, and Pinokio credential isolation.

Authorship: original key-split implementation by @Tom-Neverwinter; follow-up integration and regression coverage by Sameh Khamis. The original contribution remains in this PR.

Validation: setup doctor, complete unit/allocation suite, 58 focused tests, production build, and a production-output sentinel check proving that the server key is absent while the browser key is present. Tracking regression: 108/108 passed. Provider Settings was inspected in a real browser: both rows render with masked inputs, and only the browser row is marked client-exposed. Node 24, Node 26 and Windows CI passed. Browser checks used SwiftShader on Linux.

PR #171, opened 2026-09-03
fix(a11y): preserve Map Source focus through delayed opening

### LINKED ISSUE #54: qa:map-source-tray fails on clean main: long Enter hold check races the disclosure debounce
`npm run qa:map-source-tray` fails deterministically on a clean checkout of `main`, on the check **"long Enter hold cannot strand the disclosure keyboard path"**.

Raising it because CONTRIBUTING lists the QA gates as things that must stay green before a PR, and this one does not start green.

## Reproduce

```bash
git clone https://github.com/bilawalsidhu/gods-eye-view.git
cd gods-eye-view && npm install
GOOGLE_MAPS_API_KEY="…" CESIUM_ION_TOKEN="…" npm run dev -- --host localhost --port 4173
QA_BASE_URL=http://localhost:4173 npm run qa:map-source-tray
```

```
[PASS] exact four-source presentation; the retired left Map Stack panel is gone
[PASS] compact wing is a semantic disclosure
[PASS] Enter opens the tray and hands focus to a Map Source tile
[PASS] Escape closes the tray and restores disclosure focus
[PASS] Space opens the tray through the same keyboard path
[FAIL] long Enter hold cannot strand the disclosure keyboard path — {"expanded":"false","activeStack":null}
```

The run then aborts on the next step (`Node is either not clickable or not an Element` at `qa-map-source-tray.mjs:194`), because the tray is closed when the following check expects it open.

**6/6 runs failed**, identical output every time. Not flaky.

## Environment

| | |
|---|---|
| Commit | `880a672` (unmodified `main`) |
| Node | **24.14.0** (also reproduced on 24.4.1) |
| Puppeteer | 24.37.5, Chrome 145.0.7632.77, headless |
| OS | Linux 6.8.0 |
| Keys | real Google Maps key with Map Tiles API enabled, real ion token |

Reproduced on a supported Node specifically to rule the version out.

## This may be the harness, not the app

Walking the same key sequence manually against the same build, but with an extra
snapshot between the `Enter` release and the following `Escape`, reaches the
state the check wants:

```
Escape (close)              {"expanded":"false", active:"control-panel-toggle"}
Enter HELD 320ms (down)     {"expanded":"true",  active:"control-panel-toggle"}
Enter released              {"expanded":"true",  active:"photoreal"}
Escape after hold           {"expanded":"false", active:"control-panel-toggle"}
Enter (expect reopen)       {"expanded":"true",  active:"photoreal"}   ← what the check wants
```

So the disclosure does recover from a long hold; the keyboard path is not
actually stranded.

The disclosure open/close is debounced (`scheduleOpen` / `scheduleClose` in
`src/ui.js`, each behind `openDelayMs` / `closeDelayMs` with `:hover` and
`:focus-visible` guards), while the QA script drives it with fixed 300 ms and
320 ms sleeps. That reads like the harness racing the debounce rather than a
defect in the tray — the assertion samples state at a fixed offset that lands on
the wrong side of a pending timer.

If that is right, the fix is for the check to wait on controller truth
(`aria-expanded` settling) the way the later Bing step already does with
`page.waitForFunction`, instead of a fixed sleep.

I have not proposed a patch since I am not sure whether the intended contract is
"recovers eventually" or "recovers within 300 ms" — happy to send one either way
if you confirm which.

## Possibly related

#44 reports another QA gate failing on clean `main`.

Tom-Neverwinter:
Opening Map Source from the keyboard could attempt focus before its opening transition finished. The tray now retries until the selected source can receive focus, with a bounded timeout. Leaving the disclosure, pointer interaction, closing the tray, and UI disposal revoke pending work so a later callback cannot steal focus.

Closes #54.

The final scope is Map Source focus, its regression coverage, and its browser QA. The QA waits for observable focus and Esri fallback state, and uses the appropriate browser backend on Linux/Windows as well as macOS. Existing keyboard activation behavior is preserved; broader keyboard/voice changes are separate.

Tom-Neverwinter contributed the original delayed-focus repair and QA improvements. Manjunath (@manjunath22466) contributed the selected-source and cancellation refinements, retained in a separate authored commit. Maintainer integration and additional verification are separate commits.

The other QA portability changes, proxy error sanitization, CCTV launcher/configuration cleanup, and Natural Earth helper cleanup have been removed from this landing scope. They remain in the original commits for separately reviewed, credited follow-ups. The ground-floor fix for #44 was already handled upstream.

Validation on Node 24.14.0: doctor ready; 2,749 unit/allocation tests passed, one skipped; production build passed; 53 focused tests passed. Additional regression cases cover delayed visibility, selected-source preference, bounded attempts, focus departure, pointer cancellation, closure/reopening, stale callbacks and disposal. All 23 keyless Map Source browser checks and all 108 tracking regression checks passed. The browser run covers desktop/narrow layout and real keyboard focus; no new page or console errors were recorded.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -138,5 +138,3 @@ AISSTREAM_API_KEY=
 # CCTV_MAX_SOURCES=48                                  # overall cap on registered cameras
 # CCTV_PREFER_AUSTIN=1                                 # prioritize Austin cameras in the registry
 # CCTV_FORCE_AUSTIN=                                   # force Austin-only sources
-# CCTV_AUTO_CALIBRATE=1                                # auto-calibrate camera projection (client flag)
-# CCTV_DRAPE_MESH=1                                    # drape frames onto 3D geometry (client flag)
--- a/.env.example
+++ b/.env.example
@@ -138,3 +138,5 @@ AISSTREAM_API_KEY=
 # CCTV_MAX_SOURCES=48                                  # overall cap on registered cameras
 # CCTV_PREFER_AUSTIN=1                                 # prioritize Austin cameras in the registry
 # CCTV_FORCE_AUSTIN=                                   # force Austin-only sources
+# CCTV_AUTO_CALIBRATE=1                                # auto-calibrate camera projection (client flag)
+# CCTV_DRAPE_MESH=1                                    # drape frames onto 3D geometry (client flag)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,16 @@ of current runtime behavior, see [`docs/CURRENT-STATE.md`](docs/CURRENT-STATE.md
 
 ### Fixed
 
+- `scripts/dev-cctv.sh` now binds to `localhost` by default, matching
+  `dev-fresh.sh` and `dev-secure.sh`. Set `HOST=0.0.0.0` to opt in to LAN
+  exposure. (#265)
+- The Celestrak, terrain-heights, adsbdb, and launch-library proxies no longer
+  place raw exception text or upstream error bodies in HTTP responses; the real
+  error is logged server-side and the client gets a generic message. (#262, #263)
+- `naturalEarthRegions.js`’s internal `haversineKm` takes `(lat, lon, lat, lon)`
+  like every sibling implementation instead of longitude-first. (#274)
+- Dropped `CCTV_AUTO_CALIBRATE` and `CCTV_DRAPE_MESH` from `.env.example`; no
+  code reads them. (#270)
 - Mapped-site outages show their scheduled retry countdown and distinguish
   known Overpass rate limits, timeouts, and query failures. Search feedback no
   longer claims a refresh succeeded while the layer is unavailable or loading.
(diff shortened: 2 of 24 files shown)


bilawalsidhu:
Thanks for the volume of work here. Quick status so you know where to spend time.

Superseded: #108 and #112 by commits already on main, #109 by the keyless boot recovery that shipped in v0.1.0, and #111 and #113 by #147 and the weather layer now in flight. I'm closing those as superseded.

Landed: #14 is the base of the ALPR layer that's about to ship, with your commits and authorship intact.

Wanted next: rebase #171, #114, and #110 on main and I'll review them this week.

On hold: #13 waits on confirmation of the NIFC data terms.

The receiver taps (#116, #115, #15, #12, #11) are yours if you want them. Take one at a time, each with tests, a DATA_SOURCES entry, and a CURRENT-STATE section, and land the shared tap contract in one small PR first so the rest stack on it. Because CI can't reach the hardware, include a recorded fixture per tap so the tests run without a device.

samehkhamis:
Merged—thank you @Tom-Neverwinter for the Map Source keyboard fix and QA improvements! Thanks also to @manjunath22466 for the focus refinements.

PR #181, opened 2026-09-04
fix(firms): record source success after aggregation

mwilliamsGH:
FIRMS source success is now recorded only after its rows have been appended. If aggregation throws, the source reports failure without a contradictory success entry.

The large-feed argument-overflow fix already landed in #93. This PR retains the existing append loop and carries only the remaining source-status ordering change, regression coverage, and runtime documentation.

The new tests cover 200,000-row retention, expired-row filtering, source order, partial and total upstream failure, valid empty sources, and an injected aggregation failure. The aggregation regression fails against current main and passes with this change.

Validation on 893cdf5: 17 focused FIRMS tests; 2,728 full unit/allocation tests passed with one Windows-only skip; doctor and production build passed; all 108 browser tracking checks passed. Node 24.14.0, Node 26.x and Windows CI are green. The aggregation regression was confirmed to fail against unchanged main. FIRMS tests use synthetic data; live NASA validation was not performed.


samehkhamis (approved):
The large FIRMS feed overflow fix already landed in #93. This PR now keeps the remaining source-status ordering correction, with regression coverage. Thank you for the contribution!

PR #185, opened 2026-09-05
fix: validate file and env CCTV source URLs

gothardener:
Hey, this cleans up camera urls from file and env packs, since those are hand written and go straight into the registry right now.

The reason is pretty simple, a small typo or a plain http url can sit there unnoticed, and then the camera just stays stuck on fallback while taking up a slot. It can also mean the server tries to fetch something it never should. So now the urls get checked when the pack loads, bad ones are dropped with a warning, and the normal fallback takes over like before. I also added tests for the good and bad cases, plus a short changelog note.


bilawalsidhu:
The validation idea is sound, but the diff patches normalizeSourceItem in vite.config.js, and that function now lives in server/providers/cctv/normalize.js after the provider extraction. Closing; a resubmission against normalize.js is welcome.

gothardener:
Understood.

PR #189, opened 2026-09-06
cctv: live HLS/RTMP video feeds (proxy-owned segment store, hls.js, native video texture)

Danielslay86:
## Summary

CCTV can now play live video instead of only refreshed stills. Tested against DelDOT (Delaware), whose Wowza server turned out to be a good stress test. Most of what's below exists because of how that server behaves.

## What this adds

- **Proxy**: `/api/cctv/media/<id>/...` sub-path resolution with origin pinning, manifest rewriting, and a proxy-owned segment store with two strategies:
  - `createHlsPuller` for `.m3u8` upstreams. Polls the chunklist, re-resolves the master when a session is refused, and keys segments by the stream-global number in the filename so session rotations stitch with no overlap or loss. No ffmpeg.
  - `createHlsRemuxer` for RTMP and other stream kinds. Runs ffmpeg (`-c copy`) as a disposable segment producer with instant respawn, a stall watchdog, and a proxy-generated playlist with continuous numbering and discontinuity tags at respawn boundaries. **ffmpeg is optional.** Without it, `.m3u8` still plays through the puller and the stills path is untouched.
- **Client**: hls.js (dynamic import, so stills-only catalogs don't load it), with a Brave-safe branch and network/media recovery. Cesium's native video texture, sized from the real stream and rebound on resolution change. A playback-rate governor for cameras whose encoder clock runs slow relative to wall time. DelDOT delivers about 52 s of content per real minute; cameras with honest clocks sit at 1.0 and never leave it.
- **Panel**: the preview card plays live video via `captureStream()` of the active camera's element (one decoder, two surfaces) and widens for it. Stills keep the existing `` path.
- **Delaware pack**: two calibrated DelDOT cameras in `config/cctv_sources.delaware.json` as a working example of a video-capable source. They use `rtmpt://video.deldot.gov:80/...` (RTMP tunneled over HTTP, which proved steadier). `rtmp://video.deldot.gov:1935/...` is the same stream over native RTMP.
- **Docs**: README (live-video section, CCTV row), SECURITY (origin pinning, ffmpeg spawn constraints), TESTING (soak scenario), DATA_SOURCES (DelDOT).
- Minor: OSM imagery capped at z19 so the provider stops requesting tiles that don't exist.

## Why it looks like this

DelDOT's Wowza restarts every publish on a 60-second timer. Over HLS that loses about 15 s per minute (a discarded partial segment plus the first-chunk delay). Over RTMP it loses about 1 s, so the remuxer prefers an RTMP URL when the catalog has one. Separately, the cameras deliver less content time than wall time, which no transport can fix, so the client governs playback rate. The measurements are in the commit history and I can write them up in a doc if that's useful.

## Testing

Activate a Delaware camera and let it run. On DelDOT it holds indefinitely with roughly a 30 s live delay. TESTING.md section 6 has the soak script. Stills-only setups are unchanged.

## Known limitations

- DelDOT rate-limits by client IP. Rapid activate/reload cycles from one host (tuning, testing) trip it, and it shows up as 404s and stalls that clear after a quiet minute. Not a pipeline fault. Leave a minute between test runs.
- If hls.js rebuilds its pipeline mid-stream, the preview card can hold its last frame while the projection plane keeps playing. Re-activating the camera clears it.

## Not in this PR

A DelDOT loader (per-agency catalog fetch, Caltrans-style) and a WebGL render path for the card. Both follow separately.


OakSwingZZZ:
Nice Work! I am hoping to add Caltrans support too. Needs HLS too.

Danielslay86:
The CCTV subsystem was refactored into server/providers/cctv/* and src/ui/cctv* after I opened this, so it now targets the old file layout and won't merge cleanly. The HLS/RTMP pipeline itself still isn't in main, so I'm re-homing it onto the new module structure and will open a fresh PR against current main. Closing this in favor of that.

PR #197, opened 2026-09-08
fix(launches): preserve unknown payload details

onionviolet:
Missing launch payload records currently display “CLASSIFIED / MULTI-PAYLOAD,” and a null mass is coerced to 0 KG. Use “PAYLOAD DATA UNAVAILABLE” and “Unnamed payload,” and preserve missing or invalid mass as null while retaining explicit zero and numeric values.

Adds normalization regressions for null, absent, blank, boolean, negative, invalid, zero and numeric-string mass values. The regression fails against unchanged main.

Validation: full unit suites on Node 26.4.0 and 24.14.0, including the Node 24 allocation checks, and production build using main's package-lock dependencies. The Windows-only test is skipped on macOS; Node 26's allocation checks are intentionally skipped and exercised on Node 24.

Follow-up browser validation on this exact commit: `npm run doctor -- --json` passed; `npm run test:track` against a running localhost dev server passed 108/108 with no failures or skips. A focused visible-Chrome smoke test verified the actual Space Missions panel with synthetic records: absent payloads show PAYLOAD DATA UNAVAILABLE, and an unnamed payload with null mass does not display 0 KG. Screenshots were inspected.

The initial tracking attempt ran without a working dev server and timed out; the successful run above supersedes that setup failure. GitHub CI has not reported checks yet; no CI result is claimed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ Earthquake refreshes validate the complete feed and construct replacement entiti
 
 Non-object or array-valued properties reject the response instead of being treated as an unknown magnitude.
 
+Launch payloads with missing records now say PAYLOAD DATA UNAVAILABLE. Missing names use Unnamed payload; absent or invalid mass stays unknown instead of appearing as 0 KG.
+
 This changelog records public product changes. For the authoritative description
 of current runtime behavior, see [`docs/CURRENT-STATE.md`](docs/CURRENT-STATE.md).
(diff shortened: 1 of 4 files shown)


samehkhamis (approved):
Looks good! Thank you for the contribution and for resolving the documentation conflicts. The updated payload handling matches the intended behavior, and the rebased changes look good.

PR #198, opened 2026-09-08
fix(earthquakes): validate snapshots before replacing entities

onionviolet:
A malformed USGS row or duplicate rendered event ID can fail after the previous earthquake entities have already been cleared. Validate the complete feed and construct replacement entities before clearing the current collection, preserving last-good entities, overlays, count and timestamp when validation fails.

Reject malformed geometry, invalid coordinates or magnitude, and duplicate rendered IDs. Missing magnitude does not establish M2.5+ eligibility and is excluded. Successful empty feeds still clear the snapshot. This change is limited to snapshot validation and replacement; fetch cancellation and response ordering are unchanged.

Adds a runtime regression covering malformed rows after a valid row, last-good entity identity/overlays/stats, and recovery. It fails against unchanged main.

Validation: full unit suites on Node 26.4.0 and 24.14.0, including the Node 24 allocation checks, and production build using main's package-lock dependencies. The Windows-only test is skipped on macOS; Node 26's allocation checks are intentionally skipped and exercised on Node 24.

Follow-up browser validation on this exact commit: doctor passed, and `npm run test:track -- --url http://127.0.0.1:4187` passed 108/108 with no failures or skips. A focused visible-Chrome smoke test loaded a synthetic earthquake, injected a malformed subsequent response, and verified that the same entity stayed visible with its count and last-success timestamp unchanged. The error was reported as Malformed USGS response; before/after screenshots were inspected.

Earlier browser attempts were invalidated by a missing server and then a localhost IPv4/IPv6 port collision with an unrelated HTTP listener. The dedicated-address run above supersedes them. GitHub CI has not reported checks yet; no CI result is claimed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## September 8, 2026
+
+Earthquake refreshes validate the complete feed and construct replacement entities before clearing the previous snapshot. Malformed rows and duplicate rendered IDs retain the last good entities, overlays, count and timestamp and report a malformed response; unknown magnitude is excluded from M2.5+ rendering.
+
 This changelog records public product changes. For the authoritative description
 of current runtime behavior, see [`docs/CURRENT-STATE.md`](docs/CURRENT-STATE.md).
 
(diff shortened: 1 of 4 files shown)


samehkhamis (approved):
Looks good! Thank you for the contribution. I added a small guard and regression coverage for malformed properties values. The unit tests, build, browser checks, all 108 tracking checks, and CI are green.

PR #207, opened 2026-09-09
feat(cctv): add Traffic Scotland, Perth & Kinross, and Scottish mountain webcam sources

garrywilliams:
## Summary

- Add **Traffic Scotland** trunk-road cameras as a keyless default CCTV pack alongside Austin, Caltrans and TfL. The public site's `/tsis/cameras` catalog (415 sites, 500 views, coordinates included) becomes one camera per view (`ts-<view id>`), capped at 300 nearest Edinburgh and Glasgow (`CCTV_TRAFFIC_SCOTLAND_MAX_SOURCES`; 500 covers the whole network), kill switch `CCTV_TRAFFIC_SCOTLAND_ENABLED=0`.
- Traffic Scotland publishes no image URLs: each site's stills arrive as base64 data URIs inside an HTML fragment. The proxy fetches the fragment, validates the JPEG/PNG bytes, caches every view at that site for four minutes (upstream republishes every ~5 min in the central belt, ~20 min elsewhere), shares one in-flight fetch across concurrent views, and serves a stale frame when a refresh fails. Both `/frame/:id` and `/media/:id` route through it so the fragment is never piped to the browser as an image.
- Add a **bundled, hand-curated Scotland pack** (`config/cctv_sources.scotland.curated.json`): Perth & Kinross Council's 23 road cameras (15-minute stills; headings taken from the caption burned into each frame, so 20 carry a CURATED badge) and 22 mountain/ski webcams (Cairngorm Mountain, Glencoe Mountain Resort via Winterhighland, Glenshee, Aboyne airfield, Ben Nevis from Loch Linnhe, Leadhills). Bundled packs load as additive peers of the live packs — unlike `CCTV_SOURCES_FILE` they never suppress them — and `CCTV_BUNDLED_PACKS_ENABLED=0` skips them.
- Sources may now carry `frameRefreshMs` (server → `/api/cctv/sources` → client `staticFrameRefreshMs`), so a pack states its operator's cadence instead of relying on the provider-name table. Traffic Scotland is also added to that table at five minutes.
- `CCTV_MAX_SOURCES` default rises 900 → 1,150 so no default pack is silently truncated.

Existing packs, the SSRF posture (server-registered URLs only; the Traffic Scotland resolver additionally pins the official fragment origin and a `ts-<digits>` id shape), the fallback chain, and all client pose/calibration behaviour are unchanged. All new cameras start as RAW PRIOR or CURATED priors and calibrate with the existing gizmo.

## Data sources and terms

- **Traffic Scotland** — the catalog and fragment endpoints are the ones the public traffic.gov.scot site uses, not a documented API; the documented developer route is a registered FTP service. The site describes its information as freely available and the developer hub asks that Traffic Scotland be clearly credited, which the Data attribution popover now does. Frames already carry Traffic Scotland's own watermark and timestamp. Images are withheld upstream while a camera is in operational use, so individual views go dark without warning.
- **Perth & Kinross Council** and the **mountain webcam operators** publish plain still URLs with no stated reuse licence. Frames are proxied live, never stored or redistributed, polled at each operator's stated interval, and every operator is credited by name and link. `DATA_SOURCES.md` records this and says to remove an entry if an operator objects.
- Positions: trunk-road council sites reuse Traffic Scotland's published coordinates; the rest are placed from OSM features matching the council's own descriptions, so treat them as ±100 m priors. The Ben Nevis frame is ~1.6 MB per fetch and is the heaviest camera in the catalog.

## Validation

- Node 26.8.1: `npm test` — 2,715 passed, zero failed, one skip (allocation microbenchmarks are calibrated for Node 24 and skip on 26).
- `npm run build` — passed; existing chunk-size warning only.
- `npm run test:track -- --url http://localhost:4199` — 108 passed, zero failed, zero skipped. (With my local, restricted Google Maps key left in `.env`, the three "no console errors" checks fail because Google 3D tiles return 403 for that key; blanking the key makes the same run fully green, so the failures are environmental rather than caused by this change.)
- New unit coverage: `src/data/trafficScotlandCctv.test.mjs` (loader, origin pin, fragment decoder, shared in-flight fetch, cache expiry, stale fallback, timeout) and `src/data/curatedCctvPacks.test.mjs` (shipped pack schema — unique ids, https only, Scottish bounding box, bounded cadence, curated entries must carry a heading — plus loader filtering and kill switch).
- Live dev server: `/api/cctv/sources` registered 1,145 cameras (300 Traffic Scotland, 23 Perth & Kinross, 22 mountain); 53 Traffic Scotland cameras fall within 15 km of Edinburgh. Frames for A720 Dreghorn, both A702 Boghall views, every Perth & Kinross sample, and each mountain operator returned `image/jpeg` with `X-CCTV-Source: upstream-image`; a repeat Traffic Scotland fetch was served from the site cache in under a millisecond; `/media/ts-43` and `/stream/ts-43` behave as for other still packs; `/health` reports the new cameras as `ok/snapshot`.

https://claude.ai/code/session_01CDZGxw7NcDiUpUXMzNx1kB


bilawalsidhu:
Traffic Scotland's copyright page states the Scottish Government retains full copyright of the camera images and that users may not hyperlink directly to them, so we can't take the Traffic Scotland part. Do you have a developer agreement with them? If not, the Perth & Kinross and mountain webcams would be welcome as a separate, smaller PR against the current server/providers/cctv layout.

garrywilliams:
You're right, I missed the Live Traffic Cameras section of their copyright page. No agreement in place, so I'll drop Traffic Scotland entirely and resubmit just the Perth & Kinross and mountain webcam packs as a smaller PR on the new `server/providers/cctv` layout, following the Warendorf catalog pattern. Closing this one.

Issue #48, opened 2026-08-25
Feature Request: Simplify deployment for non-developers (Docker & One-Click Cloud Deploy)


amirshahab89:
Loving what you've built here! Right now, testing out *God's Eye View* requires a full local dev setup—cloning the repo, setting up Node/npm, and managing `.env` files.

For non-developer users like OSINT researchers, spatial analysts, or people just wanting to evaluate the app, that setup creates a pretty high barrier to entry.

Since the app relies on a Node/Vite server layer to proxy API calls securely (as noted in `SECURITY.md`), static hosting like GitHub Pages isn't ideal. However, there are two low-maintenance ways we could make onboarding way easier without breaking security or adding support overhead:

**1. Pre-built Docker Image (`ghcr.io`)**

* Add a basic `Dockerfile` and a GitHub Action (`docker-publish.yml`) to automatically push container images to GitHub Container Registry.
* This eliminates local Node version mismatches and OS quirks entirely. Users could just run:
```bash
docker run -d -p 5173:5173 --env-file .env ghcr.io/bilawalsidhu/gods-eye-view:latest

```

**2. One-Click Cloud Deploy (Vercel or Render)**

* Add a basic `render.yaml` or server-compatible `vercel.json` and a "Deploy" button to the `README.md`.
* Users could spin up their own cloud instance in minutes and paste their API keys into a web interface instead of using the terminal.

**Suggested Roadmap:**

* [ ] Add `Dockerfile` and test local builds.
* [ ] Add GitHub Action for automated container pushes.
* [ ] Add Cloud Deploy config (`render.yaml` / `vercel.json`).
* [ ] Update `README.md` with container and deploy instructions.

I don't code myself, but I'd be more than happy to help out by thoroughly testing pre-built Docker containers or cloud deployments across different browser environments if someone picks this up!

bilawalsidhu:
Consolidating #43 and #504 here, plus the packaging portion of #384. Preserve Docker/GHCR publishing, compose/NAS use, persistence and container-to-host key-setup behavior, cloud deployment, and a Windows-native one-click installer as distinct proposed delivery paths. Acceptance needs a working first launch and explicit credential/LAN trust policy, not just a Dockerfile or opening dist/index.html.

The overlapping Docker PRs #55, #208, #252, #353 and #409 remain open until one maintained replacement preserves their useful workflow, compose and documentation contributions. No Docker implementation is selected or approved by this consolidation.

[...]

Issue #43, opened 2026-08-25
Feature Request: Add Docker & Docker Compose installation/deployment option


kareemaiman:
### Summary
Currently, God's Eye View requires a local Node.js environment (specifically Node.js 24.14.x or 26.x) and manual dependency installation. Adding a containerized setup with a `Dockerfile` and `docker-compose.yml` would make self-hosting, homelab setups, and cross-platform execution effortless without needing to manage local Node version managers.

### Proposed Solution
Provide:
1. A multi-stage `Dockerfile` (or lightweight Alpine Node runtime).
2. A `docker-compose.yml` that passes environment variables (e.g. `GOOGLE_MAPS_API_KEY`, `OPENAI_API_KEY`) via `.env` file.
3. Updated instructions in `README.md` under Quick Start for Docker deployment.

#### Example `Dockerfile`:
```dockerfile
FROM node:24-alpine AS builder
WORKDIR /app
COPY package*.json ./
RUN npm ci
COPY . .
RUN npm run build

FROM node:24-alpine
WORKDIR /app
COPY --from=builder /app ./
EXPOSE 4173
CMD ["npm", "run", "dev", "--", "--host", "0.0.0.0", "--port", "4173"]
```

#### Example `docker-compose.yml`:
```yaml
version: '3.8'

services:
  gods-eye-view:
    build: .
    container_name: gods-eye-view
    restart: unless-stopped
    ports:
      - "4173:4173"
    env_file:
      - .env
```

### Benefits
- Zero Node.js version conflicts on host systems.
- 1-command deployment with `docker compose up -d`.
- Simplifies reverse proxy setups (Traefik, Nginx, Caddy).

DarshSinghKushwaha:
Fixed in PR #55 for issue #48 . Please Have a look at it, constructive criticism appreciated 👍.

bilawalsidhu:
Thanks for the report. Consolidating this into #48; its remaining requirements are preserved there. Both request Docker/deployment packaging; #48 is broader and already has linked implementation work. This closure does not mean the remaining work is fixed.

PR #208, opened 2026-09-09
feat: add a Docker install path

SM-26:
Adds a Docker install path: `docker compose up` builds a pinned Node image and serves the app on `http://localhost:4173` with no local Node install. Refs #43, #48.

### Why another one

I found #94, #55 and #42 after writing this, and I'm happy to fold this into any of them rather than add noise to the queue — say the word. Posting it because three things here are load-bearing and, as far as I can tell, aren't covered by the others:

**1. The published port binds loopback, not every interface.** Docker's usual `"4173:4173"` publishes on `0.0.0.0`. On a dev server that brokers whatever provider keys the user has configured, that quietly makes the README's "Sharing an instance" warning the *default* rather than an opt-in. `"127.0.0.1:4173:4173"` keeps the container's posture identical to `npm run dev`.

**2. The repo is bind-mounted, so state survives the container.** Without a mount, keys pasted into Provider Settings are written to `/app/.env` inside the container and die with it — so does `.gev-cache` and `.gev-logs`. That's a confusing first-run experience for exactly the non-developer audience #48 is about. `node_modules` stays in the image (anonymous volume) so a host install can't shadow the Linux-built `sharp`.

**3. No `env_file:`.** Vite's `loadEnv` already reads `/app/.env` off the mount, and a compose `env_file: .env` hard-fails when the file is absent — which is the documented keyless-start path.

Two smaller notes: the container runs as the host user, since everything it writes lands in your working tree (Vite's dep optimizer still needs to write inside the image's `node_modules`, hence the two loosened dirs); and `PUPPETEER_SKIP_DOWNLOAD=1` keeps Chromium out of the image, as it's only needed by the QA harness.

### Verified

- `docker compose up` on a clean clone with no `.env` → keyless boot, first-run panel, globe renders.
- `npm run doctor` in the container: Node v24.20.0, providers detected correctly.
- `npm run build` ✅ and `npm test` ✅ in the container.

Because Actions sits at `action_required` for fork PRs, I also replicated `ci.yml`'s `verify` job locally — same steps, same env, both matrix legs, on a clean `git archive` of this branch:

| Leg | doctor | `npm test` | build |
|---|---|---|---|
| Node 24.14.0, `GEV_REQUIRE_ALLOCATION_GATE=1` | ✅ | 2706 passed / 0 failed / 1 skipped, plus the gated allocation batch 13/13 (31 min) | ✅ 12.87s |
| Node 26.8.1, `GEV_REQUIRE_ALLOCATION_GATE=0` | ✅ | 2706 passed / 0 failed / 1 skipped | ✅ 11.70s |

The skip is the runner's own note: `SKIPPED 2 allocation microbenchmarks: budgets are calibrated for Node 24, running 26.8.1`. The `windows-onboarding` job's focused test list also passes on a Windows host (43/43), though on Node 26.7.0 rather than the 24.14.0 that job pins.
- A `GOOGLE_MAPS_API_KEY` written to the host `.env` is picked up inside the container (`[OK] Google Maps (dotenv files)`).
- `docker compose ps` → `127.0.0.1:4173->4173/tcp`.

`npm run test:track` I could **not** run: the image skips Chromium by design, and on my Windows host puppeteer's Chrome download extracts to a broken cache. This PR touches no runtime code — no `src/`, no `vite.config.js` — so I don't believe the harness is at risk, but I'd rather flag it than imply it passed. README documents `test:track` as a host-side check for Docker users.

Docs: new "Path 3 — Docker" in the README (including the `down -v` gotcha after dependency changes) and a CHANGELOG entry under Unreleased.


milu70hub:
Tested this recipe on a Synology NAS (DSM 7, Docker 24, amd64): `npm ci` ~10 min on a Celeron, image 630 MB, ~1.25 GB RAM at rest, served behind DSM's reverse proxy plus a small Caddy TLS front so the mic works on iOS. Works well. Three notes worth a line in the README:

1. `user: "${UID:-1000}:${GID:-1000}"` — on DSM set it to the share owner explicitly (the DSM admin user is usually 1026:100), otherwise `.env` / `.gev-cache` end up owned by an unrelated uid.
2. When the port is published beyond loopback the Provider Settings panel is disabled by design, so keys go into `.env` by hand — fine, but worth saying next to the LAN note.
3. Copying the tree from macOS with `tar` ships `._*` AppleDouble files that `npm test` then picks up as test files (500+ spurious failures). `COPYFILE_DISABLE=1 tar …` fixes it; a `._*` line in `.dockerignore` would too.

jklaze:
I built and ran this branch as-is and the Provider Settings endpoints refuse the host:

    GET  /api/setup/status -> 403 "Provider Settings answers only the machine running the server"
    POST /api/setup/keys   -> 403

Docker delivers the host's browser from the container network's gateway, never 127.0.0.1, so the loopback gate rejects it. The README and CHANGELOG lines about keys persisting from Provider Settings don't hold yet.

I have a follow-up commit on top of this branch that fixes it with an exact-address opt-in set only by a Docker launcher script, with tests and a SECURITY.md paragraph: https://github.com/jklaze/big-brother-view/commit/75b71d691393e0ca794a553e057375fb3f9d7b07. Happy to open it against this branch or fold it in however you prefer.

bilawalsidhu:
This is the Docker path we're taking (see #48). Please rebase onto current main and fold in the container-to-host key-setup fix. @samehkhamis or @bilawalsidhu will review.

PR #216, opened 2026-09-09
fix(a11y): add accessible names to hud sliders and inputs

kk376:
Bloom, Sharpen and generated style sliders gain accessible names. Scope and location search receive explicit descriptive names; generated sliders use their visible parameter label. Control behavior and layout are unchanged.

The first-run checkbox retains its native wrapping label, "Don't show this again". The hidden scene file input is unchanged. Focused tests cover the intended labels and preserve native checkbox markup instead of treating every input without ARIA as unnamed. Runtime documentation is updated.

Validation on 3d526f2: all 36 focused HUD and first-run tests passed, including the previously failing first-run markup check. The full suite passed 2,731 unit/allocation tests with one Windows-only skip; doctor and production build passed. Chromium accessibility-tree checks confirmed the static and generated names, native checkbox name, and unchanged slider value/render behavior. All 108 tracking checks passed. Node 24.14.0, Node 26.x and Windows CI are green.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Changelog
 
+## September 8, 2026
+
+Earthquake refreshes validate the complete feed and construct replacement entities before clearing the previous snapshot. Malformed rows and duplicate rendered IDs retain the last good entities, overlays, count and timestamp and report a malformed response; unknown magnitude is excluded from M2.5+ rendering.
+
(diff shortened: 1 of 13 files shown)


samehkhamis (approved):
Looks good! Thank you for the accessibility improvements.

kk376:
Thank you so much for taking the time to review and merge this, @samehkhamis. Really grateful for the opportunity to contribute in a small way to such an awesome project! 🙏

PR #218, opened 2026-09-09
feat(transit): live public-transit vehicles from open GTFS-Realtime feeds (#70)

spcpza:
## Why

Issue #70 asked for a transit layer, and two people offered feeds. Beyond aviation and maritime, public transit is the busiest thing moving in any city, and many operators publish it live with no key. This adds it as one layer package plus one provider, in the current component layout, following the registered-URL-only proxy rule in SECURITY.md.

## What

**Transit layer** (`src/layers/transit/`, one instance per catalog) — buses, trams, metros, trains and ferries as point primitives colored by mode, moving on the globe.

- Seven keyless, openly licensed GTFS-Realtime feeds ship in `src/data/transitFeeds.js`: MBTA Boston, CapMetro Austin, Metro Transit Minneapolis–St Paul, HSL Helsinki, OVapi Netherlands, Entur Norway, TransLink South East Queensland. Each row carries its license, license URL and attribution; each operator is credited in the Data attribution popover the moment its vehicles render.
- Only feeds whose coverage circle contains the camera's look-at point are polled (below a 3,000 km altitude gate, 40 km hysteresis once active). A session over Boston costs one MBTA request per 15 s and nothing elsewhere.
- Motion: every poll gives a vehicle a new fix, and the point glides from where it was drawn to that fix over the next 15 s. One interval behind real time, no extrapolation snap-back (the same trade the flights layer makes). Vehicles absent from two polls are removed; fixes older than ten minutes are ignored.
- Height: near the ground a vehicle reads the shared ground floor (`cachedGroundFloor`) and stays hidden until its cell answers — cells are warmed at most 300 per poll and re-read on a short timer, never per frame. From 60 km up the fleet renders at the ellipsoid, always visible through the mesh.
- Click a vehicle for a shared-host card: route, speed, heading, stop status, occupancy, report age. It follows the vehicle; Escape or a click on empty globe clears it. The handler is ambient: it yields while any tool holds the pointer and never claims it.
- Holds continuous render only while it has vehicles. Share links carry the layer as token `h` (the 19th registry entry).

[...]

PR #219, opened 2026-09-09
feat(directions): keyless A→B directions with turn-by-turn steps and FLY

spcpza:
## Why

Routing already exists in God's Eye View — voice can say "route from A to B" and `/api/route` asks the FOSSGIS OSRM servers for a street-following path — but it is unreachable without an OpenAI key (voice) and a Google key (geocoding the waypoints), and it returns geometry with no turn-by-turn steps. For a globe that boots keyless, directions should work keyless too. This is the click-to-set front end to the routing that is already there, plus steps.

## What

**Directions layer** (`src/data/directions.js`) — the row's chips are the whole interface:

- **DRIVE / WALK / BIKE** pick the OSRM profile (changing it reroutes).
- **SET A**, then click the globe; **SET B**, click again. Escape or a second press cancels. Markers clamp to the ground.
- With both placed, the route is requested and draped on terrain and 3D tiles (`ClassificationType.BOTH`) with the same flowing-dash material as voice routes (now exported from the annotation renderer), with one white dot per maneuver. Click a dot for a card: the instruction, the leg after it, and the next instruction.
- **SWAP** reverses A and B. **FLY** rides the camera along the route through the shared route-flight cinematic (`flyRoute`, the same code voice `fly_route` uses). **CLEAR** removes everything.
- The row reads `OSM routing · 3.4 km · 8 min · Drive`; while routing, `Routing…`; a router miss reads `No route found between A and B`. The layer never draws a straight line as a route.
- Continuous render is held only while a route is drawn (the dashes animate). Share links carry the layer as token `n`.

**`/api/route`** now accepts `steps=1` and returns turn-by-turn steps. The upstream is always asked for steps and the full response is cached, so the annotation engine (no steps) and this layer (steps) share one upstream OSRM call per route. `src/data/routeSteps.js` phrases OSRM's maneuver type / modifier / exit / road name into one plain-English sentence per decision and folds `exit roundabout` into the roundabout step it leaves.

**Two small shared changes.** `initCameraVerbs` accepts a missing view-target getter (so FLY works before any voice session) and never clears an installed one; the manager gains a public `refreshLayerStats()` so a layer whose state changes outside a tick (click placement, routing) can repaint its row.

**Attribution.** OSRM on the FOSSGIS servers is now credited in the Data attribution popover and documented in `DATA_SOURCES.md` with the fair-use terms; voice routes had been using it uncredited since launch.

## Verification

- `npm test`: 2,719 tests, 2,718 pass, 1 skipped (the Node 24 allocation gate; run on Node 22). 12 new: the instruction builder against OSRM maneuver types, roundabout folding, the step cap, chip and stats state tables, params validation, payload validation (a failed route never becomes a straight line).
- `npm run build`: green.
- Headless Chrome (1440×900, Metal ANGLE) against the dev server, keyless, over Helsinki: SET A + click, SET B + click → a 12-step drive; WALK → a 28-step walk; a maneuver dot clicked → card "Turn left onto Vallilantaival · Step 24 of 28 · then 64 m · 51 s · Then: Keep slightly right at the fork" (overlay diagnostics: one `directions-step` entry painted); FLY moved the camera; CLEAR reset the row. Two `/api/route` calls, both 200, no console errors.

## Relationship to the Transit PR

Independent of the Transit layer PR — both branch from `main`. They touch the same registration lines (`main.js`, `layerState.js` and its count, `spriteOrder.js`, `dataCredits.js`, the docs tables), so whichever merges second needs a trivial rebase; happy to do it.

## Follow-ups (not in this PR)

- Let voice `fly_route` see a Directions route, and a `get_directions` voice tool.
- Named endpoints once keyless search (#166) lands: type "Helsinki Central Station" instead of clicking.
- Alternative routes, avoid-highways, and an ETA that respects the Transit layer's live vehicles.


spcpza:
Sibling PR: #218 (Transit layer). Both branch from `main`; whichever merges second needs a one-line rebase on the registration lines, which I'll do.

spcpza:
Rebased onto `main` @ f883dfb (after #319–#323). The `steps=1` change now lands in `server/providers/places/routes.js`, and `src/data/routeSteps.js` is declared in the `place-providers` boundary group and the formatter scope; registration moved to `src/standalone/data.js`. Layer, tests and the two small shared changes (`initCameraVerbs`, exported flow material) are unchanged.

On the new gates: `format:check` ✓, `check:boundaries` ✓, `vite build` ✓, `npm test` 2,947/2,949 — the two failures (the Node 24 allocation gate and `devCctv`'s localhost launcher test) fail identically on pristine `main` under Node 22. Re-verified in headless Chrome against the dev server over Helsinki, keyless: A/B placed by click, a 6-step drive, a 21-step walk, FLY and CLEAR, no console errors.

bilawalsidhu:
Landed via #564 (rebased onto the current component layout, with fixes from review). Thank you @spcpza!

PR #222, opened 2026-09-09
feat: add Sydney public traffic cameras

novasofttechnologiesau:
## Summary

Adds Sydney public traffic cameras using Transport for NSW’s official Live Traffic GeoJSON feed.

- Loads and prioritizes Sydney camera locations in the CCTV catalogue.
- Proxies genuine TfNSW JPEG frames with browser-compatible image headers.
- Adds a 15-second snapshot refresh cadence.
- Adds TfNSW attribution.

## Verification

- Confirmed 217 Sydney cameras load.
- Verified live JPEG frames for five cameras, including Anzac Bridge and Airport Drive.
- `npm run build` passes.
- CCTV proxy tests pass.


bilawalsidhu:
The TfNSW source is fine (official feed, CC BY 3.0 AU), but the implementation swaps the honest User-Agent for a spoofed Chrome one plus a forged Referer inside the two shared fetch paths, so every Austin, Caltrans, and TfL request goes out disguised, and the comment says it's to get past TfNSW's bot check. It also targets vite.config.js, which no longer holds this code. Closing; a resubmission against server/providers/cctv with headers scoped to the TfNSW host, if TfNSW confirms automated access is allowed, would be welcome.

bilawalsidhu:
Follow-up: the Sydney pack is going into a consolidated CCTV integration branch as an opt-in source with a Transport for NSW key, with you credited as author of the pack. That branch will link back here when it lands.

bilawalsidhu:
Landed on main via #433 (squash commit https://github.com/bilawalsidhu/gods-eye-view/commit/12790b3): the Sydney cameras ship as the keyless Live Traffic NSW pack, with you credited as co-author. Thank you.

PR #225, opened 2026-09-09
fix(traffic): keep Street Traffic working when Overpass is unavailable

CarlosEduardoPortugal:
## Summary

Street Traffic can remain stuck at `SYNCING ROAD NETWORK` when the public Overpass mirrors are unavailable, rate-limited, or return an upstream error. This makes the layer render no roads or vehicles even when the separate TomTom flow service is healthy.

This PR adds a bounded OpenStreetMap Standard Map API path for the exact road query emitted by Street Traffic. The response is parsed server-side into the existing Overpass-shaped payload, so the client layer and TomTom matching code keep the same contract.

## What changed

- Detect only the traffic layer's allow-listed `highway` query and validate its viewport bounding box.
- Request `https://api.openstreetmap.org/api/0.6/map` with a 12-second timeout for those bounded road queries.
- Parse OSM XML nodes, ways, tags, and node references into the `{ elements: [{ type: 'way', tags, geometry }] }` shape already consumed by `src/data/traffic.js`.
- Keep the existing four-mirror Overpass rotation as a fallback when the Standard Map API fails.
- Leave arbitrary/non-traffic Overpass queries on the existing generic path, so installation context, annotation geometry, and place lookup behavior are unchanged.
- Preserve the existing cache, response caps, rate/concurrency controls, and same-origin `/api/overpass` route.
- Add parser and request-shape tests, plus update `CHANGELOG.md`, `DATA_SOURCES.md`, and `docs/CURRENT-STATE.md`.

## Why

In a local reproduction, every configured public Overpass mirror failed or refused the road request (502/406 responses and connection timeouts), while the TomTom flow endpoint remained healthy. Because road geometry is a prerequisite for rendering the traffic dots, the UI stayed in its loading state.

The standard OSM map endpoint returned the bounded road data successfully. With this change, the browser completed the traffic load and rendered the existing TomTom-backed traffic layer without any frontend-specific workaround or API-key change.

TomTom remains responsible only for live congestion flow and coverage. OpenStreetMap remains the source for road geometry, with the existing in-app attribution and ODbL documentation.

## Compatibility and trade-offs

- The Standard Map API is used only for a strict, bounded traffic query; arbitrary Overpass QL is never translated or sent to this endpoint.
- The fallback is server-side and preserves the current client response shape.
- Road geometry is still fetched at runtime and is not bundled or redistributed.
- If both OSM paths are unavailable, the existing stale-cache/error behavior remains in place.

## Validation

- `node --test src/overpassProxy.test.mjs` — 11 passed.
- `npm test` — 2,708 passed, 1 skipped, 0 failed. The skipped test is an existing Node 24 allocation benchmark when run under Node 25.
- `npm run build` — passed; only the repository's existing large-chunk warnings remain.
- `npm run test:track -- --url http://localhost:4174` — 103 passed, 5 failed in existing `display-floor/corridor` invariants unrelated to this change; the full run reported clean console and HTTP 5xx checks. The tracking, share-link, voice-context, model, and arrival-rotation checks passed.
- Manual local proxy check — a traffic road request returned HTTP 200 with `X-Overpass-Upstream: https://api.openstreetmap.org/api/0.6/map`, and the browser completed the Street Traffic load.

No API keys, environment files, cache files, logs, or generated artifacts are included in this PR.


bilawalsidhu:
OSM's API usage policy says the editing API is 'not for read-only purposes or projects' and directs data users to planet.osm or Overpass, and this makes /api/0.6/map the primary path for the app's highest-frequency query. Main already rotates four Overpass mirrors with a 7-day disk cache for the failure this targets. Closing.

PR #227, opened 2026-09-09
chore: adopt scoped formatting and infrastructure package boundaries

RohanDaCoder:
Adds a reproducible formatter setup and checks the dependency boundaries of the existing reusable infrastructure exports. `npm run format` and `npm run format:check` use the same explicit adopted-file list; CI enforces it on Linux and Windows. New modules can join that list as they are extracted.

RohanDaCoder's original Prettier contribution and merge commit are preserved. Maintainer commits pin Prettier 3.9.6, correct duplicate dependency entries and the orphan lock entry, add scoped tooling and consumer-boundary checks, and format the infrastructure component in a separate mechanical commit. The lockfile adds only Prettier to the accepted dependency baseline.

`npm run check:boundaries` builds all three package exports without loading application configuration and rejects imports outside the component's declared modules. Cesium stays external. Node services and standalone startup remain outside those browser exports. The check is intentionally scoped to existing exports; it does not claim that the entire application has been split into components.

Validation:

- Clean `npm ci`, setup doctor, formatting and package boundary checks pass.
- Unit/allocation suites: 2,909 passed, one platform skip.
- Production build, 23/23 infrastructure browser checks and 108/108 tracking checks pass.
- Formatting leaves the syntax trees of the four adopted infrastructure files unchanged. One source-text regression pin now tolerates line breaks/trailing commas while still requiring the same two stem buffers; runtime buffer-identity assertions remain intact.
- Browser inspection checks infrastructure labels and markers at multiple zooms/angles. Initial browser runs hit a stale Vite optimizer cache; fresh dependency optimization restored startup without a source change.
- An initial tracking run had four corridor/floor failures while live DEM data had not resolved. The unchanged isolated repeat passed all 108 checks; no tracking assertions or production terrain code were changed.
- Node 24.14.0, Node 26.x and Windows CI pass on the reviewed head.


samehkhamis:
Thanks, @RohanDaCoder, for contributing the Prettier setup! Merged with scoped formatting and package boundary checks.

PR #231, opened 2026-09-09
feat: add host-controlled community presence layer

maldoteth:
## Summary
- add an opt-in Community data layer fed by a host-provided JSON endpoint
- validate and cap public presence records before rendering avatar/initial markers
- expose a small postMessage placement bridge while leaving consent, precision, auth, retention, and deletion to the host platform
- add an embedded exit bridge and configurable base path for hosted integrations

## Verification
- node --test src/data/communityPresence.test.mjs src/data/layerState.test.mjs
- npm run build

The layer never requests or stores device location itself and fetches the public endpoint without credentials.


bilawalsidhu:
Three rule breaks: the layer fetches an endpoint taken from a query parameter or postMessage, the message handler checks no origin and posts back to '*', and it renders live positions of named people. It also carries a Pages deploy workflow and an ARCADE exit button for another product. Closing.

PR #235, opened 2026-09-09
feat: draw on the world by hand (DISPLAY ▸ Draw)

cora-fresh-labs:
## What

**DISPLAY ▸ Draw**: draw on the world by hand. Pick Area, Line or Pin, click the vertices on the real world, double-click or press Enter to finish, give it a label and a colour. Backspace removes the last vertex, Esc cancels the shape in progress, a second Esc leaves draw mode.

A drawn shape goes through the **same `annotationEngine.annotate()`** the voice agent uses, with its geometry supplied and `manual: true`. So it renders with the whiteboard look (draped area, flowing line, pin), persists, de-dups, shows up in `.list()` and the GeoJSON interchange, and clears with `clear_annotations` / `.clear()`. No geocoder, Places or Overpass call is made for a drawn shape.

## Why

The whiteboard today is voice-only, and a name is the only way in. Plenty of things people want to mark have no name on any map: the shed at the back of a wholesale market, a paddock, a yard, the far building. This lets a person point at exactly that and label it, and it lets a session build a board without a microphone (or a key).

## How

- `src/annotations/drawMode.js` (new, pure): the draw session, minimum vertices per shape, double-click de-dup at 0.5 m, length/area, hint text, and the `isDrawModeActive()` switch. Unit-tested.
- `src/annotations/drawTool.js` (new): the Cesium + DOM half. Vertices are world-anchored through the resolver's existing `pickWorldFromScreen` cascade (now exported), so a vertex on a roof lands on the roof. A dashed preview follows the cursor. The viewer's stock LEFT_DOUBLE_CLICK action is removed for the session and restored on exit. Exposed as `window.__gevDrawTool` for tests.
- `src/annotations/annotationEngine.js`: `resolveSpec` short-circuits manual specs (`isManualSpec` / `resolveManualSpec`). A drawn area is a real (not synthesized) outline; a drawn line is a `route` with `mode: 'manual'`, labelled with its length and never a travel time; a pin is a point.
- `src/data/trackingClickGesture.js`: while a session is open the gesture returns before `onClick`, so a vertex over an aircraft, military contact or camera is a vertex, not a selection.
- `index.html` / `style.css`: the Draw group in the DISPLAY panel, in the existing `pp-*` grammar (toggle, shape segment, label field, colour, one-line hint). Respects Clean UI like the rest of the panel.
- `src/main.js`: `initDrawTool({ viewer, annotations })` right after `initAnnotations`.
- Docs: `docs/CURRENT-STATE.md` (new section), `CHANGELOG.md`, `README.md` (one line under *Annotate it*).

## Verified

- `npm run build` green (Node 24.21).
- `npm test` green: 2,720 pass, 0 fail (+ the calibrated allocation suite, 13 pass).
- `npm run test:track` green against the dev server: 108 passed, 0 failed, 0 skipped (keyless boot).
- Headless Chromium (Playwright, ANGLE/SwiftShader): opened the DISPLAY panel, pressed Draw, four real mouse clicks on the Capitol grounds, double-click to finish → one `area` annotation, ring of 4, amber, `resolvedVia: manual`, `synthesized: false`, 3.3 ha in the hint; Esc twice leaves draw mode with the stock double-click restored. Same result on a keyless boot and with a Google key present.

## Notes for review

- Layers that bind their own LEFT_CLICK outside `bindTrackingClickGesture` (vessel and fire cards) still receive clicks during a draw session. Routing them through the same guard is a one-line follow-up if you want it; I kept this PR to the gesture module they already share.
- No new data source, no new key, nothing fetched.


bilawalsidhu:
Landed via #547 (rebased onto the current component layout, with fixes from review). Thank you @cora-fresh-labs!

Issue #212, opened 2026-09-09
Ollama/whisper support


Code4SAFrankie:
Since the major feature of the app seems to be to talk to it to get things done, can you please support open source local models via ollama and whisper. Fasterwhisper for STT and ollama for inference, instead of openai paid models. Ollama is openai api compatible.

thisismcbb:
Yeah, I think that would be useful too. I've already looked for the API endpoint – any pointers on where it's defined in the script would help.

drone540:
Support for Kokorotts-FastAPI, whisper,cpp, crisper.cpp, audio.cpp, would be great. 

They all support Open AI api compatible endpoints.

whisper.cpp just needs `--inference-path /v1/audio/transcriptions` for it to be compatible.

We can just specify base url server for Text to Speech and Speech to Text and Select voices 


### Kokoro also allows voice combinations for custom voices:

#### 1\. Single voice

The simplest case should continue to work:

    {
      "voice": "af_bella",
      "input": "Hello world."
    }
    

#### 2\. Composite/blended voice

Allow multiple Kokoro voices to be combined into a single synthesized voice:

    {
      "voice": "af_bella+af_sky",
      "input": "Hello world."
    }
    

#### 3\. Weighted composite voice

Allow individual voices to have different weights:

    {
      "voice": "af_bella(2)+af_sky(1)",
      "input": "Hello world."
    }
    

This would produce a 2:1 blend rather than a 50/50 blend.

So bonus points if that is also supported in some way. Like if a drop down voice selector allows editing the values.

olinux68:
I built a working local **Ollama voice + HUD** path in a fork: https://github.com/olinux68/gods-eye-view

When `AI_PROVIDER=ollama`, it swaps the OpenAI Realtime/Responses calls for a fully local, turn-based pipeline:
- **STT:** faster-whisper + Silero VAD
- **LLM / tool-calls + HUD:** Ollama `/api/chat` (qwen2.5:7b or llama3.1:8b recommended for function calling)
- **TTS:** Piper
- **Transport:** local WebSocket (`/api/voice/ws`) with browser action dispatch, instead of Realtime WebRTC

[...]

PR #237, opened 2026-09-09
feat(nav): keyless forward geocoding via a Nominatim proxy — fly_to_location works without a Google key

milu70hub:
Today `searchAndFlyTo` throws `No Google Maps API key available for geocoding` when there is no key, so voice/typed navigation only reaches the eight preset cities — the keyless default stops at "take me to Tokyo".

## What
- `server/keylessGeocode.js` — `GET /api/geocode?q=<place>[&bounds=s,w|n,e]` → Nominatim `/search` (jsonv2, limit 1, soft `viewbox` bias from the viewport), ≥1.1 s between upstream calls, 10-minute cache, policy User-Agent. Returns the Google Geocoding subset the client reads (`status`, `results[0].formatted_address / geometry.location / geometry.viewport / types`), with Nominatim `class/type/addresstype` mapped onto the Google `types` the navigation-mode heuristic understands (country → region-overview, city → city-overview, road → street-corridor, …).
- `searchAndFlyTo` — Google when a key exists (unchanged, Places recovery included); the proxy otherwise. No more throw.
- Tests: `src/keylessGeocode.test.mjs` (type mapping, result shape, bounds → viewbox, endpoint with mocked fetch, cache, 400/405). `src/locations.test.mjs` unchanged and green (its source-text assertions on `searchAndFlyTo` still hold).
- Docs: README key-table note, `CHANGELOG.md`, `docs/CURRENT-STATE.md`.

## Tried
Keyless container: typed and voice commands (with #212-style local provider) — "take me to Lisbon", a small coastal town, a street name with viewport bias — all fly with the expected framing. `npm test` and `npm run build` green.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ of current runtime behavior, see [`docs/CURRENT-STATE.md`](docs/CURRENT-STATE.md
 ### Added
 
 - Keyless forward geocoding: `fly_to_location` (voice, typed, and the
-  location bar) now resolves any place name without a Google Maps key through
+  location bar) and `annotate_map` targets / route waypoints now resolve any place name without a Google Maps key through
   `GET /api/geocode` (Nominatim, ≥1.1 s between upstream calls, 10-minute
   cache, policy User-Agent). Answers arrive in the Google Geocoding shape, so
   `searchAndFlyTo` keeps one code path; Google Geocoding + Places recovery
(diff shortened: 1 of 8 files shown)


milu70hub:
Pushed one more commit: `annotationResolver.geocodePlace` takes the same keyless branch, so `annotate_map` targets and route waypoints ("draw the walking route from A to B" → `fly_route`) resolve without a key as well. Tests green.

bilawalsidhu:
Keyless search is landing through #166, which uses Photon so the project stays inside the Nominatim usage policy. Closing in favor of that one.

PR #239, opened 2026-09-09
perf: subset the icon font, and drop the second icon family nothing uses

mml-studio:
## Why

Loading the app pulls **442,061 bytes from Google Fonts over six requests**, and
**330,256 of them are the complete Material Symbols Outlined variable font** —
served so the interface can draw 28 glyphs.

The sixth request is `Material+Icons+Round`: a second icon family that no source
uses. `src/cockpitMarkup.test.mjs:694` already asserts the markup must not use
it. Its 173 kB font never downloads today, because nothing activates the family,
but the stylesheet is fetched and render-blocking on every boot.

Measured in headless Chromium against `npm run dev`, cold profile, 2026-09-09:

| | requests | bytes |
|---|---|---|
| `main` | 6 | 442,061 |
| this branch | 5 | **115,319** |

## What changed

`index.html` asks Google for a subset with `icon_names=`. The same 28 glyphs
come back in **4,044 bytes** instead of 330,256, and the unused stylesheet is
gone. No new dependency, no vendored font, no build step — the CDN does the
subsetting.

## The part that can break silently, and the test that stops it

A glyph the subset does not carry gets **no missing-glyph box**. The ligature
never forms and the element renders the literal word, mid-cockpit. So the list
is not kept by hand: `src/materialSymbolsSubset.test.mjs` extracts the glyph
names from the sources and fails when one is missing from `icon_names`, naming
the file that asks for it.

It reads two shapes — the glyph as element text, and any string assigned to
`textContent`/`innerText`. The second is what catches
`icon.textContent = expanded ? 'right_panel_close' : 'right_panel_open'` in
`src/ui.js:1992`: the element is found by `querySelector('.material-symbols-outlined')`
and its glyph picked by a ternary, so `right_panel_open` appears nowhere in
`index.html`. A list written by reading the markup would have shipped the words
`right_panel_open` into the collapsed briefing panel.

The extractor errs wide on purpose. Google answers 200 and ignores an
`icon_names` entry it does not know (verified against the live endpoint), so a
false positive costs nothing, while a false negative costs the interface. Two
non-icon words ride along in the list for that reason.

## How it was verified

- `npm run build` — clean.
- `npm test` — 2,708 pass, 0 fail. (The two allocation microbenchmarks skip on
  Node 26.0.0, as `scripts/run-unit-tests.mjs` intends; no Node 24 on this
  machine.)
- `npm run test:track` — **does not finish on this machine, on `main` as much as
  on this branch**, so I have no green run to show either side. Three runs, zero
  `FAIL` lines in all three: `main` stopped after 9 assertions on a 30 s
  `waitForFunction` timeout at `scripts/track-regression.mjs:647` (the
  share-link reload waiting for the tracked aircraft to return); this branch
  stopped after 42 on a Puppeteer `ProtocolError`, then after the same 9 on the
  same timeout. Nothing in it touches fonts, and the branch never behaves
  differently from its base — but it is not the green run CONTRIBUTING asks for,
  so I would rather say so than imply otherwise.
- Removing `right_panel_open` from `icon_names` fails the new test with
  `right_panel_open (src/ui.js)`.
- In the live page, each of the 28 glyphs measures about **one em** wide via
  `canvas.measureText` under `20px "Material Symbols Outlined"`, against
  **369 px** for a name deliberately left out of the subset — so every glyph the
  app renders is really in the font, not merely requested.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,16 @@ of current runtime behavior, see [`docs/CURRENT-STATE.md`](docs/CURRENT-STATE.md
 
 ## [Unreleased]
 
+### Changed
+
+- The Material Symbols icon font is now requested as a subset of the glyphs the
+  app actually renders, and the unused Material Icons Round stylesheet is gone.
+  Loading the app pulled 442 kB from Google Fonts over six requests, 330 kB of
+  it the complete icon font served for the 28 glyphs on screen; it now pulls
+  115 kB over five. `src/materialSymbolsSubset.test.mjs` fails when a source
+  names a glyph the subset is missing, because an absent glyph renders as its
+  own name rather than as a missing-glyph box.
+
 ### Fixed
 
 - Mapped-site outages show their scheduled retry countdown and distinguish
--- a/index.html
+++ b/index.html
@@ -9,8 +9,11 @@
   <link rel="preconnect" href="https://fonts.googleapis.com" />
   <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin />
   <link href="https://fonts.googleapis.com/css2?family=JetBrains+Mono:wght@300;400;500;600;700&family=Inter:wght@300;400;500;600&display=swap" rel="stylesheet" />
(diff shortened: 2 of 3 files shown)


bilawalsidhu:
Landed via #513 (rebased onto the current component layout, with fixes from review). Thank you @mml-studio!

PR #242, opened 2026-09-09
fix(server): gate cost/log endpoints against cross-site requests and ship a full CSP

jasas1:
## Summary
Adds a same-site request gate to four dev-server endpoints that were not covered by the credential panel's `admitKeySetupRequest` gate: `/api/realtime/token`, `/api/openai/hud-summary`, `/api/google/nearby-places` and `/api/realtime/debug-log`. Replaces the frame-ancestors-only header with a full Content-Security-Policy on the dev and preview servers.

## Threat
With the dev server running, a hostile page open in the same browser can reach those endpoints cross-site. A `fetch` carries a foreign `Origin`; an `` carries no `Origin` at all but does carry `Sec-Fetch-Site: cross-site`. Either mints a paid Realtime token, spends OpenAI or Google quota, or appends attacker-chosen bytes to `.gev-logs/realtime-conversations.jsonl`.

## Change
- `src/localRequestGate.mjs`: pure `admitSameSiteRequest`. Refuses reverse-proxy signal headers, `Sec-Fetch-Site` other than `same-origin`/`none`, and a foreign or opaque `Origin`. It deliberately does not require a loopback socket, an `Origin`, or JSON, so `scripts/qa-voice-routing.mjs` and `scripts/qa-l9-matrix.mjs` (Node clients with no `Origin`) and the documented `HOST=0.0.0.0` LAN opt-in keep working.
- `vite.config.js`: one `admitSameSite(req, res)` helper called in the four handlers after the method check and before the opt-in rate limiter. `LOCAL_CSP` constant applied to `server.headers` and `preview.headers`.
- `src/keySetupCore.mjs`: `PROXY_SIGNALS` hoisted and exported so both gates share one definition. Panel behaviour unchanged.
- `SECURITY.md`: documents the gate and the CSP; drops the stale "two non-secret CCTV feature flags" claim.

`script-src` is `'self' 'unsafe-eval'`. Knockout, bundled in `@cesium/widgets`, resolves the global object with `(0, eval)("this")` at load time; with `'wasm-unsafe-eval'` alone the Cesium widget never initializes (verified in headless Chrome). Inline scripts and scripts from any other origin remain blocked.

## Compatibility
No new dependencies. Keyless fallbacks, opt-in rate limiters and the strict credential gate are unchanged.

## Testing
- `npm test`: 2720 pass, 0 fail (Node 22; allocation microbenchmarks skipped as expected). 14 new unit tests in `src/localRequestGate.test.mjs`.
- A probe boots the dev server keyless, sends cross-site, ``-style, proxied, same-origin and no-header requests to all four endpoints (403 / 403 / 403 / served / served), checks the CSP header, then loads the app in headless Chrome, clicks a first-run tile and records zero CSP violations with the Cesium canvas present.


Lob26:
Contributor review, not a maintainer verdict. Checked against `main` at `a8125ad` — still merges clean, just `BEHIND`.

This is the most carefully reasoned of the security PRs I looked at in the queue. The gate is a pure function with its refusals pinned by unit assertions, the `Sec-Fetch-Site` rationale is right (an `` or navigation carries no `Origin` but does carry `Sec-Fetch-Site: cross-site`, and that ordering is what catches it), and keeping the loopback-address requirement out of this gate while leaving it in `admitKeySetupRequest` is the correct distinction rather than a shortcut.

I went looking for CSP breakage specifically, because a too-tight `script-src` is the usual way a change like this lands and then gets reverted. Two hypotheses, both of which I was wrong about, recorded so nobody re-checks them:

- **`script-src 'self' 'unsafe-eval'` does not break Google Maps.** The Maps integration is `fetch` to REST endpoints (`src/search/http.js:56`, `src/search/defaults.js:49`), not a `<script>` tag, so it's governed by `connect-src 'self' https:` and is fine. Photorealistic 3D likewise arrives as tiles through Cesium's fetch path.
- **`vite-plugin-cesium` does not inject an inline script.** It sets `CESIUM_BASE_URL` through Vite `define` and a rollup `intro` (`node_modules/vite-plugin-cesium/dist/index.js:77`, `:86`) and injects only same-origin `<script src>` / `<link>` tags (`:130`, `:138`). No inline `<script>` needs a nonce, and `index.html` has exactly one external module script and no inline block.

So the CSP looks survivable, and `'unsafe-eval'` covering both Knockout's `(0, eval)("this")` and Cesium's WASM decoders is accurate.

**The one thing I'd want stated in `SECURITY.md`: this gate does not survive DNS rebinding in the `HOST=0.0.0.0` mode.**

In the default configuration the gate is defense in depth behind Vite's own host check — `build/vite.js:19` sets `allowedHosts: ['localhost', '127.0.0.1', '.local']`, so a rebinding `Host` never reaches a handler. That's fine.

But the LAN opt-in takes the same branch to `allowedHosts: true`, which disables the host check entirely. In that mode an attacker-controlled name resolving to the operator's address gives them a page where `Host` and `Origin` are both their own domain and agree, and `Sec-Fetch-Site` is a truthful `same-origin`. All three of the gate's checks pass, and `/api/realtime/token` mints a Realtime token. This isn't a flaw in the gate's logic — comparing `Origin` to `Host` is the right check, and the PR is explicit that it deliberately drops the loopback-address requirement to make the LAN opt-in work — but it is the reason the loopback + local-`Host` pair in `admitKeySetupRequest` exists, and dropping half of it costs the rebinding protection.

Concretely, I'd soften the `SECURITY.md` wording. "Refuse cross-site browser requests" is accurate for the threat in the title, but a reader hardening a LAN deployment will take it as broader than it is. One clause — that in the `HOST=0.0.0.0` mode the gate relies on the operator's network and not on `Origin`/`Host` agreement, because both are attacker-controlled under DNS rebinding — would close the gap between what it says and what it does. The existing `HOST=` key-exposure warning (`scripts/qa-l9-matrix.mjs:707` checks for it) is already the right place for operators to meet that idea.

**Two smaller notes.**

`admitSameSiteRequest` takes `method` but never uses it. Either drop it from the signature or gate on it — as written a reader assumes safe methods are handled differently and they aren't.

The ordering in `googlePlacesContextProxy` puts `admitSameSite` *after* the method check (`server/providers/places/google.js:57`), while `sameSiteGated` wraps the whole handler for the OpenAI routes. Minor, but it means a cross-site GET to nearby-places gets a 405 and a cross-site POST gets a 403, so the endpoint's method surface is discoverable cross-site. Gating first makes the two paths consistent.

**The gate I can't stand in for:** this is the change in the queue that most needs the maintainer workflow's runtime check. A CSP regression is invisible in unit tests and shows up only as a console violation and a dead globe. `npm run test:track` plus a console-error sweep on both `npm run dev` and `npm run preview` is what would actually clear it, and `connect-src` is where I'd look first — it has no `blob:`, so any `fetch()` of an object URL would fail.

What I verified: the merge state, the two CSP hypotheses above, `allowedHosts` in both branches, the unused `method` parameter, and the gate ordering, all by reading `main` at `a8125ad` and the vendored plugin. What I did not: I did not execute this branch, serve the app under the proposed CSP, or run any suite against it.

jasas1:
Thanks for the careful read. All three points were right, and your note about runtime checks led to a fourth. Pushed in the latest commit, merged up to `main` at `0dbde1e`.

**DNS rebinding in `HOST=0.0.0.0` mode.** Agreed, and the wording was broader than the behaviour. `SECURITY.md` now states the limit next to the gate: with the host check off, a rebinding page has an `Origin` and `Host` that agree and a truthful `same-origin`, so all three checks pass and protection rests on the operator's network.

**Unused `method` parameter.** Removed from the signature, the caller and the tests.

**Gate ordering in nearby-places.** The gate now runs before the method check, matching the wrapped OpenAI routes. A cross-site POST gets 403 instead of 405.

**The CSP did break something, in preview only.** I added a production-preview pass to the runtime probe because of your comment. In the built app Cesium's bundled workers bootstrap through `importScripts(blob:…)`, and a worker checks that against `script-src`, so workers failed under `vite preview` while the dev server (workers loaded by URL) was clean. `script-src` now includes `blob:`. I also added `blob: data:` to `connect-src`, which you flagged. Clean `main` under `vite preview` shows no such errors, so this was the policy, not upstream.

What was run against this branch: `format:check`, `check:boundaries`, `npm test`, `npm run build`, `npm run test:track` (109 passed, no console errors, served under the new policy), and a headless-Chrome probe in both `npm run dev` and `vite preview` modes. The probe sends cross-site, ``-style, proxied, same-origin and header-less requests to the four endpoints, checks the CSP header, boots the globe, clicks a first-run tile, and fails on any CSP violation, CSP-related page error, or worker script-load error.

Not covered: voice sessions and radio playback with real keys, since the probe runs keyless.

PR #243, opened 2026-09-10
fix: pin PSModulePath for the Windows credential ACL verifier (Provider Settings save fails under PowerShell 7)

jdholbrook33:
## Summary

On Windows, saving keys from Provider Settings fails with **"could not restrict the credential file to your account; nothing was saved"** whenever the dev server was started from **PowerShell 7 (pwsh)**. The `icacls` step succeeds; it is the follow-up verification that fails.

`hardenCredentialFile` verifies the DACL by spawning Windows PowerShell 5.1 with `Get-Acl`, and it inherited the parent's `PSModulePath`. A pwsh 7 parent exports a `PSModulePath` that lists its own module folders first, so Windows PowerShell resolves `Microsoft.PowerShell.Security` to the PowerShell 7 copy and cannot load it:

```
Get-Acl : The 'Get-Acl' command was found in the module 'Microsoft.PowerShell.Security',
but the module could not be loaded.
```

The hardener correctly fails closed, so no key is ever saved. The same cause makes the unit test `Windows production hardener applies its exact DACL with native tools` fail on any Windows host where `npm test` runs under pwsh.

## Fix

Pin `PSModulePath` for the verifier spawn to `<SystemRoot>\System32\WindowsPowerShell\v1.0\Modules`, the built-in directory that holds the only module the script needs. This is always `System32`, even for a 32-bit host that reaches the tools through `Sysnative`, because the spawned `powershell.exe` is the native 64-bit binary. Pinning also keeps user-writable module directories out of the verifier's search path, in line with the module's existing no-PATH policy for the native tools.

## Tests

- New: the verifier receives the pinned `PSModulePath` even when the parent exports pwsh 7 folders; other inherited variables are untouched.
- New: the 32-bit `Sysnative` route still pins the module path under `System32`.
- Existing Windows production hardener test now passes on a pwsh 7 host.

## Verification

Windows 11 Enterprise LTSC 2024, Node 24.21.0, session started from PowerShell 7.5:

- Reproduced the failure by spawning `powershell.exe` from Node with the inherited environment (status 1, module load error) and confirmed the same command succeeds with the pinned module path.
- `npm test`: 2707 pass, 0 fail, 2 skipped (previously 1 fail).
- `npm run build`: succeeds.

https://claude.ai/code/session_011bEUPtCMarBD1dmtWEDhQb


bilawalsidhu:
Same fix as #161, which arrived first and is merging. Thanks for the clear repro; closing this one to keep a single change.

PR #245, opened 2026-09-10
Security/local server hardening

U007D:
Numerous security fixes (analyzed and performed by  `codex` on `gpt-6-astra-xtra-high`)


bilawalsidhu:
Read in full. It refuses every non-loopback request, which removes the documented LAN opt-in, moves the voice log out of .gev-logs, and rewrites README/SECURITY to match; the GBFS parts already landed separately. The Puppeteer sandbox-flag cleanup in the QA scripts is worth its own small PR. Closing this one.

PR #253, opened 2026-09-10
fix(cctv): validate and bound the client Range header in the media proxy

### LINKED ISSUE #27: Validate or strip client Range headers in the CCTV media proxy

Problem:
- `/api/cctv/media/:id` forwards the incoming `Range` header to the configured upstream source.
- The route does not validate range syntax or bound requested spans before forwarding.

Evidence:
- `vite.config.js:4562-4565` copies `req.headers.range` into upstream request headers.
- `vite.config.js:4543-4611` exposes the media proxy route.

Suggested fix:
- Either strip `Range` entirely or parse and bound it before forwarding.
- Reject multi-range and malformed values.
- Ensure returned `Content-Range`/status behavior matches the validated request.

Maher-Reven:
Closes #27.

`/api/cctv/media/:id` copies `req.headers.range` straight into the outbound request to a third-party camera host — no syntax check, no bound on the span:

```js
const requestRange = req.headers?.range;
if (requestRange) upstreamHeaders.Range = requestRange;
```

## What that actually does

Both of these reproduce on clean `main` against a stubbed upstream (I drove the real route through the plugin rather than reasoning about it):

```
client Range                   | forwarded upstream           (clean main)
-------------------------------------------------------------------------
"bytes=0-1,2-3"                → "bytes=0-1,2-3"
"bytes=abc-def"                → "bytes=abc-def"
"bytes=0-10\r\nX-Evil: 1"      → "bytes=0-10\r\nX-Evil: 1"
"bytes=0-999999999999"         → "bytes=0-999999999999"
```

**1. Multi-range defeats the response size cap.** `bytes=0-1,2-3` is relayed, so the upstream answers `multipart/byteranges`. `proxyMediaResponse` forwards `Content-Length`/`Content-Range` blindly and caps on a *single* declared body, so a multipart response walks straight past the accounting that cap exists to provide.

**2. A CRLF value lets any local caller write into `/api/cctv/health`.** This is *not* header injection — undici rejects the value at `Headers.append`, which I verified. But the throw happens inside the route's `try`, so it answers 502 and records the client's **raw string** as that camera's health message:

```
GET /api/cctv/media/<id>   Range: bytes=0-10\r\nX-Evil: 1
→ 502 {"error":"Media proxy failed"}

GET /api/cctv/health
→ {"status":"degraded",
   "message":"Headers.append: \"bytes=0-10\r\nX-Evil: 1\" is an invalid header value."}
```

So anyone who can reach the dev server can flip a chosen camera to `degraded` and put text of their choosing into the health report — which cuts against SECURITY.md's *"Sanitized errors — internal error details are not echoed back to clients."*

Impact is bounded by exposure, as #27 says: this is a dev/preview server that should be on loopback. It's a hardening fix, not an emergency.

## The fix

`sanitizeCctvRangeHeader` parses the header before anything is forwarded:

```
client Range                   | forwarded upstream           (with fix)
-------------------------------------------------------------------------
"bytes=0-1023"                 → "bytes=0-1023"       unchanged
"bytes=500-"                   → "bytes=500-"         unchanged  (seek)
"bytes=-500"                   → "bytes=-500"         unchanged  (suffix)
"bytes=0-1,2-3"                → (none)               dropped
"bytes=abc-def"                → (none)               dropped
"bytes=0-10\r\nX-Evil: 1"      → (none)               dropped
"bytes=500-100"                → (none)               dropped    (inverted)
"bytes=0-999999999999"         → "bytes=0-67108863"   clamped
```

Two deliberate choices worth reviewing:

- **Drop, don't 416.** RFC 7233 §3.1 says a server ignores a Range it cannot understand or satisfy, so an unusable value just proceeds as a full GET. That keeps the change non-breaking — no client that works today starts failing.
- **Clamp explicit spans, leave open-ended ones alone.** An explicit span wider than `MEDIA_DECLARED_CAP_BYTES` could *only* ever end in the 502 that cap already raises, so clamping turns a guaranteed failure into ordinary partial content. `bytes=500-` and `bytes=-500` are left unclamped on purpose — neither asks for more than a plain GET would return, and clamping them would break seeking in the video feeds this route exists to serve.

`MEDIA_DECLARED_CAP_BYTES` moves from a function-local `const` to module scope so the Range bound and the response cap can't drift apart. That's the only change to existing behaviour outside the header path.

## Verification

- `npm test` — **2713 tests, 0 failures** (6 new)
- `npm run build` — clean (only the pre-existing chunk-size advisory)
- `npm run doctor -- --json` — exit 0

Tests live in their own `src/data/cctvMediaRange.test.mjs` rather than in `cctvProxy.test.mjs`, so this doesn't collide with #185 or #222, which both edit that file. They cover the pass-through forms, multi-range, malformed/injection shapes, non-string and duplicate-header (array) inputs, span clamping at and around the cap, inverted ranges, and unsafe-integer positions.

## Notes for review

- **#245 overlap is textual only.** It rewrites the two lines on either side of this one (adds `Accept-Encoding: identity`, `redirect: 'error'`, `signal`) but leaves the `requestRange` pair untouched, so whichever lands second needs a small manual merge. No behavioural conflict.
- **One thing I did not fix:** `setHealth` stores raw `error.message` generally, so other upstream errors can still put upstream-controlled text in `/api/cctv/health`. This PR closes the client-controlled path into it; sanitizing that field looks like it belongs with #245's broader server hardening rather than here. Happy to add it if you'd rather it land in one piece.


bilawalsidhu:
Landed via #512 (rebased onto the current component layout, with fixes from review). Thank you @Maher-Reven!

PR #236, opened 2026-09-09
feat(voice): local voice provider (Ollama / Anthropic) driving the same 28 tools — #212

milu70hub:
Answers #212 (Ollama / local model support) without touching the Realtime path or the tool schema.

## What
- `server/localVoice.js` — `GET /api/local-voice/status`, `POST /api/local-voice/turn`. Provider-neutral transcript in, `{ text, toolCalls }` out. Providers: **Ollama** (`/api/chat` with tools, `think:false`) and **Anthropic** (`/v1/messages` with tools). Selection via `LOCAL_VOICE_PROVIDER=auto|ollama|anthropic`, `OLLAMA_URL` / `OLLAMA_MODEL`, `ANTHROPIC_API_KEY` / `ANTHROPIC_MODEL`, `LOCAL_VOICE_LANG`.
- `src/voice/localVoice.js` — `LocalVoiceController`, same surface as `GevRealtimeController` for `initGevVoiceCommands` and the mic UI. Ear = Web Speech API (continuous, restarted after `end`, paused while the app speaks), brain = the turn endpoint with a bounded tool loop through the SAME `createGevActionRunner`, mouth = `speechSynthesis`. Adds a small typed-command box and an on-screen reply line under the mic (typing `?` asks the model for a command summary), so the provider also works without a microphone.
- `initGevVoiceCommands` keeps building the OpenAI controller, then probes the status route and rebinds the mic when a local provider is configured and the browser can transcribe.
- The voice instructions moved into one shared constant `GEV_VOICE_INSTRUCTIONS` (used by both providers) — text unchanged; `GEV_REALTIME_TOOLS` is byte-identical, schema pins untouched.
- Docs: README (provider row + how-to), `.env.example`, `CHANGELOG.md`, `docs/CURRENT-STATE.md`. Tests: `src/voice/localVoiceServer.test.mjs` (provider resolution, schema conversion, history sanitising, Anthropic `tool_result` grouping, both endpoints with a mocked fetch).

## Tried
- Ollama with a Qwen3 30B-A3B instruct model on a laptop, app running in Docker on a home NAS (PR #208 recipe) over a private network. Spanish commands map cleanly: "enséñame los terremotos y llévame a Lisboa" → `set_layer_visibility` + `fly_to_location`; after the tool result the model confirms in one Spanish sentence; "aléjate un poco" → `adjust_camera_zoom{out, medium}`.
- iPhone Safari needs HTTPS for the microphone (documented in README).
[...]

PR #275, opened 2026-09-10
feat(voice): add end-to-end LocalAI Realtime provider

devv-shayan:
Local voice required either browser speech APIs or a cloud Realtime session. This adds a complete LocalAI WebRTC path that runs voice activity detection, speech recognition, tool selection, and speech synthesis locally while preserving the existing 28 GEV tools and OpenAI path.

The mic panel stores a CLOUD/LOCAL preference, reports LocalAI startup and model-loading state, and keeps each live session pinned to the provider it started with. On a first run, choosing LOCAL opens Provider Settings when setup is missing. If Homebrew is required, returning to the browser rechecks LocalAI and reveals INSTALL without a reload. Local sessions receive the existing instructions and tools before their microphone is unmuted. Tool follow-ups use buffered LLM output so MiniCPM function markup cannot reach Kokoro, and the browser cannot select arbitrary models through the local relay.

The server integration follows the current provider boundaries: `server/providers/localai.js` owns process lifetime, readiness, preloading, and same-origin SDP relay; `server/standalone/local-voice.js` owns the in-app installer. Only loopback targets are started automatically, owned child processes are stopped with Vite, failed preloads can be retried, and Provider Settings tears down pending setup work with the rest of the app.

The Apple Silicon reference profile uses Silero VAD, Parakeet Realtime EOU 120M, MiniCPM5-2B MLX 4-bit, and Kokoro. Setup is derived from the YAML stages, downloads the LLM before first use, and verifies installed configs and weights. The LocalAI 4.9 thinking compatibility is separate from the MiniCPM-only MLX-LM parser; the former is skipped when the backend already contains the fix merged in [LocalAI #11962](https://github.com/mudler/LocalAI/pull/11962).

Related to #212. PR #236 is another implementation of that request using browser STT and TTS; this PR provides a fully local speech-to-speech path over LocalAI WebRTC.

Validation:

- `node --test src/voice/*.test.mjs` — 343 passed
- Focused server, setup, Provider Settings, and lifecycle integration suite — 362 passed
- `npm run build`
- `npm run format:check`
- `npm run check:boundaries` — 628 modules and 42 portable entries passed all package boundaries
- `npm test` — 3,737 passed and 1 platform skip across 3,738 tests with normal loopback access
- Manual local run: STT → MiniCPM tool calls → GEV actions → Kokoro output; thinking disabled and function markup is not spoken


devv-shayan:
hey @bilawalsidhu and @samehkhamis  I’ve got the local voice path working now with device-aware Minicpm5 selection and the setup flow integrated, before I take this further, do you want to keep moving in this direction for local voice support, or would you prefer a different architecture strategy for the project?

PR #279, opened 2026-09-10
feat: Web Receivers layer — find, tune, and watch KiwiSDR/WebSDR/OpenWebRX from the globe

DH5DAX:
## Summary

Adds **Web Receivers**: internet-controllable radio receivers (KiwiSDR, WebSDR, OpenWebRX) as a data layer that behaves like Radio — globe markers, a companion panel, and voice tools — so you can ask *"which SDRs around Berlin cover 20 meters?"*, *"tune to 14233 kHz USB on this receiver"*, or *"show me the RF spectrum from 10 to 15 MHz around here"*.

### Data
- `/api/web-receivers/catalog` (vite.config.js) merges Receiverbook's map data with the community KiwiSDR feed (band coverage, user slots, antenna, online state). 30-minute cache, seven-day stale fallback, pinned outbound requests to two server-registered URLs, private/local receiver URLs dropped. Live result: ~1850 receivers.
- websdr.org's own list carries an explicit no-reuse notice and is **not** fetched; WebSDR entries come via Receiverbook with coverage inferred from labels (and reported as unknown when absent).
- Attribution added to DATA_SOURCES.md and `src/data/dataCredits.js`.

### Tuning and spectrum
- Tuning is a URL per family (`?f=…`, `?tune=…`, `#freq=…`) loaded in a panel dock or a new tab. GEV never proxies audio or control; the panel says the receiver operator sees your browser directly.
- Spectrum-only view: KiwiSDR pages take zoom, spectrum-panel and `mute=1` in the URL, so a range opens as a silent, zoomed waterfall. WebSDR/OpenWebRX cannot be muted or zoomed by URL; the result says `muted: false` with a note so narration stays honest.

### Voice
`find_web_receivers`, `tune_web_receiver`, `show_rf_spectrum`; `set_layer_visibility`/`show_data_layers_menu` enums gain `web-receivers`; aliases like "SDRs", "KiwiSDR". Tool-inventory pins (`radioMarkup.test.mjs`, `firstRunExperience.test.mjs`) re-derived for the deliberate schema change.

### Panel
Left-stack panel with the same layout contract as CCTV: family/band filters, receiver card, TUNE/OPEN row, SPECTRUM row, embedded dock with minimize/close. Keyboard focus is reclaimed while a receiver page boots (OpenWebRX grabs it) so voice push-to-talk keeps working; a deliberate click into the dock is left alone.

## Test plan
- [x] `npm test` — 2738 tests green, including new `webReceiversProxy.test.mjs` (parser, merge, SSRF/redirect refusal, stale/degraded) and `webReceiverTuning.test.mjs` (frequency parsing, band inference, tune/spectrum URLs, ranking)
- [x] `npm run build`
- [x] `node scripts/qa-web-receivers.mjs --url http://localhost:4173` — headless proof: layer enable (1846 markers), filters, all three voice tools through the real action runner, dock open/close, teardown; no receiver is contacted by the harness
- [ ] Manual: enable the layer, click a marker, tune a KiwiSDR from the panel, run the three voice phrases above with a mic

https://claude.ai/code/session_01UR4yD1xFaDHNZ2WfhYvSo4


bilawalsidhu:
Sorry for the wait, @DH5DAX. This is the listening foundation we want for radio.

Main has moved a lot since you opened it (providers now live under `server/providers/`, and `vite.config.js` is a thin shell). Could you rebase onto current main? Once it's green, it's first in line for review in the radio lane.

Two connections:
- @radioelf (#691) wants to add web receiver points from rx-tx.info. Worth agreeing whether those feed your catalog.
- ATC tuning (#715) will use receivers that cover the airband.

On #374: love it. Once this lands, could you split it into one PR per layer, starting with repeaters? @pbloss001 opened #380 about licensed repeater data and offered to help. Big picture: #714.

DH5DAX:
Rebased — as a re-port onto the current structure rather than a rebase, since nothing of the old wiring survived the move: portable rules in `src/sources/webReceivers.js`, the directory proxy under `server/providers/web-receivers/`, the layer factory under `src/layers/webReceivers/` (radio layer as the template), the panel controller under `src/ui/`, voice schemas/descriptions split the way the other tools are. The PR description is rewritten to match. Green against `npm test`, `check:boundaries`, `format:check`, `build` and `test:track`; the headless proof (`scripts/qa-web-receivers.mjs`) runs the layer, panel, dock and all three voice tools against the live dev server without contacting a receiver.

On the two connections:

- **rx-tx.info (#691):** happy to take those points as a third catalog source — the proxy merges Receiverbook and the KiwiSDR feed by receiver URL, so a third directory joins the same way. One condition: their terms must allow reuse (websdr.org's list does not, which is why it is deliberately not in the catalog). @radioelf, if that split works for you — you keep the sondes, your rx-tx.info points feed this catalog — I'll add the adapter once this lands, or you add it on top; either way.
- **ATC (#715):** every receiver carries its parsed band list ("airband" resolves to 118–137 MHz) and `receiverCoversHz(receiver, hz)` answers whether it covers a frequency, so "nearest receiver that covers 121.9 MHz" is a filter over the catalog. That export stays stable.

On #374: agreed — one layer per PR after this lands, repeaters first.

bilawalsidhu:
Hey @manjunath22466, could you please review this and take a look? It's the Web Receivers layer, the listening half of #714. Heads-up: #732 landed right after this rebase, so there are conflicts in the shared voice-tool and catalog registries (both sides additive).

DH5DAX:
Rebased onto `1fc7955` with #732 in: the conflicts were the pinned tool digests, the layer-id mapping in `toolDescriptions.js`, the catalog control ids and the shell imports — both additions kept side by side (`local-adsb` and `web-receivers`), pins re-derived, the legacy-schema test still reproduces the old digest with both sets of additions stripped. `npm test` (4,929), `check:boundaries`, `format:check`, `build` and the headless proof are green on the new head; the branch is one commit over `main`.

DH5DAX:
Rebased onto `ce671ce` — conflict-free again, so this is ready for @manjunath22466 to look at.

The conflicts were with #736 (Radio QA harness), #737 (Fire Perimeters) and #706 (Cyber HUD): the changelog, the three pinned tool digests and two registry counts. Both sides kept side by side, pins re-derived. The unchanged-tool digest in `radioMarkup.test.mjs` came back as the same `9db1b4c24d0b37fd` this branch pinned before the rebase, which is the proof that nothing but the declared additions moved.

One thing worth a maintainer's eye: **#737 took share-link token `2`, which this branch was using**, so Web Receivers moved to `3`. That is exactly what @daikaginza described in #658 — the letters are gone and the digits are now going the same way. Whichever of the two lands second will need another bump unless tokens get assigned explicitly.

Green on the new head: `npm test` (5,068), `check:boundaries`, `format:check`, `build`, and the headless proof `scripts/qa-web-receivers.mjs`.

manjunath22466:
@DH5DAX I picked up #279 to help with the remaining receiver issues before merge. I've prepared a follow-up patch against `2b5983a3202409731d9045ce3c3b2ce1e9029389`. It's still local and isn't part of this PR.

The submitted head still has a few functional issues. The patch covers:

- Keeping timeouts active while reading response bodies, cancelling rejected responses, and adding a shared retry cooldown with stale-cache fallback. A stalled body or repeated failed refresh exposes these cases.
- Keeping published coverage separate from inferred or unknown ranges, so strict frequency/range searches only return confirmed matches. It also changes aviation AUTO from NFM to AM, including 121900 kHz.
- Stopping old catalog requests from clearing newer requests, and old voice actions from changing the next receiver session after disable/re-enable. Find, tune and spectrum also stop if enablement fails or hasn't settled.
- Reporting tab requests and frame loading, failure, timeout and unknown outcomes without claiming tuning succeeded. Closing the dock, disabling the layer or destroying the controller unloads the receiver and releases its resources.

The local patch also ranks uncertain alternatives separately, so they don't disappear behind 50 confirmed matches.

The 174 focused tests pass. Controlled browser checks cover loaded-page timers and connections stopping, plus remounting without old events affecting the new page. The redirect-cleanup case uses an injected stream whose cancellation stays pending; I haven't reproduced a live outage through the default transport.

There are still decisions we need to settle before merge:

- Can you point me to the terms or permission covering Receiverbook and KiwiSDR registry retrieval, caching, normalization and redistribution, including attribution and request frequency? Public access and the parser's code license don't settle those uses. We should remove or replace a source if we can't establish that basis.
- How should we handle the plain-HTTP Kiwi feed? We need an agreed integrity policy using authenticated publisher-controlled data, or to leave that enrichment out. We also need rules for which receiver schemes/hosts we allow, when the user authorizes navigation, and framing restrictions.
- What should own audio when switching between the dock and Radio/local FM? We need handoff in both directions, including failure, timeout, close and teardown. Mute URL parameters don't prove silence or playback control. External tabs need to be explicitly unmanaged or left out of the managed listening flow.

I haven't verified live tuning/audio or third-party framing and cross-origin compatibility. The full test suite, tracking checks and built-app acceptance still need to run before we call this ready to merge.

Would you like the follow-up patch, and how would you prefer to receive it?

DH5DAX:
Withdrawing this one, and @manjunath22466's first question is the reason.

I went looking for the basis you asked for and could not produce it. Receiverbook states no terms of use at all, including on its Impressum. The KiwiSDR list is the dyatlov map maker's output: the code is GPL-3.0, which covers the parser and not the data, and kiwisdr.com's own public registry states nothing either. So all three levels are silent. What this PR's `DATA_SOURCES.md` actually asserted was "no reuse restriction stated", and you are right that this is not the same thing as permission to retrieve, cache, normalize and redistribute. I am not going to ask this project to carry that on my say-so, so the honest move is to close rather than to argue it.

Your second question has a harder answer than I expected: `rx.linkfanel.net` offers no TLS at all, port 443 refuses the connection, so the plain-HTTP feed cannot be fixed by switching schemes. It would need exactly the integrity policy you describe, for a source with no stated terms. That is not a decision this project should have to take for my layer.

**So: no need for the patch, thank you for offering it.** I verified your functional claims against the submitted head before writing this, and you were substantially right. Five confirmed, including one I think is worse than you described, and several where the code was better than the claim. Since the PR is closing, here is the part that does not close with it.

**Two of your findings apply to shipped code in `main`, not just to this branch:**

1. `server/providers/radio/catalog.js:91-92` — the `!response.ok` path throws without cancelling `response.body`, while the redirect path immediately above it does cancel. Every non-2xx answer from a Radio Browser mirror strands the upstream socket until the agent's idle reap. `server/providers/firePerimeters.js:66-69` and `server/providers/cyclones.js:269-272` cancel on `!ok`; the radio provider is the outlier.
2. `server/providers/radio/catalog.js` has the cache plus 7-day stale fallback but no failure cooldown: nothing advances `catalogCache.cachedAt` on failure, so once the cache window lapses every inbound request re-attempts the mirrors while they are down. `refreshPromise` only de-dupes concurrent callers. `server/providers/cyclones.js:325-326` is the house pattern (`if (now() - attemptedAt < 60_000) throw new Error('cyclone_retry_later')`, with the stale fallback behind it), as do the two aircraft providers.

For the record, the radio provider gets your claim 1a right where mine got it wrong: it `await`s the capped read inside the `try`, so its abort timer stays armed across the body. Mine returned the reader's promise from inside the `try`, which cleared the timer before a byte arrived and, because the shared refresh promise then never settled, made its own stale fallback unreachable. `server/providers/regional/http.js:8-17` already documents that exact trap. Worth a lint or a shared helper so the next provider cannot reintroduce it.

One other thing that is not mine and may be worth a check: nothing picks AM for the 118–137 MHz airband. Wherever a default demodulation mode gets chosen, air-ground VHF voice is double-sideband AM, so an FM demodulator there yields unintelligible audio.

@bilawalsidhu — this closes the listening half of #714 from my side, and #691 is now the cleaner route to it: @radioelf would need the same licensing basis, but starting from the question rather than from a built layer. The mapping half is unaffected; #745 (Repeaters) stands on a different footing, since HamRig's operator granted this project use of HamRig's own tables in writing, and the one place that permission cannot reach is marked `unverified` in every row and switchable off.

Thanks again for the depth of the review. It was more useful than a merge would have been.

DH5DAX:
Housekeeping: I am removing this branch from my fork, so the diff here will stop resolving. That is not a retraction of the explanation above, which stands as written. The data owner's decision is to keep this work local only.

PR #280, opened 2026-09-10
feat(cctv): add TxDOT camera pack — doubles Austin coverage (250 → 497)

ifnull:
Adds TxDOT highway cameras as a fourth keyless CCTV source pack, alongside Austin, Caltrans, and TfL. Roughly **doubles Austin-area camera coverage**.

## Cameras this adds

Measured within 60 km of downtown Austin, at default caps:

| | Before | After |
|---|---|---|
| **Austin-area total** | **250** | **497** |
| City of Austin (existing) | 250 | 250 |
| TxDOT highway (new) | — | 247 |

The Austin district publishes **282 cameras, 254 currently `Device Online`** — all 282 geocoded. The pack registers the online ones.

Texas has 25 districts, so the ceiling is much higher than what ships by default:

| District | Cameras | Online | | District | Cameras | Online |
|---|---:|---:|---|---|---:|---:|
| HOU Houston | 1,168 | 1,090 | | LBB Lubbock | 59 | 59 |
| DAL Dallas | 835 | 763 | | BRY Bryan | 54 | 47 |
| FTW Fort Worth | 490 | 480 | | PHR Pharr | 50 | 48 |
| SAT San Antonio | 308 | 300 | | YKM Yoakum | 47 | 40 |
| **AUS Austin** | **282** | **254** | | ABL Abilene | 41 | 39 |
| ELP El Paso | 258 | 215 | | ATL Atlanta | 30 | 30 |
| BMT Beaumont | 239 | 213 | | PAR Paris | 29 | 29 |
| LRD Laredo | 117 | 116 | | AMA Amarillo | 28 | 27 |
| WAC Waco | 94 | 81 | | TYL Tyler | 27 | 26 |
| CRP Corpus Christi | 87 | 84 | | WFS Wichita Falls | 25 | 20 |
| ODA Odessa | 74 | 73 | | LFK Lufkin | 14 | 14 |
| | | | | SJT / BWD / CHS | 6 / 4 / 4 | 5 / 4 / 4 |
| | | | | **Statewide** | **4,370** | **4,060** |

Default is `AUS` only — `CCTV_TXDOT_DISTRICTS` opens up the rest.

## Finding the feed

TxDOT's cameras moved off the MapLarge/skyvdn path public tooling used to scrape. That catalog endpoint now returns `403`, and the HLS streams return `401` regardless of `Referer` or `Origin`. The ITS site they migrated to serves the same cameras from two open, keyless, session-free endpoints:

```
GET /its/DistrictIts/GetCctvStatusListByDistrict?districtCode=<D>
GET /its/DistrictIts/GetCctvSnapshotByIcdId?icdId=<id>&districtCode=<D>
```

Frames are not an image body — the snapshot endpoint answers JSON carrying a base64 JPEG. `decodeTxdotSnapshotPayload()` unwraps it and validates the JPEG SOI marker so a truncated or non-image payload can never be served as `image/jpeg`, and `fetchCctvImageFromUpstream()` accepts that shape **only** from the official `its.txdot.gov` origin. Frames observed ~1 minute fresh.

## Two things worth reviewing

**Only `Device Online` cameras register.** Offline and errored devices still return a frame, sometimes years stale. That looks live, which is worse than the synthetic placeholder the fallback chain already provides.

**`dirDescription` is deliberately unused for heading.** It's the *roadway's* direction, not the camera's facing — it reads `"North"` for 249 of 282 Austin rows, including every camera on an east-west highway. Passing it to `directionToHeading` with `allowBare` would point ~88% of the pack due north at high confidence, which is the exact false-confidence failure `directionText.js` documents from the 2026-07-04 review. Heading instead comes from an explicit travel token in the camera name in strict mode (20 of 254), with the id-hash fallback at low confidence for the rest — the same posture as headingless Austin and TfL cameras.

## The second commit is a separate decision

`92de871` reorders the merge so both Austin-area packs claim registry slots before other metros. The global `CCTV_MAX_SOURCES` cap is a plain slice, so whichever pack sits last absorbs the entire truncation — with TxDOT last that was 4 Austin cameras out of 100 surviving rows.

**This trims the TfL tail from 250 to 96 at default caps.** That's a real change for London users, which is why it's split out — happy to drop it if you'd rather not take it. `CCTV_MAX_SOURCES=1100` seats all four packs whole at 1,054 rows, inside the existing 1200 ceiling; I verified that.

## Licensing

No camera data is vendored — the catalog is fetched at runtime per CONTRIBUTING. `its.txdot.gov` serves no `robots.txt` directives, the endpoints need no auth, and TxDOT's site disclaimer is liability-only with no restriction on automated access or reuse. Same posture as the existing Caltrans pack. `DATA_SOURCES.md` updated with the attribution row.

## Verification

- `npm run doctor -- --json`, `npm test`, `npm run build` — all green on **both** commits, checked independently so history stays bisectable
- `npm test`: 2,730/2,731 pass, 0 failures — 10 new tests in `src/data/txdotProxy.test.mjs`
- End-to-end through the proxy: live 1080p frames decoding correctly, verified against Austin and Dallas cameras
- `npm run test:track`: 3 pre-existing failures on my machine, all from a local Google Map Tiles `403` (my key lacks Map Tiles API access). Confirmed against a stashed baseline that this branch introduces **no** new failures — the `display-floor` checks are nondeterministic in that no-tileset state.

## Screenshot
Screenshot 2026-09-10 at 5 40 02 PM


bilawalsidhu:

We're combining the two Texas PRs (#280 and #423): #280's design with districts as a setting, #423's code since it's already on the new layout. Say so if you'd rather it went differently.

ifnull:
Sounds good. Let me know if you need anything from me. Thanks!

bilawalsidhu:

PR #284, opened 2026-09-11
Fix trackpad pinch-to-zoom on macOS

jrmagnus:
## Summary

- macOS reports a trackpad pinch gesture as a `wheel` DOM event with `ctrlKey: true` (both Safari and Chrome do this so pages can implement their own pinch-zoom instead of the OS zooming the whole page). Cesium's default `screenSpaceCameraController.zoomEventTypes` only binds plain `WHEEL` (no modifier), so pinch events were silently dropped — only two-finger scroll (plain wheel, no `ctrlKey`) drove zoom.
- Adds a `{ eventType: CameraEventType.WHEEL, modifier: KeyboardEventModifier.CTRL }` binding so Cesium recognizes pinch as a zoom gesture.
- Cesium converts wheel `delta` into camera movement with a single fixed formula regardless of source. macOS reports a much smaller `deltaY` per unit of finger movement for pinch than for a real two-finger scroll, so even with the binding above pinch felt like it barely moved the camera. There's no public Cesium knob to rescale just the CTRL+WHEEL case, so a capture-phase listener on `viewer.container` intercepts pinch wheel events ahead of Cesium's own listener and re-dispatches them with `deltaY` amplified (8x) before Cesium processes them. Regular scroll (no `ctrlKey`) is untouched.

## Test plan

- [x] Verified two-finger trackpad pinch now zooms the globe in/out on macOS (Chrome)
- [x] Verified two-finger scroll (no pinch) still zooms as before, unaffected
- [ ] Spot-check on Safari/macOS and a non-Mac / mouse-wheel setup, since I only tested Chrome/macOS


samehkhamis:

PR #285, opened 2026-09-11
feat: live news dock with embedded broadcast streams

codeclawd:
## What

A floating **LIVE NEWS** panel: press `N` or the new top-center TV button, pick a channel tab, and a YouTube live player plays beside the globe. No key needed.

Channels: Al Jazeera English, Sky News, DW, France 24, TRT World, Euronews, ABC News, CBS News, NBC News, Bloomberg, LiveNOW from FOX, Fox News, CNN, Reuters, WION, Al Jazeera Arabic, NASA ISS.

## Why

The app already puts flights, ships, fires, and cameras on one globe. Rolling news is the signal people check next when something happens at the place they are looking at, and every channel above streams on YouTube for free.

## How

- `src/newsPanel.js` renders the tab strip and one `<iframe>`. Last channel and dock position persist in localStorage. Closing blanks the iframe so audio stops.
- `newsLiveResolver` (`vite.config.js`) serves `GET /api/news/live?channel=<id>`. It fetches `youtube.com/channel/<id>/live`, reads the canonical `watch?v=` id and the `"isLive":true` marker, and caches for ten minutes. The dock embeds `/embed/<videoId>` when live and uses `/embed/live_stream?channel=` otherwise. I started with the channel form alone; NBC, ABC, LiveNOW, Fox, and CNN run several streams at once and it answered "Video unavailable" for all of them. Resolving the id server-side fixed every one.
- The dock joins `LEFT_STACK_OBSTACLE_SELECTOR`, so the adaptive accordion ends above it instead of overlapping. It drags by its header through `_makePanelDraggable` and lives in the panel z band.
- Fox News and Reuters have no 24/7 feed. Their tabs render dashed and YouTube shows its own card while they are off-air.

## Tests

`npm test` 2713 pass / 0 fail (six new: channel lookup, embed URL shape, resolver validation, live-page parsing, cache, off-air). `npm run build` clean. `npm run test:track` 108 passed / 0 failed. Verified in Chrome: tab switching, `N` toggle, persistence across reload, accordion clearance with DATA LAYERS expanded, drag and restore, NBC playing through the resolved id.

`CHANGELOG.md` and `docs/CURRENT-STATE.md` updated.

## Two things worth your call

1. The resolver parses a YouTube page rather than calling an API. If YouTube changes the canonical tag the dock degrades to the channel-form embed, which still works for single-stream channels. An API-key path (YouTube Data API `search?eventType=live`) would be sturdier but breaks the keyless promise.
2. The channel list is editorial. Happy to move it to `config/` or trim it to whatever set you want to stand behind.


bilawalsidhu:
Thanks for the idea. Embedding live news by scraping YouTube pages isn't something we can ship, since it goes against YouTube's terms, so closing this.

PR #287, opened 2026-09-11
refactor(geo): share great-circle distance utility

### LINKED ISSUE #274: Eight independent haversine/great-circle implementations; one has silently swapped lat/lon argument order
## Summary
At least eight independent great-circle/haversine-distance implementations exist across `src/`, with inconsistent units and internal constants, and one of them — `src/data/naturalEarthRegions.js`'s `haversineKm` — silently takes its arguments in `(lon, lat, lon, lat)` order while every sibling implementation takes `(lat, lon, lat, lon)`, a latent silent-wrong-distance bug if that function is ever called with the far more common argument order.

## Evidence
Eight implementations of the same core formula, confirmed by direct inspection:
```
src/cockpitCloudEffects.js:124   function greatCircleM(a, b)                         // object args, meters, atan2
src/locations.js:642             function greatCircleKm(lat1, lng1, lat2, lng2)      // km, asin(min(1,sqrt(a)))
src/data/naturalEarthRegions.js:37  function haversineKm(lon1, lat1, lon2, lat2)     // km, asin — ARGS SWAPPED
src/data/bikeshare.js:631        function haversineKm(aLat, aLon, bLat, bLon)        // km, atan2
src/data/routePlausible.js:17    export function greatCircleKm(lat1, lon1, lat2, lon2) // km, atan2
src/data/trafficBounds.js:31     export function greatCircleKm(lat1, lon1, lat2, lon2) // km, asin
src/data/analystEngine.js:41     export function haversineKm(lat1, lon1, lat2, lon2)   // km, asin
src/data/traffic.js:1082         function distanceKm(a, b)                            // object args, flat-earth approximation (not haversine)
```
The load-bearing claim — the swapped argument order — is confirmed directly from the source:
```js
// src/data/naturalEarthRegions.js:37
function haversineKm(lon1, lat1, lon2, lat2) {
```
versus every other same-shaped sibling, e.g.:
```js
// src/data/analystEngine.js:41
export function haversineKm(lat1, lon1, lat2, lon2) {
```
Both functions share the same name (`haversineKm`) and the same four-positional-number calling convention, but `naturalEarthRegions.js`'s version expects longitude first. The other seven implementations also disagree on the sphere radius constant used (6371 vs 6371000 vs a named `EARTH_RADIUS_KM`/`EARTH_R_KM` constant) and on the final trig step (`atan2(sqrt(a), sqrt(1-a))` vs `asin(min(1, sqrt(a)))` — mathematically equivalent, but a further inconsistency), and `traffic.js:1082` is an eighth, algorithmically distinct flat-earth (equirectangular) approximation rather than a true haversine.

## Why this matters
`naturalEarthRegions.js`'s `haversineKm(lon1, lat1, lon2, lat2)` is a landmine: any future contributor who sees the name `haversineKm` and calls it with the near-universal `(lat, lon, lat, lon)` convention used by every other implementation in this codebase gets a silently wrong distance rather than an error, since both arguments are ordinary finite numbers in overlapping ranges near the equator/prime-meridian region and the function has no way to detect the mistake. Beyond that specific landmine, eight independent reimplementations of the same formula is duplicated logic that must be kept correct independently, with no shared source of truth.

## Attack or failure scenario
Not a security issue — a correctness/maintainability risk. Realistic failure: a contributor extends `naturalEarthRegions.js` or copies its `haversineKm` call pattern into a new module expecting the codebase-standard `(lat, lon)` order, producing systematically wrong distances that are easy to miss in testing if inputs happen to be far from the equator/meridian (where the error is more likely to be large and obviously wrong) versus near it (where it may look plausible).

## Root cause
Each data layer module was built independently (consistent with CONTRIBUTING.md's "each layer is one self-contained module" architecture) and each needed distance calculations, so each one reimplemented haversine locally rather than importing a shared utility — and `naturalEarthRegions.js`'s author chose `(lon, lat)` order, inconsistent with the rest of the codebase, with nothing to catch the inconsistency.

## Recommended fix
Extract one shared `haversineKm(lat1, lon1, lat2, lon2)` (or `greatCircleKm`) utility with a single, clearly documented argument order and a single named earth-radius constant, and migrate all eight call sites to it. At minimum, rename `naturalEarthRegions.js`'s function (e.g. to `haversineKmLonLatOrder`) or reorder its parameters to match every sibling, to remove the landmine even if full deduplication is deferred.

## Acceptance criteria
- No two implementations of the same distance formula disagree on argument order under the same function name.
- Ideally, a single shared utility replaces the eight independent implementations.

## Suggested labels
architecture, bug

## Priority
P2

## Severity
Medium — a real, confirmed latent-bug risk (argument-order landmine) plus a clear duplication/maintainability finding, though no current call site was found to actually misuse the swapped-order function.

## Confidence
Confirmed — read all eight implementations directly at the cited line numbers and confirmed the argument-order discrepancy is real, not a misreading.

onionviolet:
## Summary
- add one tested great-circle utility with a single `(lat, lon, lat, lon)` contract
- migrate spherical distance checks across regions, routes, traffic bounds, analyst, CCTV, bikeshare, HUD, voice, annotations, locations, and cockpit weather
- fix Natural Earth region bounds to call the shared latitude-first contract
- retain and document traffic center-shift math as an intentional local approximation

Closes #274.

## Verification
- focused geodesy and consumer tests (55/55)
- `npm test` (2,710 pass, 0 fail, 1 platform skip on Node 26.4.0)
- `npm run build`
- `npm run test:track -- --url http://127.0.0.1:4194` (108/108)


onionviolet:
Rebased onto current `main` after the adjacent launcher work landed. The geodesy finding is still present upstream, so this PR remains applicable.\n\nFresh verification on `43baa6d`:\n- focused geodesy/consumer tests: 232/232\n- production build: passed\n- full suite: 2,947 passed, 1 platform skip, 1 unrelated macOS path-alias failure in `src/devCctv.test.mjs` (`/private/var/...` vs `/var/...`)\n\nAwaiting the refreshed GitHub CI checks.

src/geoDistance.js
INDRANIL1707, line 1: I think this PR would benefit from making the contract of `geoDistance.js` explicit before it becomes the shared implementation for all of these consumers.

The current API is:

```js
greatCircleKm(lat1, lon1, lat2, lon2)
```

The positional `(lat, lon)` convention is documented, but it is still structurally possible to pass `(lon, lat)` and receive a finite, plausible-looking distance, which is particularly relevant because #274 was opened around exactly that class of silent coordinate-order error.

Could we settle and document the following as part of this PR/#274 coordination?

* the canonical coordinate representation and whether a point-object form should be preferred for shared/public spatial APIs;
* canonical units and the relationship between `greatCircleKm` / `greatCircleMeters`;
* explicit behavior for `NaN`, `Infinity`, missing coordinates, and out-of-range latitude/longitude;
* whether `greatCircleKm` is the sole canonical name, with the existing `haversineKm` exports only retained temporarily for compatibility where necessary.

I think getting this contract right now will prevent the repository from replacing duplicated implementations with a duplicated ambiguity.

src/geoDistance.test.mjs
INDRANIL1707, line 1: Since this PR makes `geoDistance.js` the shared implementation for a large number of existing consumers, I think the dedicated test should establish the core invariants more comprehensively.

The current tests cover the basic calculation, but I would add explicit cases for:

```text
identity
symmetry: d(a,b) === d(b,a)
one-degree calibration
dateline crossing
antipodal points
non-finite / invalid coordinates
explicit latitude-before-longitude regression
```

I’d also add a small number of consumer-level regression checks for code paths where the distance is used as a threshold or selection criterion.

In particular, the Natural Earth migration is worth locking down with a regression case because the previous implementation used `(lon, lat, lon, lat)` while the new shared contract is `(lat, lon, lat, lon)`.

That would give us confidence in both halves of the refactor:

```text
canonical distance function is correct
+
existing consumer semantics remain correct
```

rather than only demonstrating that the new helper produces plausible distances.

src/geoDistance.js
onionviolet: Addressed in 9c638fa. The canonical shared API is now object-based: `greatCircleKm({ lat, lon }, { lat, lon })`, with `greatCircleMeters` as the exact metre conversion. The helper validates finite WGS84-style ranges and returns `Infinity` for missing, non-finite, or out-of-range points so invalid data is excluded from threshold and nearest-item selections. It uses the IUGG mean radius (6371.0088 km), preserving the prior voice-path value. Existing exported `haversineKm` entry points remain only as compatibility wrappers that delegate to the canonical helper.

I rebuilt the change on current `main` because several consumers moved during the intervening refactors, and migrated the new module locations as part of that refresh.

src/geoDistance.test.mjs
onionviolet: Addressed in 9c638fa. The dedicated tests now cover identity, symmetry, one-degree calibration, dateline crossing, antipodal points, metre conversion, non-finite/missing/out-of-range inputs, explicit point-key shape, and a latitude/longitude swap regression.

I also added consumer regressions for the Natural Earth `[lon, lat]` bbox conversion, route plausibility invalid-position fallback, and traffic fetch-center threshold/fallback behavior. Verification on current `main`: 4,142 tests passed with one expected platform skip; 273 focused consumer tests passed; package/import boundaries passed; formatting passed; production build passed.

PR #294, opened 2026-09-11
feat: add Cesium-free spatial query primitives

### LINKED ISSUE #188: Proposal: add a Cesium-free spatial query utility for shared distance/selection logic
I’d like to contribute a small, reusable spatial primitive rather than another data layer.

### Proposed scope
Add a pure, Cesium-free utility for point-based spatial queries:

- great-circle distance in metres
- radius queries returning matches ordered by distance
- nearest-entity lookup
- deterministic tie handling
- finite-coordinate validation

The API would operate on caller-supplied position accessors so it does not couple the utility to aircraft, vessels, CCTV, or a specific entity shape.

### Why this fits the current codebase
I noticed that great-circle calculations currently exist in multiple spatially relevant modules (for example `trafficBounds.js` and `routePlausible.js`). A small shared primitive could reduce duplication without forcing a larger architectural refactor.

I would keep this PR deliberately narrow:

- no UI changes
- no new data source
- no Cesium dependency in the utility
- node:test coverage for normal cases, dateline crossing, exact radius boundaries, invalid coordinates, and deterministic ordering

A later PR could consume the utility from an existing feature, but I would keep the first change independently reviewable.

Would this direction be useful to the maintainers? I can prepare the focused implementation and tests against `main`.

INDRANIL1707:
## Summary

Adds a small, dependency-free spatial query primitive for point-based geographic data.

This implementation is intentionally independent of Cesium and of any specific entity shape. Callers provide a position accessor, allowing the same primitive to be reused across different spatial datasets and features.

### Included

* great-circle distance in metres using the haversine formulation
* radius queries ordered from nearest to farthest
* nearest-entity lookup
* deterministic handling of equal-distance results
* finite coordinate and argument validation
* support for both `lon` and `lng`
* invalid entity positions skipped during collection queries
* caller-supplied position accessors
* no mutation of the source entity array

### Tests

Coverage includes:

* identical points
* one-degree latitude calibration
* dateline crossing
* antipodal points
* `lon` / `lng` compatibility
* inclusive radius boundaries
* zero-radius queries
* result ordering
* deterministic distance ties
* invalid coordinates
* index-aware accessors
* empty and no-valid-result inputs
* invalid query arguments
* source-array non-mutation

Local verification:

```text
node --test SpatialQuery.test.js

18 tests
18 passing
0 failures
```

### Scope

This PR deliberately does not refactor existing spatial consumers such as `trafficBounds.js` or `routePlausible.js`.

The goal here is to establish the reusable primitive and its contract first, keeping the change independently reviewable. Existing consumers can adopt it in a follow-up change if maintainers consider that desirable.

Closes #188


bilawalsidhu:
The description covers a spatial-query helper, but the diff also rewrites index.html, removing 429 lines including Provider Settings, the HUD, and the location bar, and adds an undescribed viewport controller. This is a stale fork, not a scoped change. Closing.

INDRANIL1707:
> The description covers a spatial-query helper, but the diff also rewrites index.html, removing 429 lines including Provider Settings, the HUD, and the location bar, and adds an undescribed viewport controller. This is a stale fork, not a scoped change. Closing.

Thanks for catching that. The index.html changes and viewport controller were local experimental features on my fork that accidentally leaked into this branch when rebasing/merging.

The core spatial query primitives in SpatialQuery.js and its tests are completely isolated and independent. I'll clean up the branch to strip out index.html and the extra viewport changes, leaving strictly the spatial query primitives, and reopen/resubmit a clean PR for review.

PR #290, opened 2026-09-11
fix: update vulnerable dependencies and Puppeteer QA compatibility

samehkhamis:
Refresh the locked dependency tree to address the known npm audit findings. `npm audit` reports zero vulnerabilities after a clean `npm ci` (previously 11 affected packages: nine high, two moderate).

- Update Puppeteer 24.37.5 → 25.10.0 and Sharp 0.34.5 → 0.35.4.
- Refresh DOMPurify 3.4.10 → 3.4.15, protobufjs 8.6.4 → 8.8.0, PostCSS 8.5.15 → 8.5.28, and nanoid 3.3.12 → 3.3.19 through compatible lockfile updates. Cesium stays at 1.138.0; Vite stays at 6.4.3.
- Puppeteer's new dependency tree removes the affected extract-zip, ip-address and js-yaml paths. Its executable-path API is now asynchronous, so 29 QA scripts receive a one-line await adjustment, preserving their existing browser overrides and fallback paths.

Validation on Node 24.14.0:
- Clean install, setup doctor, 2,770 unit/allocation checks passed (one platform skip), production build, and zero npm audit findings.
- All 29 changed QA scripts pass syntax checking.
- Sharp PNG decode → resize → WebP encode/decode smoke test passed.
- `npm run test:track -- --url http://localhost:4174`: 108 passed; keyless Map Source QA: 23 passed, using Puppeteer 25 / Chrome 152.
- Patch applies cleanly after the separate security PR #289. No formatting or feature refactor is bundled here.

Revalidated after merging the accepted #289 base into this branch: setup doctor, all unit/allocation tests, build, zero audit findings, 108 tracking checks, and all required CI passed.

PR #296, opened 2026-09-11
fix(data): name the missing provider key on a key-gated layer's control (#143)

### LINKED ISSUE #143: UI tooltips: explain each visual preset, and name the missing key on disabled controls

Matthew-Selvam:
Closes #143 (part 2 of 2).

## What

Keyless, the Wildfires row in DATA LAYERS reads `KEY REQUIRED` but never says **which** key — the one data-layer control a provider key gates left the operator with a dead-looking row and no next step. The preset tooltips (part 1) and the map-source guidance already landed; this is the remaining half of the maintainer's issue.

The provider registry (`src/keySetupCore.mjs`) already knows every key's env vars, so the wiring is two small halves:

- A layer may declare the key it needs as a **registry id** (`requiresKeyId: 'firms'` on `createFirmsHeatmapLayer`, mirroring `showInTogglePanel`), which `DataLayerManager.getAll()` publishes on the row.
- It reports the keyless state machine-readably (`stats.keyRequired: _keyRequired` in `getStats()`).

`layerKeyRequirementTooltip()` (exported from `src/data/manager.js`) turns the pair into text, and `_syncToggleButton` writes it onto the control. Keyless, the row's toggle now tooltips:

> Needs FIRMS_MAP_KEY — add it in Provider Settings

Design notes:

- A layer needing no key — or already holding one — carries no title at all.
- An unnamed or **unknown registry id yields NO tooltip** on purpose: guidance naming the wrong env var sends the operator to the wrong provider.
- `stats.keyRequired` is exactly the field `src/loadingFeedback.js` already reads (`keyRequired === true || missingKey === true`), so the keyless state becomes a declared missing-optional-key rather than a generic feed error there too.
- No new failure states are introduced: the fires layer's `error: 'KEY REQUIRED'` string is left byte-for-byte as it was, so the existing `qa-firms` (line 328) and `qa-firstrun` (lines 643–649) keyless assertions keep passing unchanged.
- The generic contract is reusable: any future key-gated layer declares an id; no per-layer tooltip code.

## Verification

- `npm test` on the rebased HEAD (`c4cc996`): 2762 tests, +5 here, 0 failures attributable to this change.
  - One failure exists on plain `origin/main` too and is unrelated: `src/devCctv.test.mjs` fails on macOS because `mkdtemp` returns `/var/…` while the test compares against the `/private/var/…` realpath. Reproduced identically on unmodified `origin/main` (`c82fdd7`); CI runs ubuntu-latest where it passes.
- `npm run build`: green.
- Regression tests assert the **wrong thing stops happening**: the tooltip text, its absence when the key is configured (`button.title === ''`), the empty-string behavior for unknown/undeclared ids, and the end-to-end row shape after a real `setEnabled`.
- Mutation check — all five guarded bodies caught: tooltip forced empty; `requiresKeyId` dropped from the layer; `keyRequired` dropped from `getStats()`; the field dropped from `getAll()`; the title no longer written to the button.
- Docs updated in the same PR per CONTRIBUTING: `CHANGELOG.md` (Unreleased → Fixed) and `docs/CURRENT-STATE.md` (new bullet in the layers section). No new console errors; the change is one `title` assignment plus additive stats fields.
- Conflict note: rebased on current `main` after #290 merged; none of the 6 files this PR touches were modified by the 24 commits since my branch point.

Labels: `enhancement` (matches #143's label; I cannot apply labels from a fork — maintainer to add).


bilawalsidhu:
Landed via #575 (rebased onto the current component layout, with fixes from review). Thank you @Matthew-Selvam!

PR #298, opened 2026-09-11
fix(server): bound terrain-height caches, validate coordinates, throttle requests

### LINKED ISSUE #260: /api/terrain/heights: unbounded in-memory and on-disk cache growth, no rate limiter, no coordinate bounds validation
## Summary
The `/api/terrain/heights` proxy's in-memory cache has no eviction policy at all (unlike every sibling cache in the same file), the resulting cache is dumped verbatim to an uncapped on-disk JSON file, and the endpoint validates only that coordinates are finite — not that they are valid latitude/longitude ranges — with no rate limiter.

## Evidence
`vite.config.js:2242-2250`:
```js
function terrainHeightsProxy() {
  const TTL_MS = 30 * 24 * 3600_000;
  const CACHE_DIR = path.join(process.cwd(), '.gev-cache');
  const CACHE_PATH = path.join(CACHE_DIR, 'terrain-heights.json');
  const UPSTREAM_CHUNK = 256;
  const MAX_POINTS = 2000;

  /** @type {Map<string, {at:number, result:object}>} keyed by canonical 5dp lon/lat. */
  const mem = new Map();
```
There is no size cap or eviction logic anywhere in this function for `mem` — contrast with `_overpassCache`, `_militaryInstallationCache`, `_regionalBriefCache`, `_weatherEffectsCache`, and even the sibling `trackBackfillProxies` cache in the same file, all of which cap entry count. Entries accumulate at `src/data/terrainHeightsProxy.js:188` (`cache.set(missing[i].key, { at: fetchedAt, result });`), keyed by 5-decimal-place lon/lat, and the entire map is periodically flushed verbatim to disk with no size check (`vite.config.js:2294-2296`):
```js
const obj = { version: 2, points: Object.fromEntries(mem.entries()) };
await fsp.writeFile(CACHE_PATH, JSON.stringify(obj), 'utf8');
```
Coordinate validation is `Number.isFinite` only (`src/data/terrainHeightsProxy.js:24-39`, `parseTerrainPoints`) — no `-90..90` / `-180..180` bounds check. Up to 2000 client-chosen points are accepted per request (`MAX_POINTS`), and no rate limiter appears anywhere in this route's handler.

## Why this matters
A client can request large batches of distinct, unbounded (including out-of-range) coordinates repeatedly. Each unique 5-decimal-place point becomes a permanent, never-evicted in-memory entry and a permanent line in an on-disk JSON file that itself has no size cap — unbounded growth in both server memory and disk, growable at up to 2000 new points per request with no rate limit slowing the attempt.

## Attack or failure scenario
A script requests `/api/terrain/heights?points=lon,lat;...` with 2000 essentially-random coordinate pairs per call, looped continuously. Server memory (`mem`) and `.gev-cache/terrain-heights.json` both grow without bound until the process is OOM-killed or the disk fills, and each novel batch also forces a real upstream fetch to `terrain.reearth.land`.

## Root cause
This cache was implemented independently of the file's other bounded caches and never received the same eviction treatment; coordinate validation only checks numeric type, not geographic validity.

## Recommended fix
Add an LRU-style cap to `mem` (matching the pattern already used by `trackBackfillProxies`'s `TRACK_CACHE_MAX`), cap the on-disk cache file size the way `rocketLaunchesProxy`'s `maxDiskCacheBytes` does, validate `lat`/`lon` are within `[-90,90]`/`[-180,180]` in `parseTerrainPoints`, and add a rate limiter to the route.

## Acceptance criteria
- `mem` has a maximum entry count with eviction of the oldest/least-recently-used entries.
- The on-disk cache file has an enforced maximum size.
- Out-of-range latitude/longitude values are rejected with a 400 before being cached or forwarded upstream.
- The route is rate-limited.

## Suggested labels
bug, performance

## Priority
P2

## Severity
Medium — unauthenticated, unbounded resource growth (memory + disk), but requires sustained abuse rather than a single request to have real impact.

## Confidence
Confirmed — read `terrainHeightsProxy()` in full and `src/data/terrainHeightsProxy.js` in full; verified no eviction/size-cap code exists for `mem` or `CACHE_PATH`, and that `parseTerrainPoints` only checks `Number.isFinite`.

armanbabazadeh6:
Closes #260.

## Problem

`/api/terrain/heights` accepted finite-but-impossible coordinates, had no eviction on its in-memory point cache, dumped that cache to disk with no byte cap, and had no rate limiter. A client cycling the coordinate keyspace could grow memory and disk without bound while forcing unbounded upstream work.

## Changes

- `parseTerrainPoints` rejects `lon` outside `[-180,180]` or `lat` outside `[-90,90]` with a 400 before any cache write or upstream call.
- `resolveTerrainHeightRequest` evicts oldest-by-fetch-time entries above `TERRAIN_MAX_CACHE_ENTRIES` (50k); a point refreshed by the current request is never evicted.
- `serializeTerrainCache` writes `.gev-cache/terrain-heights.json` within a 16 MB ceiling (oldest dropped first); an oversized legacy file is replaced instead of parsed at full size.
- The route is throttled per client (90/min, 300/min global) and returns 429 with `Retry-After`.
- `CHANGELOG.md` and `docs/CURRENT-STATE.md` updated.

## Verification

- `npm test`: 2820 passed, 0 failed (Node 22; the 2 Node-24-only allocation microbenchmarks skip as expected).
- `npm run build`: clean, only the pre-existing chunk-size warnings.
- `npm run test:track` against the dev server: 108 passed, 0 failed.
- Live endpoint: valid point -> 200; `points=181,0` and `points=0,91` -> 400; 90 requests allowed, then 429 with `Retry-After: 5`.
- New regression tests cover coordinate bounds, cache eviction (including the request-refreshed point), bounded disk serialization, and the middleware 429 path.


GGPOShadows:
I independently hit the same unbounded-cache problem (from the #327 angle) and wrote a fix before finding this PR. Yours is broader — coordinate validation, the disk byte cap and the rate limiter are all things I didn't do — so rather than open a competing PR I'd rather hand you what I found. One of these is a real edge case in the current implementation; the other is a suggestion.

## Eviction runs before the response is rebuilt, and can drop a hit the request still needs

`evictTerrainCache(cache, maxCacheEntries)` is called inside the `missing.length > 0` block, after the insert loop but **before** `const results = requested.map(...)` reads the cache back. Eviction sorts by `at`, which is only ever set at fetch time, so a long-cached point sorts first even when the current request is reading it.

The docstring says "a point refreshed during the current request is never the one dropped", which is true — refreshed points carry `at = fetchedAt` and sort last. But a **cache hit** carries its original old `at` and is fully eligible.

Reproduced against your `evictTerrainCache` verbatim, cap of 3:

```
cache: hot(at=1), b(at=50), c(at=60)
request: ['hot' (hit), 'new' (miss)]

insert 'new' -> evict -> rebuild

cache after evict : b, c, new
hot point survived: false
rebuilt results   : [null, {"ellipsoid":9}]
```

The null makes `resolveTerrainHeightRequest` return 502 for a request every point of which was satisfiable. Reachable once the cache is at cap: `hot` here stands in for something like a fixed CCTV or installation position resolved once at startup and re-read constantly afterwards — the oldest `at` in the map, and so the first victim, despite being the hottest entry.

Two ways to close it, either is small:
- move the `evictTerrainCache` call after the `results` rebuild, or
- skip keys in `unique` when choosing victims.

## Sorting by fetch time evicts hot entries first

Same root cause, separate from the 502. Because `at` is never updated on read, the eviction order is "least recently *fetched*", not "least recently *used*". The entry read every frame is exactly the one with the oldest `at`, so it gets evicted while a cold point fetched a minute ago and never touched again survives.

Touching an entry on read makes `Map` insertion order a real LRU and lets eviction be `cache.keys().next()` — which also drops the `[...cache.entries()].sort()` (O(n log n) plus a 50k-element array copy on each fetching request) down to O(1):

```js
function touch(cache, key) {
  const entry = cache.get(key);
  if (entry === undefined) return entry;
  cache.delete(key);
  cache.set(key, entry);
  return entry;
}
```

Called where the freshness loop already does `cache.get(key)`. Worth touching on *any* existing entry rather than only fresh ones, since the final rebuild serves stale-but-valid entries too.

This is the same `Map`-insertion-order trick already used by `memSet` in `server/providers/traffic.js` and `trimOverpassCache` in `server/providers/overpass/cache.js`, so it matches existing house style.

## Offer

I have both of these implemented with four regression tests, including one that fails with a 502 against a working-set-unaware evictor. Happy to push them to your branch, open a PR against it, or just paste the diff here — whichever you prefer. Equally happy to leave it entirely; you got here first and it's your patch.

One unrelated note in case it's useful: #327's title points at `src/services/terrainHeights.js`, the client-side cache, which looks like #594's territory rather than this PR's server-side one. Might be worth a cross-reference so the two don't get deduped by mistake.

bilawalsidhu:
Thanks, and thanks for filing this first. The terrain-heights bounds are being carried forward in #686, which targets the current server layout (`server/providers/`), so closing this pre-refactor version. You'll be credited when it lands.

PR #301, opened 2026-09-11
fix(test): resolve tmpdir symlinks so the CCTV launcher test passes on macOS

### LINKED ISSUE #300: npm test fails on macOS: devCctv launcher test compares logical /var tmpdir to physical /private/var cwd
## Summary

On a fresh macOS checkout, `npm test` reports one failure:

```
✖ CCTV preset starts keyless on localhost through the normal launcher
  + actual - expected
  + '/private/var/folders/.../T/gev-cctv-launch-XXXX'
  - '/var/folders/.../T/gev-cctv-launch-XXXX'
      at TestContext.<anonymous> (file:///.../src/devCctv.test.mjs:50:10)
```

## Cause

`os.tmpdir()` on macOS returns `/var/folders/...`, which is a symlink to `/private/var/folders/...`. The fixture in `src/devCctv.test.mjs` builds the fake checkout under that logical path. `scripts/dev-cctv.sh` computes `ROOT_DIR` with `cd ... && pwd`, which keeps the logical path, but the fake `npm` shim records `process.cwd()`, which Node reports as the physical path. The assertion `assert.equal(result.cwd, result.root)` then compares `/private/var/...` to `/var/...`.

The launcher itself behaves correctly; only the comparison is wrong.

## Why CI does not catch it

The CI matrix runs on `ubuntu-latest` and `windows-latest`. Linux tmpdirs are not symlinked and the bash tests are skipped on Windows, so the failure only shows up for contributors on a Mac, which is the platform CONTRIBUTING.md describes for Keychain-based key setup.

## Environment

- macOS (Darwin 25.2.0), Node 26.8.2, npm 11.19.1
- `main` at c7ef018

## Proposed fix

Resolve the fixture root with `fs.realpath` right after `fs.mkdtemp` so both sides compare the same physical path. I have a PR ready for this.

VassagoDevteam:
Fixes #300.

## What changed

`src/devCctv.test.mjs` now resolves the fixture root with `fs.realpath` immediately after `fs.mkdtemp`. One line plus a comment; no runtime code is touched.

## Why

On macOS `os.tmpdir()` is `/var/folders/...`, a symlink to `/private/var/folders/...`. The `dev-cctv.sh` launcher keeps the logical path via `cd ... && pwd`, while the fake `npm` shim records `process.cwd()`, which Node returns as the physical path. `assert.equal(result.cwd, result.root)` therefore fails on every Mac even though the launcher behaves correctly. CI runs Ubuntu and Windows only, so it never surfaced there.

## Verification

On macOS (Darwin 25.2.0, Node 26.8.2):

- `node --test src/devCctv.test.mjs`: 3 pass, 0 fail (was 2 pass, 1 fail)
- `npm test`: 2820 pass, 0 fail, 1 skipped (was 2819 pass, 1 fail)
- `npm run build`: green
- `npm run test:track` against the dev server: 108 passed, 0 failed

No CHANGELOG or `docs/CURRENT-STATE.md` update since runtime behavior is unchanged.


bilawalsidhu:
Landed via #501 (rebased onto the current component layout, with fixes from review). Thank you @VassagoDevteam!

PR #306, opened 2026-09-11
fix(server): rate-limit the aircraft track-backfill proxies (#259)

### LINKED ISSUE #259: OpenSky track-backfill proxies (/api/opensky-track, /api/adsblol/trace) bypass the shared credit/cooldown governor
## Summary
`/api/opensky-track` and `/api/adsblol/trace` (implemented by `trackBackfillProxies()`) mint OpenSky OAuth requests and adsb.lol requests without ever consulting the shared OpenSky credit/cooldown governor that every other OpenSky-adjacent route respects, and without any rate limiter of their own.

## Evidence
`vite.config.js:4956-5043`:
```js
function trackBackfillProxies() {
  const TRACK_CACHE_MS = 60000;
  const TRACK_CACHE_MAX = 200;
  ...
  middlewares.use('/api/opensky-track', async (req, res) => {
    ...
    const icao24 = String(incoming.searchParams.get('icao24') || '').trim().toLowerCase();
    if (!/^[0-9a-f]{6}$/.test(icao24)) { ...400... }
    const token = await getOpenSkyToken();
    await proxyJson(res, `osky:${icao24}`,
      `https://opensky-network.org/api/tracks/all?icao24=${icao24}&time=0`,
      token ? { Authorization: `Bearer ${token}` } : {});
  });
```
The route's own doc comment (`vite.config.js:4944-4955`) states this OpenSky endpoint has its "own credit bucket, 4 credits per call". The main `/api/opensky` route enforces an adaptive-TTL/429-cooldown governor (`_openskyCooldownUntil`/`_openskyTtlMs`, `vite.config.js:3013-3295`); `trackBackfillProxies` never references that shared state. The only throttle here is a 60-second per-`icao24` result cache (`TRACK_CACHE_MS`) with a 200-entry LRU-style cap (`TRACK_CACHE_MAX`) — that bounds memory, not request volume, and does nothing to stop a caller from cycling through *different* `icao24` values.

## Why this matters
`icao24` is a 6-hex-digit value (16,777,216 possible values), each a fresh cache key. A caller that enumerates random hex values forces one live, credit-consuming OpenSky request per unique value, 4 credits each, with no per-IP or global cap — including while the main `/api/opensky` route is already in a 429 cooldown because the same OpenSky account is out of budget. OpenSky's free-tier daily credit budget can be exhausted in well under a thousand requests this way, taking down aircraft tracking for every legitimate user of the same server for the rest of the day.

## Attack or failure scenario
Anyone who can reach the dev/preview server (LAN exposure, a forgotten `HOST=0.0.0.0` launch, or a misconfigured reverse proxy) sends a burst of `GET /api/opensky-track?icao24=<random-hex>` requests with distinct `icao24` values. Each miss costs 4 OpenSky credits with no throttle; the account's daily budget is exhausted well before any human operator notices, and the cooldown governor that protects the main flight-tracking feature never engages because this code path doesn't participate in it.

## Root cause
Cost-bearing proxy routes were added in a separate module (`trackBackfillProxies`) from the main OpenSky proxy and its governor, and the new module reimplemented a local cache without reusing (or even referencing) the existing shared rate/cooldown state.

## Recommended fix
Route both `/api/opensky-track`'s OpenSky call and, ideally, `/api/adsblol/trace` through the same shared rate limiter / cooldown governor used by `openSkyProxy()`, or apply a dedicated per-IP request-rate limiter (e.g. the same `makeOptInRateLimiter`-style primitive used elsewhere in this file) with a mandatory, non-opt-in default, since this consumes the same finite OpenSky credit budget as the main proxy.

## Acceptance criteria
- A burst of distinct-`icao24` requests to `/api/opensky-track` is rate-limited independent of the per-icao 60s cache.
- The route either shares the main OpenSky cooldown state or has its own default-on limiter that cannot be bypassed by varying `icao24`.
- Existing single-aircraft track lookups (the intended use case) continue to work unthrottled under normal usage.

## Suggested labels
security, bug (repo has no "security"/"reliability" label at present — see maintainer's `security-audit` label used on related findings; recommend using that if applicable)

## Priority
P1

## Severity
High — unauthenticated, low-cost-per-request path to exhausting a shared, metered third-party credential with no server-side control.

## Confidence
Confirmed — read the full `trackBackfillProxies()` function and cross-checked that it does not reference `_openskyCooldownUntil`/`_openskyTtlMs` or any `RateLimiter` primitive used elsewhere in the file.

SurefireStudios:
Fixes #259.

`/api/opensky-track` and `/api/adsblol/trace` are keyed by a client-supplied aircraft id.

The 60-second per-aircraft cache in `trackBackfillProxies()` bounds memory with a 200-entry LRU but not request volume, so a caller cycling ids misses the cache every time — `icao24` is 6 hex digits, giving ~16.7M distinct keys, and neither route has a limiter.

Each `/tracks` miss spends 4 credits from OpenSky's finite daily budget, and that budget belongs to the server, not the caller.

Unlike `/api/opensky`, this path has no governor at all, so it keeps spending even after the account is already being throttled.

## The change

Each route gets its own default-on limiter, declared next to the existing `_overpassRateLimiter` and friends: `/api/opensky-track` at 30 per client/min and 120 global, `/api/adsblol/trace` at 60 and 240.

The limiter is consulted only on a cache miss, since `proxyJson()` serves a fresh cached track before touching it — so re-selecting the same aircraft never draws on quota, mirroring how `/api/overpass` passes `allowUpstream` into its cache preflight.

Separate buckets per route, because they are different upstreams with different costs, so exhausting adsb.lol does not close the OpenSky route — the same reasoning `/api/military-installations` already uses.

These are default-on rather than opt-in like the OpenAI and Google limiters, because the cost lands on the shared server account rather than the caller.

Over-cap callers get the same sanitized 429 and `Retry-After: 5` the other limited routes return, and the client already treats any non-OK backfill response as "keep the locally accumulated trail", so there is no new UI failure mode.

I deliberately did not couple this to the main `/api/opensky` cooldown state, which the issue offers as an alternative.


bilawalsidhu:
Thanks, and thanks for filing this first. The track-backfill limit is being carried forward in #685, which targets the current server layout, so closing this pre-refactor version. You'll be credited when it lands.

PR #308, opened 2026-09-11
fix(server): bound the adsbdb cache and limit its upstream calls (#261)

### LINKED ISSUE #261: /api/adsbdb: no rate limiter, unbounded persisted cache via keyspace enumeration
## Summary
The `/api/adsbdb` enrichment proxy has no rate limiter and no cache eviction, so a client cycling through the callsign/hex keyspace can force unlimited live upstream requests to api.adsbdb.com and grow the on-disk cache without bound.

## Evidence
`vite.config.js:2439-2470` (`lookup`) and `:2474-2499` (route handler):
```js
server.middlewares.use('/api/adsbdb', async (req, res) => {
  await loadOnce();
  ...
  if (kind === 'route') {
    const cs = String(rawKey || '').toUpperCase();
    if (!/^[A-Z0-9]{2,8}$/.test(cs)) return send(400, { error: 'invalid callsign' });
    const data = await lookup('route', cs);
    ...
  }
  if (kind === 'type') {
    const hex = String(rawKey || '').toLowerCase();
    if (!/^[0-9a-f]{6}$/.test(hex)) return send(400, { error: 'invalid hex' });
    const data = await lookup('aircraft', hex);
    ...
  }
```
`lookup()` caches both hits and 404s (negative caching) keyed by the raw callsign/hex string, with no eviction visible in `cache.routes`/`cache.aircraft`. No `RateLimiter` call appears anywhere in this proxy. The callsign pattern `^[A-Z0-9]{2,8}$` alone spans a keyspace from 36^2 to 36^8 — trivially enumerable for short values, and the endpoint helpfully caches every attempt forever (persisted to `.gev-cache/adsbdb.json`, per the module's own doc comment at `vite.config.js:2383-2389`).

## Why this matters
Even though each entry is cheap and the API is described as a "free community API," an unthrottled proxy that persists every distinct query forever gives an attacker a way to (a) grow the on-disk cache unboundedly and (b) hammer a third-party community API through this server with no limit, which risks that community API rate-limiting or blocking the operator's IP.

## Attack or failure scenario
A script requests `/api/adsbdb/route/<random-2-8-char-string>` repeatedly with distinct values. Each miss triggers a live fetch to api.adsbdb.com and a permanent new cache entry; there is nothing in this code path to slow it down.

## Root cause
No rate limiter was applied to this route, unlike several other proxies in the same file, and the cache has no eviction policy.

## Recommended fix
Add a rate limiter to `/api/adsbdb` and an eviction/size cap to `cache.routes`/`cache.aircraft` (and its disk-persisted form), consistent with the caching patterns used elsewhere in `vite.config.js`.

## Acceptance criteria
- The route is rate-limited per client/IP.
- The cache has a maximum size with eviction.

## Suggested labels
bug, performance

## Priority
P2

## Severity
Medium — unauthenticated request amplification against a third-party API plus unbounded cache growth; no direct cost to the operator's own paid keys, but real availability/abuse risk.

## Confidence
Confirmed — read the full `adsbdbProxy()` handler and `lookup()` function; no rate limiter or cache eviction present.

SurefireStudios:
Fixes #261.

`/api/adsbdb` caches every distinct callsign and hex it is asked about, negative results included, and persists that to `.gev-cache/adsbdb.json`.

Both stores are plain objects keyed by a client-supplied string, with no eviction and no rate limit.

The callsign pattern `^[A-Z0-9]{2,8}$` alone spans 36² to 36⁸ keys, so a caller cycling values grows the file without bound and relays unlimited traffic to a free community API.

## Changes

Cache ceiling of 10,000 entries per store, evicting oldest-first, pruning to a 9,000 low-water mark so the sort runs about once per thousand inserts rather than on every insert at capacity.

A cache file written before the ceiling existed is trimmed on first load, so existing installs shrink rather than staying oversized.

Upstream limiter at 480 per client/minute, 1920 global.

That cap looks high, so: the client drips all enrichment through one shared queue at one dispatch per `ENRICH_DISPATCH_GAP_MS` (200 ms), documented at `src/data/flights.js:778-780` as a ≤5/s ceiling, and that queue serves only these two endpoints.

So a single legitimate client can issue 300 requests/minute, every one a miss on a cold cache — a cap below that would throttle a normal first run.

The limiter is consulted only when the cache misses and nothing identical is already in flight, so a coalesced caller rides the existing request and isn't charged.

`lookup()` takes an `allowUpstream` callback, mirroring how `/api/overpass` passes one into its cache preflight, and returns a sentinel so "throttled" stays distinguishable from "not found" — a throttled lookup must not be negative-cached.

One caveat: the client's `_enrichSeen` set means a throttled key isn't retried that session, so a few aircraft would go un-enriched above the cap.

## Verification

`npm run build` green; `npm test` gives 2890 tests, 0 failures (2881 on `main`, +9 new), same 5 pre-existing skips.

Live against a dev server and real upstream: valid type and route lookups return data, `route/!` and `type/nope` give 400, unknown gives 404, and repeat lookups served from `.gev-cache/adsbdb.json` without a new upstream call.

I deliberately did not fire 480 live requests at adsbdb to demonstrate the limiter — that is the abuse this PR prevents — so the unit tests cover that path against the real middleware.

`npm run test:track` not run: it stubs `/api/adsbdb/*` in the browser at `scripts/track-regression.mjs:351`, and I have no Google Maps key.

## One note

`src/proxyErrorResponses.test.mjs` needed a fixture change, since it builds these proxies with `new Function` and the new constants are free variables it must provide.

I made it read those declarations out of the source rather than duplicating values, so retuning doesn't need a matching edit — worth flagging because #298 also touches that file.


bilawalsidhu:
Thanks, and thanks for filing this first. The adsbdb bounds are being carried forward in #689, which targets the current server layout, so closing this pre-refactor version. You'll be credited when it lands.

PR #309, opened 2026-09-11
ci: pin GitHub Actions to commit SHAs (#267)

### LINKED ISSUE #267: GitHub Actions pinned to mutable version tags (@v4) instead of commit SHAs
## Summary
GitHub Actions workflow steps reference `actions/checkout` and `actions/setup-node` by mutable major-version tags (`@v4`) rather than pinned commit SHAs.

## Evidence
`.github/workflows/ci.yml:29,34,60,65`:
```yaml
      - name: Check out source
        uses: actions/checkout@v4
        ...
      - name: Use Node.js
        uses: actions/setup-node@v4
```
The workflow currently sets `permissions: contents: read` at the top level (`ci.yml:8-9`), triggers only on `pull_request`/`push: [main]` (no `pull_request_target` anywhere in the repo), and references no third-party (non-`actions/`) action. Exploitability today is therefore low. This is distinct from the generic dependency-CVE scan dumps in issues #229/#228 — it's a GitHub Actions supply-chain hardening gap, not an npm package advisory.

## Why this matters
A tag like `@v4` can be moved to point at a different commit (by the action maintainer, or by an attacker who compromises the action's repository/npm-adjacent publish pipeline), silently changing what code runs in CI. Pinning to a commit SHA makes the exact code that runs auditable and immutable regardless of upstream changes. This is low-urgency today given the workflow's restrictive permissions and triggers, but it's a standard, low-cost hardening step, and it's worth having *before* permissions or triggers ever get loosened (e.g. if a future workflow adds `pull_request_target` for a legitimate reason).

## Attack or failure scenario
If `actions/checkout` or `actions/setup-node`'s tag is ever repointed to malicious code (upstream compromise), every job in this workflow silently executes that code the next time it runs.

## Root cause
Convenience of using floating version tags rather than pinned SHAs, which is a common but avoidable practice.

## Recommended fix
Pin `actions/checkout` and `actions/setup-node` to specific commit SHAs (with a version comment for readability, e.g. `uses: actions/checkout@<sha> # v4.x.x`), and consider adding Dependabot/Renovate configuration to keep pinned SHAs updated automatically (no `.github/dependabot.yml` or `renovate.json` currently exists in the repo).

## Acceptance criteria
- All `uses:` references in `.github/workflows/ci.yml` are pinned to commit SHAs rather than mutable tags.

## Suggested labels
ci-cd (label doesn't exist on this repo at present; closest available is none of the default labels — recommend using `bug` or a new `ci-cd`/`devops` label)

## Priority
P3

## Severity
Low — real hardening gap, but current permissions/triggers limit practical exploitability.

## Confidence
Confirmed — read the full workflow file; verified `permissions: contents: read` scope, absence of `pull_request_target`, and absence of any secrets usage.

SurefireStudios:
Fixes #267.

## The change

`actions/checkout` and `actions/setup-node` were referenced by the mutable `@v4`
tag in all four workflow steps. A tag can be repointed — by the action's
maintainer, or by someone who compromises that repository — which silently
changes what executes in CI. Pinning to a commit makes it immutable and
auditable.

```diff
-        uses: actions/checkout@v4
+        uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4.4.0

-        uses: actions/setup-node@v4
+        uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4.4.0
```

**This changes nothing about what CI runs today.** I resolved each tag through
the GitHub API first: `actions/checkout@v4` and `actions/setup-node@v4` both
currently point at their respective `v4.4.0` commits, and those are the SHAs
pinned here. So this is purely a freeze of the status quo, not an upgrade or a
downgrade.

| action | tag resolves to | pinned SHA | verified commit |
| --- | --- | --- | --- |
| `actions/checkout` | v4.4.0 | `11d5960a3267` | `backport fixes to releases-v4 (#2524)` |
| `actions/setup-node` | v4.4.0 | `49933ea5288c` | `Bump @action/cache from 4.0.2 to 4.0.3 (#1262)` |

The trailing `# v4.4.0` comments keep the SHAs readable, and they're what
Dependabot reads and rewrites when it bumps a pin.

## About the Dependabot file — easy to drop if you don't want it

Pinning has a real cost: a frozen SHA stops following upstream security fixes,
which for patch-level updates is *worse* than a floating tag. The issue suggests
Dependabot to close that gap, so I've included a minimal config — but it's a
separate file and a separate concern, so please just delete it if you'd rather
not have the bot, and the pinning half stands on its own.

I tried to keep it as quiet as possible, given the current PR queue:

- **Actions only.** npm dependencies are untouched and stay under manual review.
- **Grouped** (`patterns: ['*']`), so both actions bump in one PR rather than two.
- **Monthly**, with `open-pull-requests-limit: 2`.

Realistically that's at most one PR a month, and only when an action actually
moves.

## Verification

Both files parse as YAML, and I diffed the parsed workflow against `main` to
confirm nothing but the `uses:` values moved — triggers (`pull_request`,
`push`), `permissions: contents: read`, the `[24.14.0, 26.x]` matrix,
`persist-credentials: false`, and every `run:` step are byte-identical.

Resolved step actions after the change:

```
verify              -> actions/checkout@11d5960a326750d5838078e36cf38b85af677262
verify              -> actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020
windows-onboarding  -> actions/checkout@11d5960a326750d5838078e36cf38b85af677262
windows-onboarding  -> actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020
```

I also confirmed both SHAs resolve to real commits in their upstream
repositories rather than trusting the tag lookup alone.

No source, test, or docs change, so `npm test` / `npm run build` are unaffected —
though CI on this PR exercises the pinned actions directly, which is the real
check.


bilawalsidhu:
Landed via #505 (rebased onto the current component layout, with fixes from review). Thank you @SurefireStudios!

PR #316, opened 2026-09-11
fix(server): rate-limit the GBFS proxy (#264)

### LINKED ISSUE #264: GBFS proxy (/api/gbfs) has no rate limiter
## Summary
The GBFS (bikeshare) proxy at `/api/gbfs` has no rate limiter, so it can be used as an unthrottled, indefinite relay against any allowlisted third-party GBFS host.

## Evidence
`vite.config.js:3348-3454` (`gbfsProxy`) validates the target host against an allowlist, requires HTTPS, restricts the path to `station_information`/`station_status` endpoints, applies a request timeout, and caps response size at 5MB — but calls no rate-limiting primitive anywhere in the function. This is distinct from open issues #30 (redirect revalidation) and #31 (streaming byte enforcement while reading the response), which address different gaps in the same proxy; the gap here is specifically the absence of any cap on request *frequency*.

## Why this matters
Target validation prevents this from being an SSRF vector, but with no throttle, the dev server can be driven to make an unlimited number of requests against a third-party GBFS provider on the caller's behalf, which risks that provider rate-limiting or blocking the operator's IP, and consumes server resources (one outbound fetch + timeout timer per request) with no back-pressure.

## Attack or failure scenario
A script issues a continuous stream of `GET /api/gbfs/<encoded-allowlisted-GBFS-url>` requests; each one triggers a fresh outbound fetch with no limiter slowing the rate.

## Root cause
A rate limiter was applied inconsistently across proxy routes in this file; this one was missed.

## Recommended fix
Add the same rate-limiting primitive used by other proxy routes in this file to `/api/gbfs`.

## Acceptance criteria
- Repeated `/api/gbfs` requests from the same client are throttled independent of the existing host/path/size validation.

## Suggested labels
bug, performance

## Priority
P3

## Severity
Low — not an SSRF or data-exposure issue given the existing allowlist, but a real availability/abuse gap.

## Confidence
Confirmed — read the full `gbfsProxy()` function; no rate limiter call present.

simplesense:
## Summary

`/api/gbfs` already allowlists hosts, requires HTTPS, restricts paths to `station_information`/`station_status`, and caps the upstream body at 5 MB — but it had no request-frequency backstop. A client could drive an unlimited number of outbound fetches against any allowlisted GBFS host.

This uses the same `makeRateLimiter` primitive as Overpass / mapped-installations:

- 90 requests per client per minute
- 300 requests global per minute
- `Retry-After: 5` and a generic `{ error: 'Rate limit exceeded' }` body

The check runs after the method guard and **before** host/path/size validation, so a flood of GETs is throttled independently of those existing checks. Non-GET requests still return 405 without spending quota.

A real bikeshare session stays well under the cap: each city is two GETs (information + status) and status polls once a minute.

Fixes #264.

## Verification

- `node --test src/data/gbfsProxy.test.mjs` — 4/4 pass (happy path, throttle-before-validation, method guard does not consume quota, plugin wiring).
- `npm test` — 2890 pass. The one failure is the pre-existing macOS tmpdir symlink comparison in `src/devCctv.test.mjs` (#300 / #301), unrelated to this change.
- `npm run build` — green.

I did not run `npm run test:track`; this change is confined to the GBFS proxy middleware and does not touch globe/track rendering.


bilawalsidhu:
Thanks, and thanks for filing this first. The GBFS rate limit is being carried forward in #690, which targets the current server layout, so closing this pre-refactor version. You'll be credited when it lands.

PR #318, opened 2026-09-11
fix(setup): pin PSModulePath so Windows credential hardening can verify its DACL

CosmicAndrew:
## The bug

On Windows with PowerShell 7 installed, `hardenCredentialFile()` always returns
`false`, so Provider Settings refuses to save any key. The panel reports failure
on a file whose permissions were in fact applied correctly.

## Why

`icacls` succeeds. The failure is in the verification step.

The spawned Windows PowerShell 5.1 inherits the session `PSModulePath`, and a
PowerShell 7 install prepends its own module directories to it for the whole
session. 5.1 then cannot load its bundled `Microsoft.PowerShell.Security`, so
`Get-Acl` does not exist, `WINDOWS_ACL_VERIFY_SCRIPT` exits non-zero, and the
caller fails closed.

Measured on Windows 11 with PowerShell 7.7 present:

```
inherited PSModulePath -> exit 1
  Get-Acl : The 'Get-Acl' command was found in the module
  'Microsoft.PowerShell.Security', but the module could not be loaded.

PSModulePath pinned    -> exit 0, DACL verified
```

The DACL itself was always correct. Querying the file after `icacls` shows
exactly the three intended principals with inheritance disabled:

```
BUILTIN\Administrators:(F)
NT AUTHORITY\SYSTEM:(F)
<machine>\<user>:(F)
```

## The fix

Pin `PSModulePath` for the spawned interpreter to its own `Modules` directory.
It is derived from `tools.powershell`, which `resolveWindowsNativeTools()` has
already canonicalised and checked, so this does not re-trust the environment a
second time.

## Verification

This repo's own test catches it. On an affected machine,
`Windows production hardener applies its exact DACL with native tools`:

- **fails** on `main` unmodified
- **passes** with this change

Both runs were done in a clean worktree checked out at `7596522`, changing only
this file. The rest of `src/keySetupHardening.test.mjs` passes in both states.

Note this is fail-closed, so the impact is a refusal to save rather than a
credential written with weak permissions.


bilawalsidhu:
Thanks for the contribution. Closing as superseded by merged #508 and current server/standalone/key-setup-hardening.mjs. The native Windows PowerShell module path and inherited PSModulePath isolation are present. The focused hardening tests pass on this Mac; the real Windows-only probe was skipped, so this is not a fresh Windows runtime certification.

PR #324, opened 2026-09-12
fix(proxies): register upstream proxies for the preview server too

Jejugo:
## Summary
- `configureServer` only wires Vite dev-server middleware; a production deploy runs `vite preview`, which requires `configurePreviewServer` instead.
- Every upstream proxy except `ais-live-proxy` and `track-backfill-proxies` was missing that hook, so under `vite preview` (e.g. a Docker deployment) `/api/opensky`, `/api/adsblol/mil`, `/api/celestrak`, `/api/tomtom`, `/api/firms`, `/api/terrain/heights`, `/api/adsbdb`, `/api/overpass`, `/api/route`, `/api/gbfs`, and `/api/cctv` all silently fell through to the SPA fallback (`index.html`), which the client then failed to parse as JSON and surfaced as a generic "network error" for every live-data layer.
- Refactored each affected proxy to a shared `install(middlewares)` function invoked from both `configureServer` and `configurePreviewServer`, mirroring the pattern already used by `ais-live-proxy` and `track-backfill-proxies`.

## Issue Found

**Symptom:** in a production build (`npm run build && npm run preview`, e.g. the Docker deployment described in the project's own packaging notes), every live-data layer — flights, ships, satellites, wildfires, traffic, cameras, terrain heights, aircraft lookups, OSM/routing data — reported a generic "network error" / UNAVAILABLE state, regardless of whether the corresponding API key was configured correctly.

**Root cause:** each of these layers is backed by a Vite plugin that proxies a third-party API (`/api/opensky`, `/api/celestrak`, `/api/tomtom`, etc.). These plugins register their middleware only inside `configureServer`, which Vite invokes exclusively when running `vite dev`. A production run instead uses `vite preview`, which calls `configurePreviewServer` — a separate hook. 9 of the 11 proxy plugins never implemented `configurePreviewServer`, so under `vite preview` those `/api/*` routes simply didn't exist. Requests to them fell through to the SPA's `index.html` fallback (a 200 response of HTML), the client's `response.json()` call then failed to parse it, and the generic `catch` block reported a network error — masking the real problem and making it look
like an auth/API-key issue instead of a routing one.

**How it was found:** ran `npm run build && npm run preview` locally and curled each `/api/*` endpoint directly — every affected route returned the SPA's `index.html` instead of a JSON response from its proxy handler, while `ais-live-proxy` and `track-backfill-proxies` (the two plugins that already had `configurePreviewServer`) responded correctly.

**Fix:** extracted each affected plugin's route handler into a shared
`install(middlewares)` function and invoked it from both `configureServer` and
`configurePreviewServer`, matching the pattern already used by `ais-live-proxy` and
`track-backfill-proxies`.

## Test plan
- [x] `node --check vite.config.js`
- [x] `npm run build && npm run preview`, then verified each endpoint returns a proxy JSON response instead of the SPA fallback:
  - `/api/opensky` → 200, real OpenSky payload
  - `/api/adsblol/mil` → 200
  - `/api/celestrak/active` → 200
  - `/api/tomtom/...`, `/api/adsbdb`, `/api/cctv` → proper 404/400 JSON from the proxy handler (not `index.html`)
  - `/api/gbfs`, `/api/terrain/heights` → proper 400 JSON validation errors from the proxy handler


samehkhamis:
Thank you, @Jejugo! Your provider-hook fix has landed in #372, adapted to the current modules with you preserved as the commit author. The separate preview port configuration change is not included.

bilawalsidhu:
The substance of this landed on main as a0f07b7 and 280fc00 (PR #372), under your authorship: every provider now registers for preview, and unmatched /api routes return JSON. The remaining diff targets vite.config.js, which no longer holds proxy code. Closing with thanks.

PR #325, opened 2026-09-12
Add traffic lights layer and live feed proxy

belugaman9981:
Introduce a new Traffic Lights layer that loads nearby OSM signal locations, renders an inspector panel, and tracks layer state alongside existing map layers. Add timing models plus a Vite server proxy for regional/public signal feeds, including automatic Hamburg SensorThings support and optional operator-configured adapters. Update attributions, environment/docs, and coverage/tests for timing rules, provider parsing, proxy validation, QA flow, and the voice-layer schema.


bilawalsidhu:
At the branch head vite.config.js is a 0-byte file, so the traffic-signals proxy is never registered and the build cannot run; the PR also bundles a separate 'simple' interface that becomes the default for every first visit. Closing; the traffic-lights layer on its own, from a fresh branch against server/providers, would be worth resubmitting.

Issue #292, opened 2026-09-11
Windows: Provider Settings silently refuses to save keys when the dev server is launched from PowerShell 7 (PSModulePath inheritance breaks Get-Acl)


Wilkam48:
# Windows: Provider Settings silently refuses to save keys when the dev server is launched from PowerShell 7 (`PSModulePath` inheritance breaks `Get-Acl`)

## Summary

On Windows, saving any key through **POWER UP → Provider Settings** fails when `npm run dev` was started from a PowerShell 7 (`pwsh`) session. No `.env` is written and no key is stored.

The cause is not the key or the panel. `hardenCredentialFile()` in `src/keySetupHardening.mjs` verifies the resulting DACL by spawning Windows PowerShell 5.1 and calling `Get-Acl`. That child process inherits `PSModulePath` from the parent, which under pwsh 7 contains PowerShell 7's module directories. Windows PowerShell 5.1 then finds PowerShell 7's `Microsoft.PowerShell.Security` module alongside its own, hits duplicate extended type data, and refuses to load the module — so `Get-Acl` does not exist, verification fails, and the write correctly fails closed.

Failing closed is the right behaviour. The problem is that a normal launch environment triggers it.

## Environment

- Windows 11 Pro 10.0.26200
- Node v24.18.0, npm 11.16.0
- gods-eye-view 0.1.1, terminal install (`git clone` + `npm ci`), not Pinokio
- Windows PowerShell 5.1.26100.9168 present and healthy
- PowerShell 7.6.5 installed (from the Microsoft Store / WindowsApps package)

`npm run doctor` reports everything OK; the failure only appears on save.

## Reproduction

1. Open a PowerShell 7 (`pwsh`) terminal.
2. `npm run dev`
3. Open the app, click **POWER UP**, paste any valid key (e.g. `CESIUM_ION_TOKEN`), click **SAVE KEYS**.
4. The key is not saved and no `.env` is created.

## Evidence

`hardenCredentialFile()` called directly on a temp file in the repo root:

```
hardenCredentialFile => false
```

Instrumenting the three steps shows `whoami` and `icacls` both succeed, and the PowerShell verification step fails:

```
whoami status 0 "willsound\administrator","S-1-5-21-...-500"
icacls status 0  Successfully processed 1 files; Failed processing 0 files
verify status = 1
Get-Acl : The 'Get-Acl' command was found in the module 'Microsoft.PowerShell.Security',
but the module could not be loaded.
[...]

PR #332, opened 2026-09-12
fix(windows): isolate ACL verifier from PowerShell 7 modules

onionviolet:
## Summary\n\n- give the Windows PowerShell 5.1 ACL verifier its native `Microsoft.PowerShell.Security` module path\n- remove inherited `PSModulePath` variants case-insensitively before spawning the verifier\n- similarly replace case-variant copies of the two private verifier variables, so Windows environment-key folding cannot select stale values\n- document the restored Provider Settings behavior\n\nFixes #292.\n\n## Why\n\nWhen the dev server starts under PowerShell 7, its inherited `PSModulePath` can point Windows PowerShell 5.1 at incompatible PowerShell 7 type data. `Get-Acl` then fails to load and credential writes correctly fail closed. The verifier only needs the native 5.1 security module, so it now receives that deterministic path while preserving unrelated environment values.\n\n## Verification\n\n- `node --test src/keySetupHardening.test.mjs`: 12 passed, 1 Windows-only production probe skipped on macOS\n- `npm run format:check`: 59 adopted files passed\n- `npm run build`: passed\n- `npm test`: 2,946 passed, 1 platform skip, 1 pre-existing macOS `/var` vs `/private/var` failure in `src/devCctv.test.mjs` (#300, fix open as #301)\n\nThe refreshed Windows CI job is the acceptance gate for the real platform path.


bilawalsidhu:
Thanks for the contribution. Closing as superseded by merged #508. Current key-setup hardening includes the native module path and case-insensitive environment-key isolation, with corresponding tests. Thank you for the verifier test coverage. The focused Mac run passed with the real Windows-only probe skipped; any new Windows reproduction should be reported against current main.

PR #336, opened 2026-09-12
feat(cctv): add the Rijkswaterstaat motorway camera pack

evertide:
## Why

`CONTRIBUTING` lists city packs as the best first lane. This adds the Dutch
motorway network: Rijkswaterstaat publishes 26 cameras as open data with
coordinates, covering the A1–A27.

## What

A fourth keyless source pack alongside Austin, Caltrans and TfL, following the
existing pack shape.

**The frames need a Referer, and the failure without one is silent.** Measured
2026-09-11: `stream.inmoves.nl` answers a frame request with HTTP 401 and a
1093-byte 320×180 placeholder PNG when no Referer is sent, and with the real
JPEG when one is. A pack pointing there without the header does not error — it
renders as blank cameras, which is the hardest failure to diagnose from the
client. `cctvUpstreamHeaders()` keys the header on hostname and is used at
both upstream fetch sites, the frame proxy and the media proxy, so the two
cannot drift apart.

**The pack leads the merge order rather than trailing it.** The catalog cap is
a blind `slice()` of the merged list, so a 26-camera national pack added last
is the first thing silently dropped once the other packs grow past
`CCTV_MAX_SOURCES`. Current totals are 826 against a default cap of 900, which
is close enough that the ordering is load-bearing rather than theoretical.

Motorway gantries get their own priors: shallower pitch, longer range and a
higher mount than a street camera. The register carries no bearing, so heading
falls back to the id hash at low confidence, the same as headingless Austin and
TfL cameras, and the drag-to-calibrate gizmo corrects it.

## Verification

End to end through the proxy: `/api/cctv/frame/rws-24` (A13 at Delft) returns
12,641 bytes of `image/jpeg`. Four tests assert both sides of the header
decision; three mutations — never send it, always send it, match the whole URL
instead of the host — each fail at least one. Full suite green.


bilawalsidhu:
Rijkswaterstaat's copyright page applies CC0 to text only and excludes images, which require permission, and the stills host only serves frames to requests presenting their site as the Referer. We can't merge on that basis. If you know of an open-data image path for these cameras, or have permission from RWS, please say so and we'll revisit.

evertide:

PR #337, opened 2026-09-12
feat(ndw): live Dutch road speeds as a viewport-bounded layer

evertide:
## Why

The globe shows where traffic is; it does not say how fast it is moving. NDW
publishes the Dutch national road-speed network as open data, keyless: 20,532
measurement sites refreshed every minute.

## What

A `Road speeds (NL)` layer — one point per gantry, coloured by its slowest
lane — and the two readers the feed needs.

**The colour is the slowest lane and never the average**, because a blocked
lane beside a free one is what a jam is. Measured on the Van Brienenoordbrug at
16:37 local: one direction ran 102 km/h while the other sat at 39 km/h with
8,400 vehicles an hour behind it. An average paints that green.

**Folding by position is what makes it legible.** A motorway cross-section
publishes one site per lane and vehicle class — fourteen share a single
coordinate on the A20 at Kethelplein — so drawn raw it is fourteen stacked
points saying fourteen different numbers about one place. 20,532 sites fold to
12,297 gantries.

**Neither body fits in a string.** `trafficspeed.xml.gz` is 53 MB of DATEX II
with not one coordinate in it; the site table that holds the coordinates is
336 MB, which is ~670 MB of UTF-16. Both readers are chunk-fed and split on
record boundaries, carrying the unterminated tail across chunk edges.

Everything expensive stays in the proxy, behind a six-hour TTL for the site
table and single-flighted so a viewport pan cannot start a second read. The
browser is handed only the rows inside its own viewport; a view wider than 2.5°
asks for nothing and clears rather than pinning a stale regional picture under
a globe view. A failed poll leaves the previous points up.

Naming is uneven and the layer says so rather than papering over it: the
provinces publish `N470 km 8.949 Re`, Rijkswaterstaat publishes MONICA codes
like `00D01001D404D007000B`. Where sensors at one mast disagree, the human name
wins, whichever order they arrive in.

## Verification

Measured against the live feeds: 20,532 measurements parsed in 0.4 s, the site
table joined at 100% in 1.0 s, 281 MB peak heap — filtering on the wanted ids
before parsing the rest of a 17 KB record is the difference between that and
570 MB. A Randstad viewport returns 600 gantries in 111 KB: 334 flowing, 163
busy, 98 slow, 5 stopped.

Two defects the tests caught before this shipped: `<values>` also starts with
`<value`, so matching the plural wrapper returned every site name with its own
XML tag glued to the front; and the record scanner matched an opening tag by
prefix, so a record could start inside its own child —
`measurementSiteRecordVersionTime` contains the whole of
`measurementSiteRecord`.

22 tests, each asserting both sides of its decision. Six mutations — averaging
zero-input lanes, averaging flow instead of summing it, matching the plural
value wrapper, dropping the tag-name boundary, a fixed band, and an unbounded
box — each fail at least one. Full suite green.


evertide:

Issue #70, opened 2026-08-26
[feature request]Add a transport layer and use GTFS datasources to populate rail/bus lines


dawtcalm:
Many bus/rail operators publish gtfs data now, so beyond your aviation/maritime/traffic data a transit layer would be VERY interesting!

350d:
Im ready to contribute Estonian public transportation realtime data.

dawtcalm:
this is probably a good start for collecting from many cities at once: 
https://mobilitydatabase.org/

bilawalsidhu:
Shipped in #587: live public-transit vehicles from open GTFS-Realtime feeds. Thanks!

PR #338, opened 2026-09-12
feat(trains): place Dutch trains from stop-time predictions

evertide:
## Why

Issue #70 asked for transit. Trains are the half that cannot come from a
vehicle-position feed: the open Dutch train feed publishes 2,982 TripUpdates
and **not one coordinate**. It is arrival and departure predictions per
station, so a train has no position until one is inferred.

## What

A `Trains (NL)` layer that places each train between the two stops it is
currently between, interpolated by time.

**The result is modelled, not observed, and the layer says so on every
selection card.** A straight line between two cities reads as the track;
through a river bend it visibly does not, and a globe implies a precision the
feed never had unless the layer contradicts it.

**A train between arriving at a platform and departing it is drawn at the
platform.** Treated as 100% of the previous leg it would sit a carriage short
of the station for as long as it waits — which is exactly when a viewer is
looking at it.

Four pieces, each testable on its own: a TripUpdate decoder (`gtfsRealtime.js`
deliberately skips FeedEntity field 3), a `stops.txt` parser, the placement,
and a proxy caching the feed for 30 s and the stop table for 12 hours.
Placement runs per request because the clock is the input that moves; the feed
only changes once a minute.

Reading the stop table is the awkward part: the national GTFS archive is 232 MB
and `stops.txt` inside it is 1.5 MB compressed. `src/data/gtfsZipMember.js`
reads one member out of the remote archive with three ranged requests — end
record, central directory, member — using `fetch` and `DecompressionStream`, so
no archive dependency is added and the same code runs under `node:test`.

## Three defects worth recording

- **GTFS-RT times are int64, not sint64.** Read as zigzag, a departure of
  1789180200 decodes as 894590100 — May 1998 — and every train is positioned by
  a clock nearly thirty years out.
- **`Number('')` is 0**, a finite coordinate off the coast of Ghana, so a blank
  `stop_lat` has to be rejected before the conversion rather than by a range
  check after it.
- **The upstream answers a throttled request with an HTML page.** `pbf` walks
  `<htm` as wire types until it hits a group tag and throws, taking the layer
  with it. The decoder now returns an empty feed instead, and neither the trips
  nor the stop table is ever replaced by an empty parse.

## Verification

Measured live: 2,982 trips in the feed, 292 running, 26 more than three minutes
late. 27 tests across the four pure modules, each asserting both sides of its
decision. Five mutations — ignoring the platform dwell, reading times as
zigzag, drawing cancelled trips, allowing a blank coordinate, and letting
unplaceable stops stretch a leg — each fail at least one. Full suite green.


evertide:

PR #339, opened 2026-09-12
Add IP Address/Server Tracker - spherical RTT-Based multi-vantage centroid based Tracking

HadiMuhammed:
IP Address or Domain Name Latitude and Longitude Tracking Using RTT to Distance Converstion
Got Success in Germany (Explained in Readme in Docs Folder)

This one is different,
Normal IP tracking usually performs a database lookup. It fetches existing records associated with an IP address But
This API works differently. It performs active network measurements using GlobalPing probes distributed across multiple locations.

GlobalPing Network Location Estimation API
Adds a server-side API for estimating the network location of IP addresses and server domain names using GlobalPing’s distributed multi-vantage measurement network.

The API sends probes to a target, collects minimum round-trip times, and calculates a spherical RTT-weighted geographic centroid from the probe coordinates. It also provides a refined estimate that groups co-located probes, uses median RTT values, and reports geographic spread as an uncertainty indicator.

The API supports optional GlobalPing authentication, configurable probe limits and continent filters, raw probe landmarks, estimated coordinates, refined coordinates, and a Google Maps link.

GET /server/ip-tracker?q=<ip-or-domain>&token=<Global-Ping-Token>
/server/ip-tracker?q=speedtest.hetzner.de&token=token

There are Limitations - Explained in Readme in Docs Folder
Uses ICMP requests, most of Devices have firewall which blocks ICMP (NEED MORE WORK)


bilawalsidhu:
This lets any browser reaching the dev server order live probes against an arbitrary host through the operator's GlobalPing quota, accepts the token as a query parameter, has no rate limiter, and is advertised as an IP tracking API. That's outside the project's rules on client-supplied targets and on tracking. Closing.

PR #341, opened 2026-09-12
Add CCTV camera sources for Chennai

DOOMSLAYERHEX:
Added multiple CCTV camera sources for Chennai with details including ID, name, location, and streaming URLs.


bilawalsidhu:
The five cameras here point at Google's sample MP4 bucket, not Chennai feeds, with invented poses and a license field that isn't a license, written into the default Austin pack. A real Chennai source would need the feed's terms in DATA_SOURCES.md and a credit in dataCredits.js. Closing.

PR #344, opened 2026-09-12
MapUpdate vite.js

nivspatel-commits:


bilawalsidhu:
This hardcodes a Google API key into build/vite.js (and the resulting object pattern doesn't parse). Keys never go into source; please revoke that key in Google Cloud, since it is now public in this PR's history. Closing.

PR #348, opened 2026-09-12
fix(terrain): stop /api/terrain/heights timing out when Re:Earth slows

ItzDESTAQ:
Cold terrain requests can exceed the browser's 30-second deadline when Re:Earth slows. Both browser and upstream requests now use sequential batches of 64 points. Successful chunks remain cached if another chunk fails; absent readings and malformed responses remain distinct, and missing heights never become fabricated zeroes.

The original terrain resilience work is contributed by @ItzDESTAQ. Maintainer follow-up matches browser batching to the server, narrows absent-reading validation, adds regression checks, and retains development and preview route support.

Validation on the final file tree:
- Doctor, unit/allocation tests (2,980 passed, one platform skip), formatting, package boundaries and production build passed.
- Tracking: 108 passed; grounded-aircraft terrain-outage checks: 5 passed.
- Controlled cold-cache comparison at 186 ms/point: current main and server-only batching each delivered 0/200 real heights before the browser deadline; matching browser/server batches delivered 200/200, with the first 64 cached in about 12 seconds and the full set in 37 seconds.
- Live browser: 200 new points across San Diego, Heathrow and San Francisco resolved in 7.8 seconds with no fallbacks. An 800-camera cold catalog resolved all priors without a failed terrain request. Settled London and San Diego placement was inspected with Google 3D tiles loaded; aircraft outage evidence uses SwiftShader.

This reduces timeout failures. It does not eliminate the initial cold-loading delay: CCTV still waits for the full sequential catalog batch before applying late priors, which took about 50 seconds in the local check.


samehkhamis:
Thank you, @ItzDESTAQ, for the terrain resilience fix!

PR #349, opened 2026-09-12
feat: add Windy Webcams CCTV provider

Mfficiency:
## Summary

- add an opt-in, server-only Windy Webcams API v3 provider for Geneva, Vaud, and Valais
- validate and register Windy image/detail URLs through the existing CCTV proxy
- refresh expiring webcam URLs, expose source links, and include required attribution
- add focused provider and key-safety tests plus the required source/runtime documentation

This is intentionally meant to be a clean, simple push: it is branched directly from `main` and contains only the minimal Windy webcam integration. It excludes the unrelated Geneva catalog, navigation controls, and broad formatting changes from the original development branch.

## Verification

- `npm run format:check`
- `npm run check:boundaries`
- `npm test` (2,954 passed, 6 skipped)
- `npm run build`
- focused Windy provider and server-only key tests


bilawalsidhu:
Thanks for a careful implementation. We read the full Windy Webcams API terms and two clauses block this for now: users may not publish any graphics displaying the webcam images (so no screenshots or recordings of the app with Windy on), and the free tier's images carry ads that must not be blocked. There is also a size clause (original or smaller only), and the required attribution string is "Webcams provided by windy.com - add new webcam". Parking this rather than closing. If you see a reading of those clauses we're missing, or a way to gate the pack as personal and non-recordable, we'd like to hear it.

PR #350, opened 2026-09-12
fix: geocode place search without a Google Maps key

sendmebits:
## Summary
- Named-place fly-to (location search box and voice “take me to …”) currently **throws** when `GOOGLE_MAPS_API_KEY` is unset, even though the globe is designed to run keyless.
- This keeps Google Geocoding first when a valid Maps key is present, and falls back to the **existing** Nominatim proxy (already used for cockpit reverse-geocoding) so search still works without Google billing.
- Not a new vendor: same public Nominatim instance, User-Agent, 1 req/s queue, ODbL attribution, and `DATA_SOURCES.md` / data-credits update.

## Why
A Cesium ion token cannot substitute for Geocoding. Without this fallback, a keyless (or ion-only) setup can talk to the globe and load 3D tiles, but typed/spoken place navigation fails closed.

## Policy / license
Nominatim is already listed in `DATA_SOURCES.md` (ODbL 1.0 + Nominatim usage policy). Forward search:
- shares the existing 1.1s Nominatim queue
- identifies as `GodsEyeView/0.1 (+https://github.com/bilawalsidhu/gods-eye-view)`
- caches only a short in-memory window of user queries (not a bulk gazetteer)
- does not store Google responses

## Test plan
- [ ] No Google key: type `Tokyo` (or say “take me to Tokyo”) and confirm the camera flies there
- [ ] No Google key: a nonsense query still shows location not found (does not throw)
- [ ] With a valid `GOOGLE_MAPS_API_KEY`: search still uses Google first
- [ ] Preset city/POI pills still fly without any geocode request
- [ ] Data attribution popover still credits © OpenStreetMap contributors via Nominatim
- [ ] `node --test src/locations.test.mjs src/nominatimGeocode.test.mjs` passes


sendmebits:
Rebased onto current `main` and moved the Nominatim search endpoint into the extracted regional place provider (`server/providers/regional/place.js` + `geocodeProxy()`), instead of the old monolithic `local.js`.

sendmebits:
Rebased onto current `main` (including the Photon place-search chain). Nominatim `/api/geocode` is now last-resort after Google and Photon, instead of replacing the new `placeSearch` providers.

sendmebits:
Manually tested and this is working.

bilawalsidhu:
Landed via #573 (rebased onto the current component layout, with fixes from review). Thank you @sendmebits!

Issue #40, opened 2026-08-25
Reduce oversized production chunks and data assets


jsawyerdev:
Reviewed commit: 880a672b5e16ad3e41d318801d3a5203f9201923

Problem:
- The production build emits very large chunks/assets and Vite chunk-size warnings.
- Several data-heavy features are lazy, but the resulting chunks are still large enough to hurt cold feature use and mobile/network-constrained sessions.

Evidence:
- `npm run build` reports chunk-size warnings.
- Build output includes `egm96-universal...js` around 2.77 MB uncompressed / 1.87 MB gzip.
- Build output includes `regions...js` around 1.99 MB uncompressed / 686 KB gzip.
- Build output includes `index...js` around 1.35 MB uncompressed / 416 KB gzip.
- Build output also includes multi-MB data assets such as datacenters and dams datasets.

Suggested fix:
- Audit which datasets must ship by default versus load on demand from static assets/API endpoints.
- Split heavy data by region/feature and fetch only after the user enables the layer.
- Add bundle budgets for initial JS and feature chunks.

This was found during an adversarial repository review; impact depends on how the dev/preview server is exposed.

Issue #24, opened 2026-08-25
Move realtime debug-log redaction to the server


jsawyerdev:
Reviewed commit: 880a672b5e16ad3e41d318801d3a5203f9201923

Problem:
- Sensitive debug-log redaction is enforced by the browser client, not by the server endpoint.
- A direct POST to `/api/realtime/debug-log` can write arbitrary payloads into the log file.
- This contradicts the security documentation claim that debug logs are redacted before writing.

Evidence:
- `src/voice/gevRealtime.js:2004-2011` posts client-sanitized debug payloads.
- `src/voice/gevRealtime.js:2143-2178` contains client-side redaction logic.
- `vite.config.js:5020-5043` writes whatever the server receives.
- `SECURITY.md:44` describes redacted debug logging as a safety property.

Suggested fix:
- Treat client redaction as a convenience only.
- Enforce server-side schema validation and redaction before writing any debug event.
- Reject unknown event shapes instead of storing arbitrary JSON/text.

This was found during an adversarial repository review; impact depends on how the dev/preview server is exposed.

bilawalsidhu:
Thanks for the report. Consolidating this into #23; its remaining requirements are preserved there. Same endpoint and remediation; client-side redaction does not protect direct POSTs. This closure does not mean the remaining work is fixed.

Issue #23, opened 2026-08-25
Add auth, rate limits, and bounded async writes to /api/realtime/debug-log


jsawyerdev:
Reviewed commit: 880a672b5e16ad3e41d318801d3a5203f9201923

Problem:
- `/api/realtime/debug-log` accepts unauthenticated POST bodies up to 8 MB.
- It writes to disk synchronously per request and has no total log-size cap or rotation.
- Repeated requests can fill disk and block the event loop.

Evidence:
- `vite.config.js:1327-1329` defines `.gev-logs/realtime-conversations.jsonl` and an 8 MB request cap.
- `vite.config.js:5020-5043` implements the endpoint with `fs.mkdirSync` and `fs.appendFileSync`.

Suggested fix:
- Require auth or disable the endpoint unless explicitly enabled for local debugging.
- Add route-level rate limits.
- Use bounded asynchronous logging with rotation/retention limits.

This was found during an adversarial repository review; impact depends on how the dev/preview server is exposed.

bilawalsidhu:
Consolidating #24 here. Keep both performance/resource bounds and server-side sanitization: direct POSTs must not bypass redaction of credentials or sensitive content; bound request size, queued writes and retained files; avoid synchronous filesystem work on the request path; and specify the non-local access policy. Client-side redaction alone is insufficient. #354/#355 remain implementation inputs requiring review.

bilawalsidhu:
Consolidating into #16, which tracks access to the cost-bearing endpoints when the server is exposed beyond localhost.

PR #354, opened 2026-09-12
fix(realtime): bound the debug-log sink and stop blocking on its writes

Lob26:
**Intent:** harden `/api/realtime/debug-log`, the one route in its family left unhardened. It deliberately does not add authentication, and does not touch the two sibling routes' existing opt-in limiter.

Refs #23. **Rebased onto the split provider layout** — the original branch was written against `server/providers/local.js`, which has since gone from 4671 lines to 89. The route now lives in `server/providers/openai/debug-log.js`, and every defect moved with it intact.

## What changed

Both siblings enforce a limiter through `openai/rate-limit.js`; this handler had none, and held the only synchronous `fs` writes left on a request path.

- **Always-on per-IP limiter** (120/min, 400 global) via `makeRateLimiter`. Deliberately *not* `openAiRateLimiter()`: that one defaults to unlimited so throttling cannot silently change what a user spends. This sink spends disk and event-loop time rather than provider quota, so the cap is unconditional and sits far above what a voice session writes.
- **Serialized async appends** with a 32 MB rotation ceiling and one retained generation. `REALTIME_DEBUG_LOG_MAX_BYTES` (8 MB) caps a single *request body*; nothing bounded the file those bodies accumulate into, so any local page could grow `.gev-logs/realtime-conversations.jsonl` for as long as the dev server ran. Serializing also closes a read-then-write race on the size check, where concurrent callers all `stat` under the ceiling and none rotates.
- **Generic 400/500** separating a malformed record or oversized body from a failed write, instead of echoing `error.message` — an errno and an absolute path for an fs failure.
- `readRequestBody` marks an oversized body with `code = 'BODY_TOO_LARGE'`, the marker its sibling `readRequestBodyCapped` already sets, so the two cases are distinguishable without matching on text.

State belongs to the handler the factory returns rather than to module scope, which is what made the tests below possible against the real route.

No contract change: the success path is still `204` with an empty body.

## How it was verified

Re-verified on the synced candidate `4fd168f` (`origin/main` merged in at `2746ec7`; only `CHANGELOG.md` conflicted, and both sides are kept). Node 26.5.0 / Windows 11:

- `npm test` → `pass 4016 · fail 0 · skipped 10`
- `npm run format:check` → `Checked 850 source files.`
- `npm run check:boundaries` → all groups pass
- `npm run build` → `✓ built in 17.04s`
- `npm run doctor -- --json` → `ready: true`, node `26.5.0` "supported runtime"
- `npm run test:track` → **`109 passed, 0 failed, 0 skipped`**, including `no console errors across the FULL run` and `no HTTP 5xx responses across the FULL run`

That last line replaces the earlier "not run": Chrome was the missing piece and is installed now, so the third gate `CONTRIBUTING.md` requires is real evidence here rather than a declared gap.

Two skips declared rather than omitted: `npm test` skips 2 allocation microbenchmarks because those budgets are calibrated for Node 24 and this is 26.5.0 (the harness prints that itself), and the build's chunk-size warning is pre-existing (#40). Neither relates to this change.

The original run, on the pre-merge head `44817e1`, read `tests 3589 · pass 3580 · fail 0 · skipped 9` and `Checked 638 adopted files.` — those counts moved because `main` did, not this branch.

Coverage extends `src/tooling/localServices.test.mjs`, which already drives the real handler over a temporary `sourceRoot` — so these are behavioural tests, not source assertions. **Both new tests fail on the previous implementation** (`pass 6 · fail 2`) and pass with this one.

Per `docs/MAINTAINER_WORKFLOW.md` gate 2, `CHANGELOG.md` and `docs/CURRENT-STATE.md` are updated in this PR.

## Notes for reviewers

The 120/min figure and the 32 MB ceiling are the two numbers most worth an opinion.

One assumption I tested and dropped: I first justified the append queue as preventing interleaved partial records. It is not needed for that — `appendFile` is `O_APPEND`, and the test still passed with serialization removed. The queue is there for the rotation race, and the test asserts that instead; without it the live log reached 3702 bytes against a 300-byte ceiling in the standalone reproduction.

The rate-limit test asserts 119 accepted writes rather than 120, because the limiter counts requests and the malformed record above already spent a slot. Intended, and called out so it does not read as an off-by-one.

## Out of scope / follow-up

- **No authentication.** These proxies have no auth mechanism today, so introducing one is a maintainer design decision. The rate-limit and bounded-write halves of #23 are addressed; its auth half is not.
- #24 (moving debug-log redaction server-side) is the same route but a separate concern.
- The identical raw-error echo on the two sibling routes is #355, kept separate so neither PR overlaps a line of the other.

## Delivery readiness

- **Rollback**: revertible alone; no shared state, no migration, no config.
- **Config/secrets**: no new environment variables; the limiter is unconditional by design.
- **Breaking change**: none. The only observable differences are a `429` past 120 writes/min and a fixed error string where a raw one used to appear.

> 🤖 Claude Code helped


samehkhamis:

Issue #263, opened 2026-09-10
Inconsistent error-message sanitization: celestrak/terrain/adsbdb proxies echo raw JS error text


jsawyerdev:
## Summary
The Celestrak, terrain-heights, and adsbdb proxies each echo the raw JavaScript exception text (`err?.message || err`) back to the client on unexpected errors, instead of a generic message, unlike the sanitized-error convention used elsewhere in `vite.config.js`.

## Evidence
Three separate handlers, each with the same pattern:
- `vite.config.js:1647-1649` (celestrak): `send(500, \`celestrak proxy error: ${err?.message || err}\`, 'ERROR');`
- `vite.config.js:2375-2377` (terrain heights): `send(500, { error: \`terrain heights proxy error: ${err?.message || err}\` });`
- `vite.config.js:2496-2498` (adsbdb): `return send(500, { error: String(err?.message || err) });`

This is a second, independent counter-example (alongside issue "launches forwards raw upstream error body") to the `docs/KNOWN-ISSUES.md` "Closed / Intentional" claim that "Error payloads are sanitized" as a general property of `vite.config.js`'s proxy layer — it holds for most routes but not these three.

## Why this matters
`err.message` in a JS `catch` block can include stack-adjacent detail, malformed-input echoes, or library-internal strings not intended for end users. This is low severity here (no evidence of secret material reaching these specific catch blocks), but it is a real, fixable gap in an otherwise-documented security property, and a template for what will keep happening if new proxy routes are added without checking the existing error-sanitization helper.

## Attack or failure scenario
An unexpected exception in any of these three handlers (a network library throwing an unusual error object, a parsing edge case) surfaces its raw `.message` text directly in the HTTP response body.

## Root cause
No shared "sanitize this error before sending" helper is consistently used across all proxy routes in this file; three routes fell through the cracks.

## Recommended fix
Introduce (or reuse, if one already exists) a single helper that maps any caught error to a generic, route-specific message for the client while logging the real error server-side via `console.warn`/`console.error`, and apply it to these three routes.

## Acceptance criteria
- None of the three cited routes place `err.message` (or the error object) directly into the HTTP response body.
- A shared pattern/helper exists so future routes don't reintroduce this.

## Suggested labels
bug

## Priority
P3

## Severity
Low — information-disclosure hygiene issue, not a credential or data leak; requires an unexpected internal error to trigger.

## Confidence
Confirmed — read all three cited call sites directly.

[...]

PR #355, opened 2026-09-12
fix(server): stop relaying upstream and JS error text to the client

Lob26:
**Intent:** stop the three remaining paths that relay raw upstream or JS error text to the client. It deliberately leaves `/api/realtime/debug-log` alone — that route is #354, so the two do not overlap a line.

Refs #263. **Rebased onto the split provider layout**: the routes now live in `server/providers/openai/{hud-summary,realtime}.js` and the camera health entry in `server/providers/cctv.js`.

## What changed

The routes #263 names — celestrak, terrain, adsbdb — were sanitized by #289. The same defect had moved:

- **`/api/openai/hud-summary`** relayed `data.error.message` verbatim on a non-ok upstream. That is OpenAI's own text, carrying request ids, organization hints and quota wording. Its `catch` echoed the JS error too.
- **`/api/realtime/token`** echoed the JS error when minting failed, which for a network fault is a resolver message naming the upstream host.
- **`cctvProxy`** stored the raw error as a camera's health `message`. This one is the least obvious, and the reason the PR is worth its size: `GET /api/cctv/health` serializes `message`, and **`src/layers/cctv/frames.js:192` renders it as the camera's status label**. The detail reached the screen through a different door than the sanitized `502` sitting beside it — at worst an uppercased `CONNECT ETIMEDOUT <ip>:443` with a local path. It now reads `MEDIA FETCH FAILED`.

That last one is a user-visible change. An improvement, I think, but a change — worth your eyes.

Logs name the failure and the upstream status without the text, matching the Launch Library and terrain paths. These lines get pasted into issues.

## How it was verified

Re-verified on the synced candidate `3308569` (`origin/main` merged in at `2746ec7`; only `CHANGELOG.md` conflicted, and both sides are kept). Node 26.5.0 / Windows 11:

- `npm test` → `pass 4016 · fail 0 · skipped 10`
- `npm run format:check` → `Checked 850 source files.`
- `npm run check:boundaries` → all groups pass
- `npm run build` → `✓ built in 11.14s`
- `npm run doctor -- --json` → `ready: true`, node `26.5.0` "supported runtime"
- `npm run test:track` → **`109 passed, 0 failed, 0 skipped`**, including `no console errors across the FULL run` and `no HTTP 5xx responses across the FULL run`

That last line replaces the earlier "not run": Chrome was the missing piece and is installed now, so the third gate `CONTRIBUTING.md` requires is real evidence here rather than a declared gap.

Two skips declared rather than omitted: `npm test` skips 2 allocation microbenchmarks because those budgets are calibrated for Node 24 and this is 26.5.0 (the harness prints that itself), and the build's chunk-size warning is pre-existing (#40). Neither relates to this change.

The original run, on the pre-merge head `9e71081`, read `tests 3589 · pass 3580 · fail 0 · skipped 9` and `Checked 638 adopted files.` — those counts moved because `main` did, not this branch.

Coverage extends the existing harnesses rather than adding files — `localServices.test.mjs` drives both OpenAI routes over a mocked upstream, and `mediaProviders.test.mjs` drives a failing media fetch and then reads `/health` to prove the second door is closed. **Both new tests fail on the previous implementation** (`pass 10 · fail 2`).

Per `docs/MAINTAINER_WORKFLOW.md` gate 2, `CHANGELOG.md` and `docs/CURRENT-STATE.md` are updated in this PR.

## Notes for reviewers

The fixtures deliberately carry a recognisable payload — a request id, an org hint, an IP and a Windows path — and the assertions check those strings are *absent* from the response, not only that the new message is present. A test asserting the new string alone would pass against a handler that appended both.

## Out of scope / follow-up

- `/api/realtime/debug-log`'s identical echo is #354.
- The CCTV catalog download and source-file read paths still log raw error text server-side. They are logs rather than client responses, and outside what #263 describes, so I left them.

## Follow-up in this branch

`3308569` fixes a comment, not behaviour. The comment introducing the CCTV log line read "Keep the detail in the log" while the line it introduces carries none — so the next reader would have taken the omission for an oversight and added the errno back. Detail-free is deliberate here (a media-fetch failure names the camera's upstream host, and these lines get pasted into issues), and the comment now says that, plus why the `catch` at the bottom of the same file may still log its error text.

> 🤖 Claude Code helped


samehkhamis:

Issue #266, opened 2026-09-10
Pinokio auto-update runs unauthenticated git pull + reinstall with no diff or verification step


jsawyerdev:
## Summary
The Pinokio "Update" action runs an unauthenticated `git pull --ff-only` followed immediately by dependency reinstall and install scripts, with no diff review, commit-signature check, or user confirmation of what changed.

## Evidence
`scripts/pinokio-update.mjs` (full file):
```js
#!/usr/bin/env node
import { installPinokioDependencies, runChecked } from './pinokio-install.mjs';

runChecked('git', ['pull', '--ff-only']);
installPinokioDependencies();
```
Invoked from `pinokio/update.js`, reachable via a fixed menu item in the Pinokio launcher aimed at a non-technical end-user audience (per README.md, Pinokio is the recommended path for "people who only want to run the app").

## Why this matters
This is design-inherent to Pinokio's full-local-privilege model, not a new coding flaw — installation already trusts the configured remote. But it's worth documenting as a concrete trust boundary: whoever controls the configured `origin` remote (or can MITM/DNS-hijack it) gets code execution on every user who clicks "Update," with zero visible diff or confirmation step, which is a materially different risk profile than the initial, deliberate install action.

## Attack or failure scenario
If the git remote is ever compromised, or a user's local remote is repointed (e.g. by a malicious fork instruction), clicking "Update" silently pulls and executes whatever is on that remote with no warning.

## Root cause
The update flow was designed for simplicity (one-click, no prompts) without adding even a minimal safety check appropriate for a non-technical audience clicking a menu button.

## Recommended fix
Before running `npm ci`/install scripts, show the user a summary of the pending changes (e.g. `git log --oneline HEAD..origin/main` and a diffstat) and require confirmation, or at minimum log what changed to a visible location. Consider verifying commit signatures if the project adopts signed commits.

## Acceptance criteria
- The update flow surfaces what is about to change before executing install scripts, rather than pulling and running unconditionally.

## Suggested labels
enhancement

## Priority
P3

## Severity
Low — consistent with the existing trust model of a one-click installer; flagged as a hardening opportunity rather than a new vulnerability.

## Confidence
Suspected — design-inherent limitation rather than a coding defect; confirmed the code does exactly what's described, but whether this warrants a UX change vs. accepting the existing trust model is a product decision.

bilawalsidhu:
[...]

PR #356, opened 2026-09-12
fix(pinokio): disclose incoming changes before Update pulls and reinstalls

Lob26:
**Intent:** make Pinokio's Update disclose what it is about to run before it runs it. It deliberately does not add a confirmation prompt, and does not change what Update installs or how.

Refs #266.

## What changed

`scripts/pinokio-update.mjs` was five lines: `git pull --ff-only` followed immediately by dependency install. It executed whatever the configured remote served, with no indication of what changed or where it came from — and it is a fixed menu item aimed at people who only want to run the app, so the one-click path was also the path with the least context about what it just ran.

Update cannot prompt; Pinokio drives it non-interactively through `shell.run`. So the fix is disclosure rather than confirmation:

- Resolve the tracking branch, print it and **the remote's URL**, then `git fetch` that remote.
- Print the incoming commits (`git log --oneline HEAD..@{u}`) and their diffstat.
- Only then `git pull --ff-only` and reinstall.

The remote URL is the part that matters. The concern in #266 is an origin that has been repointed, and that is invisible in every other line of Update's output.

Reliability was the constraint I cared about most: a preview step that can break the update path is worse than no preview. A missing upstream or detached HEAD prints a warning, returns, and lets `git pull --ff-only` report the real problem. `readGit` returns `null` on any git failure and every caller treats that as "unknown" rather than fatal.

Side effects now sit behind `isDirectInvocation(process.argv[1], MODULE_PATH)`, matching `scripts/pinokio-start.mjs`. Without that, importing the module to test it would pull and run `npm ci` as an import side effect.

## How it was verified

On `fix/pinokio-update-transparency`, Node 26.5.0 / Windows 11:

- `npm test` → `tests 2957 · pass 2951 · fail 0 · skipped 6`
- `node --test src/pinokioLauncherContract.test.mjs src/pinokioUpdatePreview.test.mjs` → `tests 13 · pass 13 · fail 0` (4 new; the existing launcher contract still passes unchanged)
- `node scripts/pinokio-install.mjs` → `[Pinokio] Installation ready.` — this is the command the `windows-onboarding` CI job runs
- `npm run format:check` → `Checked 71 adopted files.`
- `npm run build` → `✓ built in 2.78s`
- `npm run test:track` → **not run**: needs a live dev server, a Google Maps key and Chrome. Declared rather than omitted.

Both real paths were also exercised by hand against this checkout, not only through mocks:

```
[update] No upstream branch is configured — cannot preview changes.
[update] Continuing; `git pull --ff-only` will report the problem.
```
```
[update] Tracking origin/main
[update] Fetching from https://github.com/bilawalsidhu/gods-eye-view.git
[update] Already up to date — reinstalling dependencies only.
```

Reverting the script alone makes all 4 new tests fail. One of them asserts `fetch` precedes reading the commit range, because a report built on a stale remote ref describes the wrong changes.

## Notes for reviewers

The tests drive `reportIncomingChanges` over injected git output using the same source-extraction approach as `src/proxyErrorResponses.test.mjs`, so production keeps no test-only seam and no test hits the network.

`readGit` and `reportIncomingChanges` are exported for that. If you'd rather they weren't part of the module's surface, they can be un-exported and reached purely by extraction.

## Out of scope / follow-up

- No commit-signature verification. #266 raises it as a possibility, but it only means something once the project signs commits, which is a maintainer decision.
- Still no confirmation step, by design — see above. If Pinokio has an interactive primitive I've missed, that would be the stronger fix and I'd rather use it.

## Delivery readiness

- **Rollback**: revertible alone; touches one script and adds one test file.
- **Config/secrets**: none. The env passthrough in `pinokio/update.js` is untouched.
- **Breaking change**: none. Update's outcome is identical; it now prints before acting.

> 🤖 Claude Code helped


bilawalsidhu:
Landed via #507 (rebased onto the current component layout, with fixes from review). Thank you @Lob26!

PR #371, opened 2026-09-12
fix(status): stop false LOAD FAILED from a relayed adsb.lol 429 and the installations zoom prompt

### LINKED ISSUE #370: Global status chip reads LOAD FAILED on a relayed adsb.lol 429 and on the installations zoom-in prompt
## Summary

While tracking a grounded aircraft at JFK and pressing NEXT in Cockpit, the global status chip repeatedly reads **LOAD FAILED** even though every layer is still populated and the new target loads a few seconds later. Reproduced on `main` @ aacfa06 with the chip and the data manager instrumented; two independent triggers, both false alarms.

## What the chip is

`#global-loading-status` is driven by `src/loadingFeedback.js`. It aggregates every registered layer; when a loading/refresh batch settles, the terminal label is LOAD FAILED if **any participating layer carries an `error` string at that instant** (`terminalFromParticipantStats`) or a `refresh-failed` manager event landed during the batch. The manager itself derives `refresh-failed` from the same field (`refreshFailureFromStats` reads `stats.error`).

## Trigger 1: a relayed adsb.lol 429 becomes a 45 s failure loop

- `server/providers/aircraft/adsb-lol.js` is documented as "on upstream failure, serves a stale cached response", but only the `catch` path does that. A non-OK status is relayed verbatim, so an upstream **429 is forwarded to the browser** even when a 12 s-old cache is sitting right there.
- `src/data/militaryFlights.js` treats a 429 as a 45 s cooldown (`BACKOFF_INTERVAL`). Every `update()` inside that window returns early with `_lastError = 'adsb.lol rate limited'`, so every refresh — the 15 s poll, and any tracking resolve — is reported as `refresh-failed` **without a network call**.
- Captured timeline (ms since page load), military layer only:

```
155042 refresh-transition
155208 refresh-failed  Error: adsb.lol rate limited   <- real upstream 429 relayed by the proxy
170060 refresh-transition
170062 refresh-failed  Error: adsb.lol rate limited   <- 2 ms: cooldown, no fetch
185049 refresh-transition
185049 refresh-failed  Error: adsb.lol rate limited   <- 0 ms: cooldown, no fetch
200047 refresh-transition
200242 refresh                                        <- cooldown over, recovered
```

Each `refresh-failed` flipped the chip to LOAD FAILED while the layer still held 140 aircraft. Cockpit NEXT is just when the operator is looking at the chip.

## Trigger 2: the installations "zoom in" prompt is stored as an error

- In Cockpit the camera looks at the horizon, so `viewer.camera.computeViewRectangle()` cannot bound the view and `militaryInstallations.js` sets `status: 'zoom-in'` **with the prompt text in `state.error`**.
- `layerFeedState` and the layer row already treat `zoom-in` as guidance (nominal), but `normalizeLayerLoading` and `refreshFailureFromStats` do not: any batch the installations layer participates in ends LOAD FAILED. Two of the six captured flips had only this layer in "error":

```
125084  military-installations st=zoom-in err="Zoom in to load mapped installation context" n=0
200312  military-installations st=zoom-in err="Zoom in to load mapped installation context" n=0
```

The layer never called Overpass in this state; nothing actually failed.

## Fix (PR to follow)

1. **Proxy**: on an upstream 429/5xx serve the cached body as `X-ADS-B-Cache: STALE` (200) and back off from upstream for `Retry-After` (bounded 5–120 s, default 30 s for 429 / 15 s for 5xx), mirroring the OpenSky proxy's cooldown. A 429 with nothing cached is still relayed.
2. **Installations**: `zoom-in` is set without an error string; the row and toast read the prompt from `statusMessage` (already produced by `installationFeedback`).
3. **Chip**: `normalizeLayerLoading` ignores `error` for guidance statuses (`zoom-in`, `empty`, `idle`), the same carve-out `layerFeedState` already has.

Follow-up worth its own issue: a refresh that fails while the layer still has data probably deserves a STALE terminal state rather than LOAD FAILED. The first-run code already ledgers the same aggregation gap for the keyless FIRMS "KEY REQUIRED" case.

## Environment

macOS, Node 26.3.0, `main` @ aacfa06, keyed install (OpenSky OAuth, all providers configured), Chrome.

bilawalsidhu:
Fixes #370.

Military aircraft remain available during an adsb.lol 429/5xx when the proxy holds a cached response. Retry-After is bounded to 5–120 seconds, with defaults of 30 seconds for 429 and 15 seconds for server errors. Cold-cache failures remain errors and carry a retry delay. Cached fallback cancels the failing body immediately, and cooldown starts when the response arrives.

Cached responses expose their age. The browser retains the original observation times, labels fallback data STALE, and resumes fresh updates after recovery. Installation zoom guidance stays in the layer row without turning a successful load into LOAD FAILED; independent feed and manager errors still count as failures.

Bilawal’s original commit is retained, with a follow-up fixing cached timestamps and adding regression checks. The browser’s cold-cache retry behavior and broader global loading-state design are unchanged.

Validation:

- 194 focused tests passed, including cached position replay, slow/stalled 5xx responses, Retry-After dates/bounds, recovery, and independent errors during guidance.
- Full unit/allocation suite passed (one platform skip); doctor, build, format and package boundaries passed.
- All 108 tracking/browser regression checks passed, with no console errors or HTTP 5xx across the run.
- Chromium inspection at global and oblique views confirmed the STALE aircraft row and installation zoom guidance. The first QA attempt encountered stale Vite optimized dependencies; rebuilding those in the temporary checkout resolved startup, and the final run exited successfully.


samehkhamis:
Thanks, Bilawal! Merged with the cache-age and error-handling follow-up; your original commit is preserved.

PR #392, opened 2026-09-13
fix: handle antimeridian crossing in clampBoundsAroundCenter

### LINKED ISSUE #352: clampBoundsAroundCenter() produces invalid longitude bounds when the viewport crosses the antimeridian
`clampBoundsAroundCenter()` calculates longitude span using a direct subtraction:

```js
const lonSpan = Math.min(bounds.east - bounds.west, maxSpanDeg);
```

This assumes `east >= west`.

That assumption breaks for valid geographic rectangles crossing the International Date Line, where a small viewport may be represented as:

```text
west = 179.98
east = -179.98
```

In this case the actual longitude span is approximately `0.04°`, but the current calculation produces:

```text
-179.98 - 179.98 = -359.96
```

resulting in a negative `lonSpan`.

### Reproduction

```js
const bounds = {
  south: -0.02,
  north: 0.02,
  west: 179.98,
  east: -179.98,
};

const center = {
  lat: 0,
  lon: 179.99,
};

console.log(clampBoundsAroundCenter(bounds, center, 0.05));
```

### Current result

Using the current implementation, the result is approximately:

```js
{
  south: -0.02,
  north: 0.02,
  west: 359.97,
  east: 0.01
}
```

The returned longitude bounds are outside the normal `[-180, 180]` range and no longer represent the intended small region around the dateline.

### Expected behavior

A viewport spanning:

```text
179.98°E → 179.98°W
```

should be treated as an approximately `0.04°` longitude span crossing the antimeridian.

The clamped result should remain geographically close to the requested center and preserve longitude wrapping correctly.

### Root cause

The implementation treats longitude like a normal linear coordinate:

```js
bounds.east - bounds.west
```

but longitude is cyclic.

When `east < west` because the rectangle crosses ±180°, the span must wrap through `360°`.

### Suggested fix

Calculate a wrapped longitude span, for example:

```js
function longitudeSpanDeg(west, east) {
  const raw = east - west;
  return raw >= 0 ? raw : raw + 360;
}
```

Then:

```js
const lonSpan = Math.min(
  longitudeSpanDeg(bounds.west, bounds.east),
  maxSpanDeg,
);
```

The generated `west` and `east` values should also be normalized back into `[-180, 180)`.

For example:

```js
function normalizeLon(lon) {
  return ((lon + 180) % 360 + 360) % 360 - 180;
}
```

### Regression test

Add an antimeridian case to `trafficBounds.test.mjs`:

```js
test('span clamp handles bounds crossing the antimeridian', () => {
  const bounds = {
    south: -0.02,
    north: 0.02,
    west: 179.98,
    east: -179.98,
  };

  const center = {
    lat: 0,
    lon: 179.99,
  };

  const result = clampBoundsAroundCenter(bounds, center, 0.05);

  assert.ok(result.west >= -180 && result.west < 180);
  assert.ok(result.east >= -180 && result.east < 180);

  // Wrapped longitude span should remain about 0.04 degrees.
  const span =
    result.east >= result.west
      ? result.east - result.west
      : result.east - result.west + 360;

  assert.ok(Math.abs(span - 0.04) < 1e-12);
});
```

### Impact

Near the International Date Line, traffic-road fetching can receive malformed or excessively displaced longitude bounds, potentially causing missing or incorrect traffic data for the visible region.

Ashfaqbs:
## Summary

Fixes #352.

`clampBoundsAroundCenter()` computed the longitude span as a plain `bounds.east - bounds.west`, which assumes longitude is linear. It isn't: a small viewport crossing the antimeridian can be represented as `west = 179.98, east = -179.98`, and the plain subtraction gives `-359.96` instead of the actual ~0.04° span. The clamped output then also drifted outside the normal `[-180, 180)` range.

- Added `longitudeSpanDeg(west, east)`: wraps the span through 360° when `east < west`.
- Added `normalizeLonDeg(lon)`: normalizes a longitude back into `[-180, 180)`, reusing the same wrap idiom already used by `destinationPoint()` in this file.
- `clampBoundsAroundCenter()` now uses both, so the returned `west`/`east` stay in range and the span stays correct near the date line.

Non-antimeridian behavior is unchanged: `longitudeSpanDeg` reduces to a plain subtraction when `east >= west`, and `normalizeLonDeg` is a no-op for longitudes already in range — verified by the pre-existing span-clamp tests still passing unmodified.

## Test plan
- [x] Added the antimeridian regression case to `trafficBounds.test.mjs` (from the issue's repro)
- [x] `node --test src/data/trafficBounds.test.mjs` — 10/10 pass
- [x] `npm test` (full suite) — 3085 pass, 0 fail, 6 skipped (Node-24-calibrated allocation microbenchmarks, expected on Node 22)
- [x] `npm run build` — clean
- [x] `node scripts/format.mjs --check` — clean
- [ ] `npm run test:track` — not run. This harness drives the flight-tracking camera/3D regression suite via headless Chromium against a running preview server; it doesn't touch traffic bounds or this code path, and my local Node version (22) is below the project's required 24-26, so I didn't want to report a false pass/fail on an unrelated harness. Happy to run it if you'd like before merge.


Lob26:
Contributor review, not a maintainer verdict. Checked against `main` at `a8125ad`; this one still merges clean.

**The span bug is real and the fix for it is correct.** At HEAD, `bounds.east - bounds.west` for a box straddling ±180° is about `-359.96`, so `Math.min(-359.96, 0.05)` keeps the negative value and the returned box is inverted with a ~360° span. `longitudeSpanDeg` fixes that, and the `((lon + 540) % 360) - 180` wrap is the right idiom.

**But normalizing `west`/`east` on the way out moves the failure into the callers rather than removing it.** `clampBoundsAroundCenter` has two call sites, both in `src/layers/traffic/viewport.js:167` and `:213`, and the returned box immediately feeds three consumers that all assume `west <= east`:

`getBoundsCenter` (`viewport.js:97`) is a plain mean:

```js
lon: (bounds.west + bounds.east) / 2,
```

For the test's own input the PR returns `west = 179.97, east = -179.99`, so the center comes out at **`-0.01` instead of `179.98`** — off by ~180°, about 20 000 km. That value is the `center` passed to `distanceKm(center, layerState._lastViewCenter)` in the `MIN_CENTER_SHIFT_KM` gate and is then stored as `_lastViewCenter`, so the fetch gate is comparing a point on the opposite side of the planet.

Worth being precise about the trade here: the *pre-PR* code, broken as its box is, gets the center right by accident — `(west + east) / 2` with the symmetric inverted span collapses back to `center.lon`. So this specific value regresses.

`boundsOverlap` (`viewport.js:134`) degenerates twice over: `overlapE = Math.min(a.east, b.east)` against `overlapW = Math.max(a.west, b.west)` trips the `overlapE <= overlapW` early return, and `aArea = (a.north - a.south) * (a.east - a.west)` is negative so `aArea > 0` fails. The overlap skip therefore never fires near the antimeridian and every camera change re-fetches — extra Overpass load in exactly the region the PR is trying to fix.

Third, `loadRoadsForBounds` turns the box into an Overpass bbox, which needs `west < east`.

**The test passes while the caller is broken**, which is the part I'd flag hardest. It asserts the wrapped span and that both edges are in `[-180, 180)` — both true — but not the property the callers depend on. A test asserting `getBoundsCenter(result)` lands near `179.98` would fail on this patch, and that is the assertion that would have caught it.

**Two ways forward, and I'd suggest the first:**

*Scope it to the span.* Keep `longitudeSpanDeg`, drop `normalizeLonDeg` from the return. You get `west = 179.97, east = 180.01` — outside the conventional range, but monotonic, so the span is right, `getBoundsCenter` returns `179.99`, and `boundsOverlap` keeps working. A clean, small, honest fix; the remaining "coordinates exceed 180" issue is then one visible problem at the Overpass boundary instead of a wrong number flowing through the gate logic.

*Or make the whole path cyclic* — a wrap-aware longitude midpoint in `getBoundsCenter`, seam splitting in `boundsOverlap`, and a bbox that Overpass accepts. That's the complete fix, but it's a different and much larger PR, and I'd want the maintainers to say whether traffic at the antimeridian is worth that before anyone writes it.

Either way the `longitudeSpanDeg` helper and its docstring are keepers.

What I verified: the arithmetic for the test's own input, both call sites, and the three consumers, all by reading `main` at `a8125ad`. What I did not: I have not run the suite or the app against this branch.

Ashfaqbs:
Good catch, thanks for the thorough walkthrough — you're right that normalizing west/east back into range re-inverts the box across the antimeridian and breaks the `west <= east` assumption downstream.

Went with your first suggestion in 830c372: dropped `normalizeLonDeg` entirely, kept `longitudeSpanDeg` for the wrapped span calculation, and left `west`/`east` unnormalized so the box stays monotonic. Also updated the antimeridian test to assert `west <= east` and that the plain-mean center lands back near the requested center, instead of just checking the values are in `[-180, 180)`.

Agreed on leaving the "coordinates outside the conventional range at the Overpass boundary" as a separate, visible follow-up rather than folding it into this PR.

samehkhamis:

PR #399, opened 2026-09-13
Add support for DriveBC CCTV this covers all of British Columbia

infamy:
Adds support for drivebc cctv.
Limits to 250, otherwise BC alone blows the 1200 cap.


bilawalsidhu:

We'll carry the per-camera credit field through, since a few of the feeds are TransLink or city cameras rather than provincial.

bilawalsidhu:

PR #408, opened 2026-09-13
fix: correct CCTV panel reset key

vegettto:
## Summary
- Update the CCTV panel recovery command to target the current v8 position key.
- Keep the known-issues and startup guidance aligned with the runtime storage version.
- Add a regression test that derives the storage version from ui.js.

## Testing
- `node --test src/tooling/panelStorageDocs.test.mjs`
- `node --check src/tooling/panelStorageDocs.test.mjs`


bilawalsidhu:
Landed via #509 (rebased onto the current component layout, with fixes from review). Thank you @vegettto!

PR #312, opened 2026-09-11
i18n: EN + ES + FR + RU + UK interface localization with configurable locale pair

meirm:
# i18n: English + Spanish + French + Russian + Ukrainian interface localization, with a configurable locale pair

## Summary

The entire application-owned interface is localizable and ships fully translated in **English, Spanish, French, Russian, and Ukrainian** — with the language pair configurable via environment. All machine-facing contracts (layer IDs, status enums, share-link schemas, API fields, voice tool schemas, provider names, attribution) are byte-for-byte unchanged, and the default English experience is byte-identical to `main`.

- **~916 catalog keys per locale** across four namespaces (`shell`, `cockpit`, `layers`, `setup`), mirrored key-for-key across **all five locales** under strict CI parity — a missing translation in any locale fails the build.
- **Full plural support**: `Intl.PluralRules`-driven selection; en/es/fr use `one/other` (plus locale-valid additions), ru/uk ship grammatically complete **`one/few/many/other`** sets (Очищен 1 слой / 2 слоя / 5 слоёв / 1.5 слоя). The parity gate enforces Intl-valid variant supersets per locale.
- **Locale resolution:** `?lang=` override → stored preference (`gev:locale:v1`) → browser language (regional variants normalize: `es-MX`→`es`, `fr-CA`→`fr`, `ru_KZ`→`ru`, `uk-UA`→`uk`) → configured default.
- **Configurable pair** via `.env`: `GEV_DEFAULT_LOCALE` (default `en`) / `GEV_SECONDARY_LOCALE` (default `es`). `en` always ships as the source-of-truth fallback; invalid/degenerate pairs fall back to `en`+`es` with a dev-only warning. Browser-language detection keeps priority over the configured default. Combos like `ru`+`uk` render a three-way selector (RU | UK | EN).
- **Dock selector** renders the configured pair dynamically with native-name accessible labels; switching preserves the share-link hash exactly and never leaks `?lang` into copied links.
- **Zero new dependencies** — plain JS modules, `Intl`, `localStorage`, `URL`.

## What changes

| Area | Detail |
|---|---|
| `src/i18n/` | `locale.js` (normalization for 5 locales, pair resolution, guarded storage, `<html lang/dir>`), `index.js` (`t()` with `{named}` interpolation + Intl plural categories, Intl formatters, `applyDocumentTranslations()` for `data-i18n*` markup, hash-preserving reload) |
[...]

PR #412, opened 2026-09-13
Feature: Multi Language UI Support - translation

HadiMuhammed:
Added Multi-Language Support for UI

The interface supports English, Spanish, French, German, Japanese, and Arabic. English is the default, and the selected language is remembered in the browser. Stable interface text, data-layer names, scene titles, and accessibility labels use the local translation catalog. Dynamic provider text can optionally use the server-side hybrid translation proxy with an English fallback.

Language Change Button can be seen after the "Full Globe View Button"


bilawalsidhu:
Thanks for the contribution. Closing this automatic page-translation implementation as not planned. Source review found first-observed text retained for reused dynamic nodes and no post-await locale/disposal guard before translated text is written. Static catalogs and accessibility/RTL ideas are preserved for review in #312. A deliberate locale architecture remains welcome; this does not approve #312 unchanged.

Issue #270, opened 2026-09-10
.env.example documents two CCTV env vars (CCTV_AUTO_CALIBRATE, CCTV_DRAPE_MESH) that no code reads


jsawyerdev:
## Summary
`.env.example` documents two CCTV-related environment variables that no code in the repository ever reads.

## Evidence
`.env.example:141-142`:
```
# CCTV_AUTO_CALIBRATE=1                                # auto-calibrate camera projection (client flag)
# CCTV_DRAPE_MESH=1                                     # drape frames onto 3D geometry (client flag)
```
A repo-wide search for `CCTV_AUTO_CALIBRATE` and `CCTV_DRAPE_MESH` (including case variants) across `src/**/*.js`, `src/**/*.mjs`, `vite.config.js`, and every `*.json`/`*.sh` file returns zero matches outside `.env.example` itself.

## Why this matters
A contributor following CONTRIBUTING.md's "Add a CCTV source pack" guidance who sets either variable to tune camera calibration/draping will see no effect and get no error or warning — the documented knob simply does nothing. This is a classic doc-drift bug: either the feature was removed without updating `.env.example`, or it was planned but never implemented.

## Attack or failure scenario
Not a security issue — a documentation/UX correctness bug that costs a contributor debugging time when the variable silently has no effect.

## Root cause
`.env.example` and the actual CCTV layer implementation diverged at some point (feature removed or never built) without the example file being updated.

## Recommended fix
Either implement the two flags as documented, or remove them from `.env.example`.

## Acceptance criteria
- `.env.example` no longer documents environment variables that have no corresponding code path, or the code path is added.

## Suggested labels
documentation, bug

## Priority
P3

## Severity
Low — doc-drift with no functional or security impact beyond wasted contributor time.

## Confidence
Confirmed — grepped the entire working tree for both variable names outside `.env.example`.

Lob26:
Verified against `main` at `a8125ad`: this looks already fixed.

Neither variable appears anywhere in the tree — not in code and not in `.env.example`:

```
$ grep -rn "CCTV_AUTO_CALIBRATE\|CCTV_DRAPE_MESH" --include=*.js --include=*.mjs --include=*.sh --include=*.example .
(no matches)
```

`.env.example` still documents the other ~30 `CCTV_*` knobs (lines 145-176), so the section is intact — just these two are gone. Closeable unless you read the issue as also asking for something broader.

bilawalsidhu:
[...]

Issue #209, opened 2026-09-09
open camera intergrations


s-b-repo:
http://www.insecam.org/
https://www.earthcam.com/
https://www.i-traffic.co.za/cctv?start=0&length=10&order%5Bi%5D=0&order%5Bdir%5D=desc
https://www.skylinewebcams.com/en/webcam.html
https://www.outdooractive.com/en/webcams/city-of-johannesburg/webcams-in-city-of-johannesburg/227986467/
https://henleyair.tv/
https://trafficvision.live/blog/gauteng-traffic-cameras
https://www.sanparks.org/travel/webcams/overview
https://www.earthcamtv.com/
https://aviationwebcams.co.za/
https://www.cameraftp.com/cameraftp/publish/publishedcameras.aspx

bilawalsidhu:
Keeping this as a source-discovery request for now. Please select one provider and provide an official endpoint/schema, redistribution or embedding terms, attribution, rate limits, and a small set of stable camera IDs. A public website listing alone is not enough to implement a supported source pack.

s-b-repo:
i was think scraping cameras then we could add them from a cdn or once or make it so github constanlty has a updated list ips are public and so are the camers we arent hacking them so we could list them and set up a github worker which updates a hard coded list

bilawalsidhu:
Closing. Directories of unsecured private cameras aren't a source we'll use. Individual public camera sources with clear terms are welcome as CCTV pack proposals.

PR #415, opened 2026-09-13
feat(cctv): add Finland — Fintraffic road weather cameras as a fourth source pack

### LINKED ISSUE #269: No documented policy on face/plate content in live-relayed CCTV frames
## Summary
Live CCTV frames relayed through the app (Austin, Caltrans, TfL JamCams, and community-contributed source packs) may contain identifiable faces or license plates, and neither DATA_SOURCES.md nor the CCTV-source-pack contribution guidance in CONTRIBUTING.md addresses blurring, redaction, or a retention statement for this content.

## Evidence
`DATA_SOURCES.md:34-36` lists the CCTV data sources (Austin Open Data, Caltrans, TfL JamCams) with license/attribution terms but no mention of face/plate handling. README.md's "Responsible & Open" section states the project "does not build features for named-person search, face recognition, or tracking individuals" — a real and meaningful commitment about *features*, but it doesn't address the separate question of what raw camera frames contain when relayed as-is. CONTRIBUTING.md's "Add a CCTV source pack" guidance (the recommended first contribution) discusses coordinates, attribution, and server-registered frame URLs, again with no mention of frame content.

## Why this matters
This mirrors the upstream providers' own public camera feeds (the frames aren't processed or enhanced beyond relaying), so it isn't a new technical risk introduced by this project — frames are relayed live with `Cache-Control: no-store` and are not persisted to disk. But for a project explicitly positioned as an "OSINT"/"spatial intelligence" tool with a stated ethical line around individual tracking, an explicit written statement of the project's position on raw frame content (mirrors upstream, no blurring applied, not used for identification) would close an easy-to-raise gap between the stated ethical line and what a camera frame can actually show.

## Attack or failure scenario
Not an exploitable vulnerability — a documentation/policy gap. The risk is reputational/policy ambiguity rather than a technical failure mode.

## Root cause
DATA_SOURCES.md and CONTRIBUTING.md document licensing and technical integration for CCTV sources but were never extended to cover frame-content policy.

## Recommended fix
Add a short, explicit statement to DATA_SOURCES.md (or SECURITY.md/README's "Responsible & Open" section) clarifying that camera frames are relayed as-is from public sources with no blurring/enhancement, are not persisted, and are not used for identification — consistent with the existing "no named-person search" commitment.

## Acceptance criteria
- DATA_SOURCES.md or an equivalently discoverable doc states the project's position on CCTV frame content (blurring, retention, identification use).

## Suggested labels
documentation, privacy (privacy label doesn't exist on this repo; `documentation` is available)

## Priority
P3

## Severity
Low — documentation gap; no evidence of a technical difference from the upstream sources' own public exposure.

## Confidence
Suspected — this is primarily a documentation completeness judgment call, not a confirmed technical defect; the maintainers may consider the existing README language sufficient.

avocadoAgent:
Austin, Caltrans and TfL cover two countries. This adds a fourth keyless pack: Fintraffic's road weather cameras, covering all of Finland, via Digitraffic's open API. Related to #209 (open camera integrations).

### What changed

- **New source pack** (`server/providers/cctv/sources.js`). One keyless GeoJSON request to `tie.digitraffic.fi/api/weathercam/v1/stations` — ~810 stations, of which ~808 are `GATHERING` and carry ~2,260 in-collection presets (measured 2026-09-13; the catalog grows). A station's *presets* are its fixed views and share its position, so one preset is one camera; 300 are kept, prioritized against seven anchors along Finland's main road spine so the cap keeps national coverage rather than just Helsinki. Stations that aren't `GATHERING` and presets that aren't `inCollection` are dropped, so the mesh carries no dead cameras.
- **Frame URLs are built, not read.** `FINTRAFFIC_IMAGE_ORIGIN + presetId + '.jpg'`, with the preset id gated on `/^C\d{7}$/` and required to start with its station id. No upstream field can steer the proxy off-host or escape the path. The catalog fetch refuses redirects and sends the `Digitraffic-User` header the service asks clients to identify themselves with, plus gzip.
- **No heading exists in this dataset**, so presets take the id-hash fallback and the low-confidence pose personality, exactly like headingless Austin and TfL cameras. (The per-preset `direction` on Digitraffic's detail endpoint is road-register relative — `INCREASING_DIRECTION` means "towards higher road addresses" — not a bearing, and it isn't in the bulk list anyway.) Because the hash keys on the preset id, a station's co-located presets fan apart instead of stacking on one bearing.
- **Road-surface views get a steep pose.** About a quarter of Finnish presets are "Tienpinta" views: a camera a metre or two up staring at the asphalt. Given the road pose they aim at the horizon. They now get pitch -55° / range 220 m / mount 6 m — each value the boundary `src/data/cctv.js` already clamps registered poses to, so the prior is as close to a downward-staring camera as the renderer represents and nothing is silently rewritten. Identifying them from the preset id is an **observed convention, not a documented contract**, and the constant says so with the evidence: 575/575 presets over a systematic 202-station national sample have `directionCode` equal to their id suffix, and 140/140 suffix-09 presets are named "Tienpinta" — but a separate sample found one counterexample (`C1853409`), so it's ~99%. It steers a draggable pose prior, never which cameras exist, so a rare miss costs one gizmo nudge.
- **10-minute frame cadence** (`cctvLod.js`). Digitraffic publishes one JPEG per station collection interval — 600 s on every station sampled, borne out by the images' `Last-Modified` headers. The default 5-minute cadence would fetch each image twice per update.
- **Attribution.** CC BY 4.0, attribution required: registered in `DATA_CREDITS` so it surfaces in the Data attribution popover, plus the `DATA_SOURCES.md` licence row and source note.
- **Docs.** `CHANGELOG.md`, `docs/CURRENT-STATE.md`, `README.md` (which said Austin/California/London and "~800 cameras"), and `.env.example` for the two new env vars — both of which are actually read, `CCTV_FINTRAFFIC_ENABLED` in `catalog.js` and `CCTV_FINTRAFFIC_MAX_SOURCES` in `sources.js` (cf. #270). `CCTV_FINTRAFFIC_ENABLED=0` is the kill switch; the pack fails independently of the other three.

### One trade-off to decide on: the default source cap (`87d6ae8`)

That commit raises `DEFAULT_CCTV_MAX_SOURCES` 900 → 1,200, and it's isolated so you can drop it on its own.

The merge truncates by position (`mergedSources.slice(0, maxCount)`). Austin + Caltrans + TfL already fill 800 at their default caps, so a fourth pack appended after them is cut to 100 — and to nothing the moment an earlier pack's cap is raised. 1,200 isn't a new ceiling: it's the hard bound `getCctvSources` already clamps `CCTV_MAX_SOURCES` to, and the size `HEALTH_MAX_ENTRIES` was chosen to cover. It holds all four packs at their defaults (250+300+250+300 = 1,100) with headroom for file/env entries.

This is the only change here that shifts a runtime default for every user rather than just for the new pack. Drop the commit and everything still works — Finland is just truncated. Happy to go the other way instead (a smaller Fintraffic cap that fits under 900) if you'd rather not move the default.

### On #269 (face/plate content)

These are road **weather** cameras: highway and trunk-road views published by the national traffic agency under CC BY 4.0, refreshed on a 10-minute cadence, with a quarter of them pointed at the road surface. Nothing is cached or rehosted — frames are proxied live from `weathercam.digitraffic.fi`, same as the existing packs. A 10-minute stills cadence is a poor instrument for following anyone, and the pack adds no identity surface. It doesn't resolve #269, which asks for a stated policy rather than a per-pack judgement, but it doesn't add to the gap either.

### How I verified it

- `npm ci`, `npm run doctor -- --json`, `npm run format:check` (257 files), `npm run check:boundaries`, `npm test` (**3191 tests, 3190 pass, 0 fail**, 1 pre-existing runtime-gated skip), `npm run build` — all green on Node 26.8.2, on the branch rebased onto `cf149ec`. `package-lock.json` untouched.
- **`npm run test:track` I could not run** — my machine has no browser-capable environment (Chrome installs but is missing 20 system libraries). Flagging it honestly rather than claiming it; happy to act on anything it reports.
- 9 new unit tests (`src/data/cctvFintraffic.test.mjs`) covering preset→camera mapping, the official-origin frame URL, filtering of dead stations / uncollected presets / malformed ids, elevation handling, the Digitraffic header and redirect refusal, the cap and fail-soft paths, and the two poses.
- Live against a dev server: `/api/cctv/sources` returns 1,100 cameras (250/300/250/300), `/api/cctv/frame/…` returns `image/jpeg` with `X-CCTV-Source: upstream-image`, surface presets carry exactly one pose and road presets another, and `CCTV_FINTRAFFIC_ENABLED=0` drops the total back to 800.
- Station/preset counts, the 600 s cadence, the elevation semantics and the preset-id convention were all measured against the live API on 2026-09-13, not taken from docs.

A screenshot of Finland in the mesh follows in a comment.


bilawalsidhu:

bilawalsidhu:

PR #420, opened 2026-09-13
Fix Mapped Installations: permanent "unavailable" and silently dropped features

GladiatorrX9:
Mapped Installations never loaded for me, and it turned out to be three independent defects. Any one of them alone was enough to hide most or all of the layer, which is why it never recovered on its own.

Each is fixed separately below, with the evidence that identified it.

## 1. A 4xx from one mirror ended the whole mirror chain

`fetchOverpassPayload` fell through to the next mirror only on rate-limits, runtime errors, and 5xx. A **4xx was treated as an authoritative answer and returned**, so the first mirror spoke for every mirror below it.

Since `/api/military-installations` treats any `status >= 400` as a failure, that surfaced as a 503 → `"temporarily unavailable"`, and the 30s→240s retry re-hit the same dead mirror forever.

4xx is now a skip. The most informative refusal is still remembered, so if *every* mirror declines (e.g. a genuinely malformed query 400ing everywhere) the error text survives rather than being flattened into "all upstreams failed". The decision is extracted as a pure exported `overpassMirrorDisposition`, so the policy is testable without a server — matching the existing `resolveOverpassPreflight` convention.

## 2. The project's User-Agent is blocklisted by overpass-api.de

This is the part worth a maintainer's attention.

| User-Agent sent | `GET /api/status` |
|---|---|
| `gods-eye-view-overpass-proxy/1.0` (current) | **406** |
| `gods-eye-view-overpass-proxy/9.9` | **406** |
| `overpass-proxy/1.0` | 200 |
| `foo-overpass-proxy/1.0`, `myapp-overpass-proxy/2.3` | 200 |
| `curl/8.4.0`, browser UA | 200 |

Generic `overpass-proxy` strings pass from the same IP, so this is a **targeted rule against this project's exact string**, not a pattern filter or an IP ban. It 406s every request including `/api/status` with no query at all.

Combined with defect 1, that single 406 from the first mirror took the feed down permanently for anyone running this.

Replaced with a string that identifies the app and carries a contact route, which is what the [OSM API usage policy](https://operations.osmfoundation.org/policies/api/) asks for:

```
gods-eye-view/0.1 (+https://github.com/bilawalsidhu/gods-eye-view)
```

I've left a comment noting that if this is ever blocked again, the signal is to cut query volume — not to rename the client. **Happy to drop this commit if you'd rather contact the operator about the existing string**; the other two fixes stand on their own.

Also added one verified full-planet fallback mirror. A health sweep found the list had no fast survivor when the `.de` cluster is unhappy: kumi failed the real installation query in every measurement (504 at 55s, then a 22s abort) and private.coffee did not answer inside 180s. Cold latency went ~35s → ~6s.

## 3. Ways and relations were dropped for having no location

The query asked for `out center tags geom`, but **Overpass geometry modes are exclusive and `geom` wins**, so `center` was never in the reply. Every way and relation arrived carrying `bounds`/`geometry` and no centre point, and the normalizer discarded all of them.

Austin mapped **six** features and drew **one** — the lone tagged node. Camp Mabry, Texas Guard Armory, and the Texas National Guard sites were invisible.

Features are now placed from `bounds` (the same box `out center` derives from), falling back to `geometry` and to relation member geometry. The query no longer asks for a mode it cannot receive.

Worth noting: every existing normalizer fixture used `center:`, which real payloads never contain — which is why this never showed up in tests. The new cases use the verified live shape.

## Verification

Against live upstreams, not mocks:

| | before | after |
|---|---|---|
| Endpoint | 503 `unavailable` | 200 `ready` |
| Austin | 1 record | **6 records** |
| Tokyo | — | **10 records** |
| Cold latency | never succeeded | ~6s |
| Warm (cache hit) | — | 3ms |

Test suite: **2567 pass**. There are 25 failures on this branch, but they are **pre-existing on `main`** — `applyVisualState is missing from ui.js` and related cockpit tests — and I confirmed that by stashing: `main` alone gives the identical 25. None touch installations or Overpass. This branch adds 5 tests, all passing, and changes no other behaviour.

https://claude.ai/code/session_012mp9pXE78qJXHZTW9CemSS


bilawalsidhu:
Landed via #570 (rebased onto the current component layout, with fixes from review). Thank you @GladiatorrX9!

PR #425, opened 2026-09-13
CCTV: add Finland, Ontario, British Columbia (keyless) + Washington, NSW, Singapore (opt-in) — 800 to 1556 cameras

GladiatorrX9:
Takes the CCTV catalog from **800 cameras in 3 regions to 1556 in 8**, across six new sources. Three need no key or signup at all.

> **Stacks on #420.** The first commit here is that PR's — these loaders use the `OVERPASS_USER_AGENT` constant it introduces. Review #420 first; if it merges, this rebases to four commits. Happy to split further if you'd rather take these one source at a time.

## What's added

| Source | Available upstream | Loaded | Key |
|---|---|---|---|
| 🇫🇮 Fintraffic Digitraffic | 811 stations / **2277 presets** | 150 | **none** |
| 🇨🇦 Ontario 511 | 944 | 150 | **none** |
| 🇨🇦 DriveBC | 1062 | 150 | **none** |
| 🇺🇸 WSDOT | 1705 | 150 | free, opt-in |
| 🇦🇺 Live Traffic NSW | 217 | 150 | free, opt-in |
| 🇸🇬 LTA DataMall | 8 | 8 | free, opt-in |

Every pack is switchable (`CCTV_<PACK>_ENABLED=0`) and sized (`CCTV_<PACK>_MAX_SOURCES`, where `0` also disables). **The keyed ones return `[]` before any network call when their key is absent**, so a keyless clone is byte-for-byte what it is today — verified: 1250 keyless, 1556 with all three keys, and none of the keyed providers present in the keyless run.

Heading quality across the catalog improves from 273 high / 157 medium to **273 high / 317 medium / 966 low**, because DriveBC, WSDOT and NSW all publish real bearings.

## Findings worth a maintainer's attention

Each of these would have shipped a silently broken pack. They're recorded as comments at the code, not just here.

**NSW's CDN fakes success.** It serves frames only to browser-shaped clients; to anything else — including this repo's own `gods-eye-view-cctv-proxy/1.0` — it returns **HTTP 200 with a 307-byte HTML "Page not found" body**. `resp.ok` alone would accept that. The existing `content-type.startsWith('image/')` guard in `fetchCctvImageFromUpstream` already converts it to an honest miss, but the outcome would have been every NSW camera quietly falling through to Street View with nothing logged. That function now takes an optional per-source `userAgent`, declared by the pack as `imageUserAgent` and deliberately **not** exposed through `/api/cctv/sources`.

**WSDOT's `Direction` is mostly not a bearing.** Of 1705 cameras, **1182 are `"B"`** (both directions) and 13 are `"O"`. Mapping those through a compass table would have aimed two-thirds of Washington due north *with `medium` confidence attached*. Only N/S/E/W are trusted.

**Digitraffic 406s without `Accept-Encoding: gzip`.** It's mandatory on their JSON interfaces. A 406 here is indistinguishable at a glance from a User-Agent block (cf. #420), so it's called out in a comment. The requested `Digitraffic-User` header is also sent — it raises the rate limit.

**Digitraffic's bulk list omits `imageUrl`.** Presets arrive as `{id, inCollection}`; `imageUrl` and `direction` exist only on the per-station detail endpoint, and 811 of those against 60 req/min is ~14 minutes — not viable at startup. The loader uses the documented URL pattern and one call, accepting id-hash headings as the cost.

**Two opposite elevation semantics, one line apart.** Digitraffic's third GeoJSON coordinate is a placeholder `0.000000` on every station, so `|| 30` is correct there — `0` means "not published". DriveBC publishes real elevations, so `??` must keep a genuine `0`. Both directions are pinned by tests so the next reader doesn't "fix" one into the other.

**Ontario describes views in prose.** `"Toronto Bound"` says where the *road* goes, not where the camera looks, so only unambiguous compass words are promoted to a heading; everything else stays `low`. Longest-first matching keeps `"northeast"` from degrading to `"north"`. Its frames are PNG, not JPEG — handled by the existing pass-through content-type.

**NSW uses hyphenated codes** (`"N-E"`), de-hyphenated before lookup; otherwise 60 diagonal cameras degrade silently.

**WSDOT documents `http://` endpoints.** The loader forces `https`, verified to serve the full catalog, so the proxy never makes a cleartext upstream call on the operator's behalf.

## Global cap

Raised 900 → 1700. At 900 the three existing packs already fill 800, and since new packs are appended last, the global `slice()` would have truncated **them** rather than Austin or TfL — per-pack caps would have looked respected while the new cameras quietly vanished. Lower it or any `CCTV_*_MAX_SOURCES` if the catalog size isn't worth it on your hardware.

## Also included

A **Kuala Lumpur pilot pack** (`config/cctv_sources.kualalumpur.json`) plus `config/README.md`, documenting the previously-undocumented source-pack mechanism — including that `CCTV_SOURCES_FILE` silently disables the live packs unless `CCTV_FORCE_AUSTIN=1` is set. It's explicitly **not** a live feed (`sourceKind: "pilot"`); KL's operators publish no coordinates, so there is nothing to connect. Its positions are real OSM traffic-signal nodes and its elevations come from this repo's own terrain proxy. **Drop this commit if you'd rather not carry a demo pack** — the live sources don't depend on it.

## Verification

Live, not mocked. Every pack serves `X-CCTV-Source=upstream-image` with health `ok`/`snapshot`: Digitraffic Helsinki Ruoholahti, Ontario Lake Shore Blvd at Yonge, DriveBC Georgia St Vancouver, WSDOT I-5 at 5th Ave/Columbia Seattle, NSW Erskine St Sydney, LTA Tuas Checkpoint.

**29 new tests** pinning each upstream's field contract, so a rename fails loudly rather than yielding an empty pack — the exact failure mode #420 exists to fix.

Suite **2591 pass**. The 24–25 failures are pre-existing `ui.js` cockpit tests (`applyVisualState is missing from ui.js`); I diffed the failing set against a stash of these changes and it is **identical**, so nothing here regressed. One of them is flaky, hence the 24/25 wobble.

https://claude.ai/code/session_012mp9pXE78qJXHZTW9CemSS


bilawalsidhu:

Only the New South Wales pack is being taken from this one, as an opt-in source that needs a Transport for NSW key. Ontario, BC and Finland come from the PRs that were already on the new layout.

bilawalsidhu:

PR #430, opened 2026-09-13
fix(layers): release bundled GeoJSON entities on toggle-off instead of hiding them

bilawalsidhu:
## Problem

Owner field test (2026-09-13, main @ 3143d00): turn on **Data Centers + Submarine Cables + Dams**, turn them all off again, go back to flights/military/satellites. Everything is noticeably more sluggish for the rest of the session even though the three layers are "off".

Root cause: the three bundled layers (`localGeojsonCore.js` for datacenters/dams, `telegeographySubmarineCables.js`) implement `disable()` as `dataSource.show = false`. Cesium's `DataSourceDisplay.update` still walks every visualizer of a *hidden* data source on every frame, and the ~11.5k entities (polylines, billboards, stems, pick records) stay resident. The per-frame walk plus the GC pressure from hundreds of MB of parked entities inflates every frame's cost, not just the geojson share.

## Fix

- `disable()` now **removes the data sources from the viewer** and drops the built records (stems, pick maps, entity contexts, sweep gate state) instead of hiding them.
- The **parsed dataset is cached** (NDJSON features for datacenters/dams; cable + landing-point JSON for cables), so a re-enable rebuilds the entities without a refetch or re-parse. `destroy()` clears the cache.
- A `disable()` that lands while a load is still in flight releases the finished build instead of parking it hidden (the existing generation/ownership pattern in the cables layer is untouched; the stale generation still removes exactly what it added).
- Cables: `_markerBlendDone` resets on release so the rebuilt marker collections get their own translucent-blend pass.

No API or UI changes. `getStats().count` still reports the dataset size while off, as before.

## Measurements

Headful Chrome (real GPU) driven by Puppeteer against the dev server; flights + military + satellites on, camera rotating continuously so the render governor is out of idle; each row is a 250-frame window. Per-frame timings are wrapped `scene.render` / `dataSourceDisplay.update`.

**Before (main), toggling the three layers on then off:**

| window | DataSourceDisplay.update | scene.render | resident entities | JS heap |
|---|---|---|---|---|
| fresh | 0.02 ms | 10.0 ms | 0 | 144 MB |
| geojson ON | 2.75 ms | 14.2 ms | 11,537 | 578 MB |
| geojson OFF (hidden) | 1.09 ms | 13.9 ms | 11,537 | 857 MB |
| OFF +30 s | 0.99 ms | 11.4 ms | 11,537 | 1,603 MB |
| manual `entities.removeAll()` | 0.03 ms | 9.3 ms | 0 | 699 MB |

Control run (same elapsed time, never touching geojson): 5.6 → 6.6 → 5.8 ms render, heap flat at ~150-180 MB. So after OFF the scene is roughly 2× the control's cost, and the `removeAll()` row is what pointed at hidden entities as the cause.

**After (this branch), same script:**

| window | DataSourceDisplay.update | scene.render | resident entities | JS heap |
|---|---|---|---|---|
| fresh | 0.02 ms | 5.0 ms | 0 | 128 MB |
| geojson ON | 1.85 ms | 6.1 ms | 11,537 | 492 MB |
| geojson OFF | 0.02 ms | 5.2 ms | 0 | 740 MB (pre-GC) |
| OFF +30 s | 0.02 ms | 4.8 ms | 0 | 342 MB |

Entities go to zero on OFF, the data-source walk disappears, render returns to the fresh baseline, and the heap comes back down once GC runs. (Absolute render numbers differ between the two sessions by machine load; the within-run deltas are the signal.)

Re-enable after OFF rebuilds from the cache: no network request, no re-parse.

## Tests

- `src/data/localGeojsonLifecycle.test.mjs`: disable now asserts zero data sources and zero entity contexts; the disable-during-load case asserts the completed build is released and re-enable rebuilds with `fetch` called once; new test for the toggle-off/toggle-on cycle (sources 0 → 1, fetch once, count stable).
- `src/data/telegeographySubmarineCables.test.mjs`: the post-cable-GeoJson abort race updated for the cached-JSON path (load B skips the fetch, its own cable build is the only new GeoJson call, stale A still never builds landings); new test asserting a toggle-off removes all three sources and a toggle-on rebuilds them with `fetch` stubbed to throw.

Gates on the branch: `npm test` 3235/3237 (the two failures are the pre-existing macOS tmpdir cases in `devCctv.test.mjs` and `previewServing.test.mjs`, which also fail on untouched main), `npm run format:check` clean, `npm run build` clean, `npm run check:boundaries` clean.

## Reviewer notes

- The cables layer's `releaseDataSources()` is also what `destroy()` now uses, so the three-remove block there collapsed into one call.
- Re-enable does rebuild ~11.5k entities each time (a few hundred ms of `GeoJsonDataSource.load` on this machine) rather than being instant. That is the trade: a toggle-off that actually frees the scene versus a toggle-on that is slightly slower than the old un-hide. Happy to gate the rebuild behind a short grace timer if you'd rather keep quick flips cheap.


samehkhamis:
Thank you, Bilawal! Merged with your original commit preserved.

Validation on the merged candidate: 3,258 tests passed (one existing skip), doctor, formatting, package boundaries, build, all three CI jobs, 109 tracking checks and eight cable overlay checks. Three browser off/on cycles each released all 11,537 entities and rebuilt without refetching the datasets. The 23 infrastructure checks passed with OSM selected; the initial detailed-basemap run stalled under software rendering and was stopped. Two settled views were inspected. These checks confirm cleanup and rebuild behavior, not an independent reproduction of the real-GPU timing or heap measurements.

PR #438, opened 2026-09-14
feat(cctv): add the Open Calgary camera pack

rileygramlich:
## What

Adds Calgary as a fourth CCTV city pack alongside Austin, Caltrans and TfL — about 214 public
traffic cameras from Open Calgary, keyless, served through the existing frame proxy. This is the
"add a CCTV source pack" invitation in CONTRIBUTING.

I live in Calgary, which is mostly why I picked it: it's the city I can actually sanity-check
camera by camera, and it turned out to matter for the heading problem below.

## The one thing worth reviewing carefully

Calgary's dataset looks like it tells you which way each camera points. It doesn't.

Every record has a `quadrant` field (`NE`/`NW`/`SE`/`SW`), and `camera_location` ends in the same
token — "9 Avenue / 3 Street SE". If you live here you read that immediately as the address grid:
the city is split into four quadrants and nearly every street name carries one. It says nothing
about where the camera is aimed.

The trap is that `directionToHeading()` parses it quite happily. Feed it "9 Avenue / 3 Street SE"
and you get 135°, confidently, for all 214 cameras — and all 214 are wrong. That's the same
failure `src/data/directionText.js` already documents for Austin street names containing cardinal
words (59 of ~1000 there). Calgary just hits it on every single row instead of a few.

So this pack derives no heading at all. Cameras get the shared id-hash fallback at
`headingConfidence: 'low'`, same as TfL JamCams, and you correct them by dragging the gizmo.
There's a test pinning that behaviour so nobody "fixes" it later by wiring up the quadrant.

## Smaller details

- **Frame URLs.** The catalog publishes `http://trafficcam.calgary.ca/...`, but the host serves
  HTTPS and 301s to it, so I upgrade the scheme before registering and pin to that exact origin.
  Same idea as the Caltrans and TfL host pins.
- **Dataset id** came from the Socrata catalog API rather than being hardcoded from somewhere —
  `k7p9-kppz`, "Traffic Cameras".
- **Parsing** lives in `src/data/calgaryCameras.js` with 12 unit tests, following how
  `firmsCsv.js` and `directionText.js` keep server-side parsing testable outside Vite.
- Gated behind `CCTV_CALGARY_ENABLED` (default on) with `CCTV_CALGARY_MAX_SOURCES` to cap it.

## One change outside the pack, which I'd rather flag than bury

I raised `DEFAULT_CCTV_MAX_SOURCES` from 900 to 1100.

Four packs at their default caps come to 1014 (250 + 300 + 250 + 214). The global cap keeps the
first N after the merge and Calgary merges last, so at 900 it registered 100 of 214 and logged the
"exceeds cap" warning. Raising the ceiling looked like the smaller change — capping Calgary at
under half seemed worse than growing the catalog by about 11% — but it does affect everyone, not
just people looking at Alberta. If you'd rather hold 900 and cap Calgary instead, say so and I'll
switch it.

## Licensing

Open Government Licence – City of Calgary, which requires attribution. Registered in
`src/data/dataCredits.js` and documented in `DATA_SOURCES.md`:
"Contains information licensed under the Open Government Licence – City of Calgary."
No city logo is used.

## How I verified

On Node 24.14.0 and 26.8.2:

- `npm ci` clean
- `npm test` green on both (allocation gate enforced on 24.14.0 via `GEV_REQUIRE_ALLOCATION_GATE=1`)
- `npm run build` green
- `npm run test:track` 108 passed, 0 failed
- Live against the dev server: 1015 cameras registered across the four packs, no cap warning, and
  `/api/cctv/frame/calgary-86` and `calgary-142` both return `image/jpeg`

Happy to add screenshots of the mesh over Calgary with a live frame in the monitor panel if
that's useful for review.


bilawalsidhu:
Landed via #514 (rebased onto the current component layout, with fixes from review). Thank you @rileygramlich!

PR #359, opened 2026-09-12
feat: add seven Brazilian Bike Itaú cities to the GBFS bikeshare layer

MiguelGFerreira:
## Summary

Adds São Paulo, Rio de Janeiro, Salvador, Recife, Porto Alegre, Brasília and Curitiba to `RAW_GBFS_CITY_REGISTRY`. These are the bikeshare layer's first stations outside the United States — 1,017 stations across the seven systems.

All seven are Tembici systems running on the same PBSC platform (`*.publicbikesystem.net`) as the existing Austin, Honolulu and Chattanooga entries, so this needs no new feed handling:

- No proxy allowlist change — `isAllowedGbfsHost` already accepts any `*.publicbikesystem.net` subdomain via its wildcard.
- No new helper or factory — the entries are written longhand, matching the other PBSC entries in the registry.
- Feed URLs use the `/customer/gbfs/v2/en/station_{information,status}.json` shape that the registry validator and `isAllowedGbfsPath` both require.

| City | Stations | Host |
|------|----------|------|
| São Paulo, SP | 240 | `saopaulo.publicbikesystem.net` |
| Rio de Janeiro, RJ | 403 | `riodejaneiro.publicbikesystem.net` |
| Salvador, BA | 76 | `salvador.publicbikesystem.net` |
| Recife, PE | 90 | `recife.publicbikesystem.net` |
| Porto Alegre, RS | 90 | `portoalegre.publicbikesystem.net` |
| Brasília, DF | 68 | `brasilia.publicbikesystem.net` |
| Curitiba, PR | 50 | `curitiba.publicbikesystem.net` |

Attribution for Tembici / Bike Itaú is added to `DATA_SOURCES.md` and `DATA_CREDITS` (`src/data/dataCredits.js`), and the change is recorded in `CHANGELOG.md`.

## Test plan

- [x] `npm run build` — green
- [x] `npm test` — 2967 tests, 0 failures
- [x] `npm run test:track` — 108 passed, 0 failed
- [x] `npm run format:check` and `npm run check:boundaries` — green
- [x] All 14 feed endpoints (`station_information` + `station_status`) return HTTP 200 with live station data
- [x] Verified end-to-end through the app's own `/api/gbfs` proxy — the wildcard allowlist accepts the Brazilian hosts without modification
- [x] Loaded the layer in the browser and flew to São Paulo: stations render with live names and availability (Largo de São Bento [15/15], Consolação [16/19], Praça do Ciclista [8/28], Av. São João [9/11])

[...]

PR #441, opened 2026-09-14
fix bikeshare layer: address the GBFS proxy by path, not query string

MiguelGFerreira:
## Summary

The bikeshare layer currently fails for every city on `main`. Every station request answers `400 Bad Request` and the layer reports `UNAVAILABLE — GBFS fetch error`.

The extracted station source asks the proxy for `/api/gbfs?url=<encoded-upstream>`, but the proxy reads its upstream target from the **path**, not the query string:

- `server/providers/gbfs.js` — `url.pathname.replace(/^\/+/, '')`
- `src/layers/bikeshare/ingestion.js:17` — builds `/api/gbfs/${encodeURIComponent(upstreamUrl)}`
- `src/layers/bikeshare/source.js:12` — builds `/api/gbfs?url=${...}` ← the outlier

Two of the three already agree on the path shape. This changes `source.js` to match, and updates its unit test, which asserted the query-string shape against a mocked `fetch`, which is why the suite stayed green while the layer was dark.

Reproduced against the running dev server:

```
GET /api/gbfs?url=https%3A%2F%2Faustin.publicbikesystem.net%2F...  → 400
GET /api/gbfs/https%3A%2F%2Faustin.publicbikesystem.net%2F...      → 200
```

## Test plan

- [x] `npm run build` — green
- [x] `npm test` — 3265 tests, 0 failures
- [x] `npm run test:track` — 109 passed, 0 failed
- [x] `npm run format:check` and `npm run check:boundaries` — green
- [x] Before: enabling Bikeshare over Austin shows `UNAVAILABLE — GBFS fetch error`
- [x] After: the same toggle loads and renders stations (W 11th/Congress [7/11], E 11th/San Jacinto [8/11])

Found while working on #359. The two changes are independent. This one is branched from `main` and touches no registry data.


bilawalsidhu:
Landed via #569 (rebased onto the current component layout, with fixes from review). Thank you @MiguelGFerreira!

PR #442, opened 2026-09-14
Add map tilt and north-up compass controls

yashveeeeeeer:
Hey, I was using your God Eyes view locally, and I found out it's really tough to use because the globe is going out of focus. I added two buttons, which are usually standard for all of the map views. Thank you.


INDRANIL1707:
**Great addition overall.** The camera orientation logic and target-frame handling in `cameraOrientationControls.js` look clean, and I like that the controls are integrated through `StyleManager` rather than adding page-level event wiring.

A few UI/qualification points stood out while looking through the change:

**1. Top-center action density / responsive layout**

The new `#tilt-map-view` and `#north-up-view` controls are inserted into the existing `#top-center-actions` group alongside the other globe actions. Could we verify the group at the smallest supported viewport widths and in the relevant panel/mode states, to make sure the additional controls do not crowd or shift neighboring actions/HUD content?

**2. Compass needle rendering**

`#north-up-view` uses a relatively positioned button with an absolutely positioned `N`, while the compass needle is rotated via `transform`. It would be worth checking the rendered result across different device-pixel ratios/browser sizes to ensure the `N`, needle, and button boundary remain visually aligned and do not clip or drift.

**3. Mode-specific layout continuity**

The new controls are hidden with the existing `cockpit-mode` and `scene-playback-mode` rules. Could we also verify that the surrounding action group collapses cleanly when those controls are hidden, without leaving unexpected spacing or shifting adjacent actions?

These look like qualification items rather than blockers to the core camera implementation. I’d be happy to contribute responsive/visual regression coverage for them as a follow-up patch if that would be useful.

yashveeeeeeer:
Thanks for pointing this out. I fixed the first item in a follow-up commit.

On smaller screens, the buttons now move to the top-right. The style label hides, and on very small screens the title changes to the logo only, so nothing overlaps. I checked this at 320px, 375px, 520px, and 720px, including 2x display scaling.

I also checked the compass at 1x and 2x scaling, plus the hidden states in scene playback and Cockpit mode. Both are working correctly, so I do not think the other two items need code changes.

bilawalsidhu:
Landed via #571 (rebased onto the current component layout, with fixes from review). Thank you @yashveeeeeeer!

yashveeeeeeer:
> Landed via #571 (rebased onto the current component layout, with fixes from review). Thank you @yashveeeeeeer!

THANKS @bilawalsidhu

PR #447, opened 2026-09-14
Compose configurable geocoding, place search and route services

samehkhamis:
Geocoding, nearby/text place search and routing currently choose endpoints inside their callers. This change composes those operations through independent provider functions, with configurable endpoints for compatible protocols. Location search, annotations, HUD labels and voice use the supplied service; existing default providers, tool schemas and labelled direct-line route fallback are preserved.

Node Google Places, OSRM routing and Nominatim adapters accept construction-time endpoint/transport configuration. Route/span limits remain in place, configured caches are isolated and cancellation rejects late response bodies. Added adapter/replacement/cancellation tests and documented the contracts in APPLICATION.md.

Validation: doctor; 3,396 passing tests (one existing skip); scoped formatting; package boundaries; build. An additional 249 focused search/voice tests passed after cache-lifetime cleanup. All three CI jobs pass on `8b36c28`. The broad voice-action browser harness reports 34 pass / 3 fail on both this candidate and unchanged main `3b91fef`: Texas region boundary resolution, its dependent follow-up query and the Alps outline. These are reproduced baseline failures, not passing checks. Navigation, annotations, the walking route, scene context, camera ownership and radio behavior passed. All 109 tracking checks passed. Real forward/reverse geocoding, nearby/text search and routing passed in the browser; the Capitol annotation was visually inspected at two angles without page errors. No model/tool schema or pricing changes.


INDRANIL1707:
I’d like to contribute to this PR if the direction is acceptable to both the author and maintainer.

The new `createGeospatialServices()` abstraction looks like a good seam for one additional layer of provider composition. I’m thinking of three closely related additions that would remain within the scope of this PR:

1. **Provider provenance**
   Preserve normalized source metadata with geospatial results, e.g. provider, operation, attribution and retrieval time. This would let downstream consumers distinguish where a result originated without exposing provider-specific response objects.

2. **Capability/provider registry**
   Introduce a small registry/descriptor layer so providers can explicitly declare their supported operations and metadata, while keeping the current `providers.{reverseGeocode,textSearch,nearby,route}` API backward-compatible. This would make provider discovery and capability reporting more scalable as additional geospatial sources are introduced.

3. **Explicit fallback composition**
   Add a small utility for composing providers in a caller-selected order. Fallback would be explicit rather than implicit retry behavior, and cancellation would never fall through to another provider. For example, a caller could intentionally compose `[primaryRouteProvider, backupRouteProvider]` while retaining predictable failure semantics.

I would keep all three deliberately small and test-driven:

* no changes to tool schemas;
* no request-controlled upstream URLs;
* no hidden provider retries;
* existing provider behavior remains the default;
* existing normalized result contracts remain intact.

The implementation could live primarily around `src/search/geospatial.js` plus a small provider-registry/composition module and focused tests.

Would this extension be welcome in #447, or would you prefer these pieces as a follow-up PR?

PR #452, opened 2026-09-14
feat(cctv): enrich identified cameras with datasheet FOV + specs

ch-bas:
## What

When a CCTV source pack (or a seed) names its hardware `model`, resolve it against a camera-model spec table:
- the coverage **frustum uses the manufacturer's horizontal FOV** instead of the pose estimate, and
- the **HUD meta line gains the model's specs** (resolution, night-vision range, PTZ).

Cameras with no identified model are unchanged — unknown never implies anything (matches the layer's existing honesty about `RAW PRIOR` poses). This is a small enabler for the "add a CCTV source pack" contribution lane in CONTRIBUTING.md: a pack that publishes its hardware models gets accurate cones + specs for free. No existing camera's behavior changes (Austin Open Data doesn't publish per-camera models, so nothing lights up until a pack opts in).

## Data source

Spec records are vendored verbatim from the [CCTV Camera Database](https://github.com/ch-bas/cctv-camera-database) — an open, **CC0-1.0** dataset of 12,000+ camera models transcribed from manufacturer datasheets (84% carry FOV). This PR lands a small starter table (4 models) plus the join/parsing helpers; packs or a future full-extract vendor can grow it. Records describe camera **products** (optics, resolution, night vision) — no endpoints, stream paths, or access data.

## Changes

- `src/data/cctvModelSpecs.js` — spec table, case-insensitive + alias lookup, wide-end FOV parser, HUD summary formatter
- `src/data/cctvModelSpecs.test.mjs` — lookup / alias / unknown-null, FOV parsing across the real `fov_deg` string shapes, summary rendering
- `src/data/cctv.js` — apply enrichment in both catalog builders; pass `fovSource` + `spec` through the public camera state
- `src/ui.js` — label a datasheet-driven FOV and append the spec summary

## Checks

`npm test` (2713 pass / 0 fail) and `npm run build` green on Node 24. New behavior is inert until a source declares `model`.

Happy to grow the spec table (or wire the full CC0 extract) if this direction is useful.


ch-bas:
Rebased over the #449 layer refactor — still mergeable against current main (the hooks now live in `src/layers/cctv/`).

Worth connecting: #639 independently reports the same defect class from the heading axis — pose estimates rendering indistinguishably from measured values. #643 fixes that for bearings; this PR is the FOV half of the same provenance principle (`fovSource: 'datasheet'` when a camera's hardware model is identified, tagged in the HUD). Together they cover both pose quantities the layer currently presents more authoritatively than they're known.

PR #457, opened 2026-09-14
feat(clouds): keyless NOAA GOES cloud cover layer

beneduzi:
## What

Adds a **Cloud Cover** data layer rendering near-real-time NOAA GOES-R full-disk **GeoColor** imagery (GOES-19 East, GOES-18 West) from the keyless NOAA/NESDIS STAR CDN, plus the geostationary→geographic reprojection core.

## Why

The app had no cloud/weather imagery layer. GOES is keyless, U.S. public domain, and already refreshed every ~10 min by NOAA. This is the first family of a broader "global cloud coverage" effort (Meteosat/FengYun/Himawari follow-ups); **GOES alone covers the Americas and adjacent oceans only, so this PR does not claim global coverage.**

## How

- `server/providers/geostationary/projection.js` — pure GOES-R ABI fixed-grid math (`satelliteNav`, `scanAnglesToGeodetic`, `geodeticToScanAngles`, `GOES_GRID_HALF_EXTENT_RAD`) with an exact ellipsoid visibility test.
- `server/providers/geostationary/raster.js` — inverse-resamples a full-disk RGBA image into equirectangular parts with an out-of-disk alpha mask, horizon binary-search bounds, and antimeridian splitting.
- `server/providers/goes.js` — `/api/goes/manifest`, `/api/goes/frames/<frameId>/<part>.png`, `/api/goes/status`. Fetches the STAR JPEG, reprojects with `sharp` (decode/encode only; the warp is pure JS), caches content-addressed frames for 5 min (single-flight, per-satellite failure isolation, last-good on failure).
- `src/layers/clouds/` + `src/data/clouds.js` — the layer owns only its own Cesium imagery layers, replaces frames without leaking, and reports guidance instead of drawing when the globe is hidden (Google Photorealistic 3D).
- Registered as `clouds` (token `h`), attribution added, docs updated.

## Coverage / limits

- Americas + adjacent oceans only (GOES-19 + GOES-18). Not a global mosaic.
- Requires a globe imagery stack (Satellite/OSM); photoreal mode hides the globe.
- GeoColor is a rendered scene (clouds/land/snow/night lights), not a cloud-only mask.
- Output is 1024 px tall per satellite frame to bound GPU texture memory.

## Verification

- `npm test` — 3412 pass, 0 fail (1 skipped; allocation microbenchmarks skip on non-Node-24).
- Projection tests anchor on analytic values (sub-satellite point, disk edge, equator closed form), not just round-trips. Proxy tests cover manifest/frame/status, TTL caching, per-satellite failure, and last-good frames.

## Notes

No new keys. `sharp` is already a devDependency; the local dev/preview server needs dev dependencies installed.


bilawalsidhu:
Thanks @beneduzi for the GOES cloud cover layer. It informed the satellite clouds in #617, which is now merged: longwave infrared from GOES-19/18 over North America and a NESDIS global mosaic (about 60°S–60°N), on the same history timeline as radar and lightning.

#617 draws NOAA's hosted infrared imagery rather than reprojecting full-disk GeoColor, so the code here wasn't carried over. Closing this PR in favour of #617.

PR #458, opened 2026-09-14
feat(glm): keyless NOAA GLM lightning layer

beneduzi:
## What

Adds a **GLM Lightning** data layer showing recent lightning flash observations from NOAA's Geostationary Lightning Mapper on **GOES-19 (East)** and **GOES-18 (West)**, keyless via NOAA Open Data on AWS.

## How

- `server/providers/common/noaa-s3.js` — bounded S3 `ListObjectsV2` (XML), granule-key timestamp parsing, and capped object download.
- `server/providers/glm/flashes.js` — NetCDF-4 decode via `h5wasm/node`: `flash_count`, `product_time` units, packed `scale_factor`/`add_offset`/`_FillValue`, unsigned int16 fields, and per-flash records (`id`, `satelliteId`, lon/lat, time, energy, area, quality flag).
- `server/providers/glm.js` — `/api/glm` and `/api/glm/status`. Keeps a **bounded two-minute rolling window** of flashes, lists/downloads unseen granules (hour-rollover aware), dedupes by granule/flash id, expires old flashes, caps the response at 20,000 newest, and serves last-good on failure. Per-satellite failures never drop the other feed.
- `src/layers/glm/` + `src/data/glmLightning.js` — owns one Cesium `PointPrimitiveCollection`, stable flash ids, energy-bucketed color, truthful unavailable state. Registered as `glm-lightning` (token `l`).

## Validation

The decoder was run against a **real GOES-19 granule** (`GLM-L2-LCFA/2026/257/12/OR_GLM-L2-LCFA_G19_s20262571200000_...nc`): 236 flashes, correct 20 s time window, plausible Americas coordinates, unsigned flash ids, finite energy/area.

## Honest limits

These are **optical flash detections from a geostationary view**, not ground-truth cloud-to-ground strike records. Overlapping East/West views can observe the same physical flash, so counts are not globally deduplicated. Requires `h5wasm` (devDependency) on the local dev/preview server.

## Verification

- `npm test` — 3408 pass, 0 fail (1 skipped; allocation microbenchmarks skip on non-Node-24).
- Tests cover S3 XML/granule parsing, packed-value decoding, provider window/dedupe/expiry/truncation/stale behavior, and the client layer lifecycle.

## Notes

`h5wasm` added to devDependencies. No keys required.


bilawalsidhu:
Thanks @beneduzi for the GLM lightning layer. It informed the lightning in #617, which is now merged: observed 15-minute strike density over the Americas and Pacific from NOAA nowCOAST, with no API key, on the same history timeline as radar and satellite clouds.

That is a ground-network density grid rather than the individual GLM flashes you built, so the code here wasn't carried over. If you'd like to bring flash-level GLM data in on top of the new layer, a PR against current main would be very welcome: the observed products share one timeline, panel and image proxy under `src/layers/weather`, and `docs/CODE-BOUNDARIES.md` describes the panel. The same goes for anything else you'd like to see in the weather stack. Closing this PR in favour of #617.

beneduzi:
I will look in this new layer of weather products and try to include this layer in the new format. Thanks for the reply

PR #459, opened 2026-09-14
feat(wind): keyless NOAA GFS wind layer with animated particles

beneduzi:
## What

Adds a **Wind** data layer animating the global 10 m wind field from NOAA GFS, with nullschool-style animated particles over the Cesium globe.

## How

- `server/providers/wind/` — cycle selection, `.idx` parsing, byte-range fetch, ecCodes (WASM) decoding, and resampling.
- `server/providers/wind.js` — `/api/wind/manifest`, `/api/wind/grid/<id>.bin` (Float32 U then V), `/api/wind/status`. Selects the latest available 0.25° cycle, byte-range fetches only `UGRD`/`VGRD` 10 m messages (~1 MB each instead of the whole ~500 MB file), resamples to a compact grid (1° default), and caches per cycle for an hour (single-flight, last-good on failure). Keyless.
- `src/layers/wind/` + `src/data/wind.js` — particle overlay: bilinear field sampling with longitude wrap, per-frame advection, Cesium camera projection, and globe-occlusion culling via `EllipsoidalOccluder`. Registered as `wind` (token `k`).

## Validation

Run end-to-end against the **real 2026-09-14 06z cycle** on Node 25: 180×91 grid at 2°, exact 131,040-byte payload, plausible mean |u| ≈ 4.3 m/s and |v| ≈ 4.0 m/s. The ecCodes WASM decoder requires **Node ≥24**, which `package.json` already requires.

## Honest limits

This is a **forecast, not an observation** — particles show model flow, not measured wind. The particle rendering is a 2D canvas overlay projected through the camera, not a 3D primitive, so it is a visual approximation. No visual browser QA was run in this environment (Node 22 here); the rendering is covered by unit tests with fakes.

## Verification

- `npm test` — 3412 pass, 0 fail (1 skipped; allocation microbenchmarks skip on non-Node-24).

## Follow-up

ECMWF IFS as a second wind model (keyless ECMWF Open Data, `10u`/`10v`) is the intended next PR on top of this one.


bilawalsidhu:
Thanks @beneduzi for the GFS wind layer. It became the base of the wind work in #617, which is now merged: NOAA GFS and ECMWF IFS wind with animated particles, a colour field and a reading at the map centre, alongside observed radar, satellite clouds, lightning and cyclone advisories in a new WEATHER panel.

Your commits from this PR were carried into #617 with your authorship, and the merge commit credits you as a co-author. Closing this PR in favour of #617.

PR #485, opened 2026-09-14
docs: fix dam count inconsistency in README (703 -> 704)

SadMan-ZaMaN:
DATA_SOURCES.md lists Dams (704), but README.md referenced "703 more" in the Field Missions section. Updated the README to use the correct count of 704 for consistency.


bilawalsidhu:
Thanks for checking the count. The current wording is intentional: Three Gorges is one of 704 dams, so there are 703 more. Changing it to 704 more would imply 705 total. Closing without merging.

PR #486, opened 2026-09-14
fix(voice): keep late tool results scoped to their original session

kapelame:
## Description

Stopping voice while a tool or viewport capture is pending can let its late completion restore listening state or send an old call ID/image into a replacement conversation. Capture the session epoch and data channel before asynchronous work and re-check both before publishing results or queuing the next response. Cleanup remains in `finally`; same-session interruption still receives its terminal tool output.

Merged upstream `main` at `2746ec753ce4d6ebafee881a212c3f327907c673` into the existing branch, preserving its history and the new `actionExecutor || runner` dispatch. Resolved the controller conflict and applied the repository's Prettier 3.9.6 formatting. A short comment and resolve/reject regression pin the channel check independently of epoch changes. Runtime documentation and changelog remain included.

## Validation

Tested candidate `19db88fca05bd29bd6ae2f33c95f99c7b2fde526` on macOS arm64, Node 24.14.0 / npm 11.9.0, with a task-local runtime and temporary directory, cleared inherited environment, and no provider keys:

- `PUPPETEER_SKIP_DOWNLOAD=1 npm ci --no-audit --no-fund` — passed.
- `npm run doctor -- --json` — passed; keyless configuration.
- `npm run format` followed by `npm run format:check` — 850 files checked; only the changed controller required formatting.
- `node --test src/voice/gevRealtime.test.mjs` — 167 passed.
- `TMPDIR=<task-local absolute directory> GEV_REQUIRE_ALLOCATION_GATE=1 npm test` — 4,050 passed, 1 Windows-only ACL integration skip, 0 failed, including all 14 allocation probes.
- `npm run build` — passed with the existing chunk-size warning.
- `npm run test:track -- --url http://127.0.0.1:4187` — 109 passed, 0 failed, 0 skipped in Chrome 152.0.7977.83; no console errors or HTTP 5xx. The dedicated dev server served this candidate and was stopped after the run.
- `git diff --check` — passed.

All nine late-completion scenarios fail against the current upstream controller loaded in memory. Removing only the tool guard's channel comparison also makes both new same-epoch cases fail; the candidate passes them. No source files were changed for these negative controls.

No paid OpenAI requests or real microphone session were used. Voice tests exercise the real controller with mocked tool/channel/UI/capture fixtures; browser tracking uses the keyless app and synthetic feeds. Native Windows behavior remains for Windows CI.


Lob26:
Contributor review, not a maintainer verdict. Checked against `main` at `a8125ad`, which now carries the voice-session refactors (`4507f3a`, `584c960`, `55b240e`) plus the mechanical format pass over `src/` and `server/`.

**The bug class is real and the mechanism is the right one.** `stop()` bumps `startEpoch` (`realtimeController.js:919`) and nulls `this.dc` (`:969`), but the tool-execution path in `handleRealtimeEvent` awaits the runner and then keeps going regardless, so a late result can call `setStatus` back to `listening` and push an old `call_id` into a replacement conversation. Capturing the epoch and channel before the await and re-checking after is consistent with how `connect()` already guards itself at `:534` and `:548`.

I tried to break it on two axes and it held on both, so recording that rather than leaving it implied:

- **Cleanup still runs.** The `return`s sit inside the `try`/`catch` of a block with a `finally` at `:1494` that deletes from `activeToolAbortControllers` and `activeRadioToolControllers`, and `return` does not skip `finally`. The test's `activeToolAbortControllers.size === 0` assertion is load-bearing and correct.
- **No radio reservation leak.** The early return does skip `settleRadioToolHandoffReservation` at `:1505`/`:1545`, which looked like a leak, but `stop()` already clears `radioToolHandoffReservations` at `:942` and `settle` no-ops on a missing token (`:1825`, `if (!reservation) return;`). So the token is gone either way.

**Two things to fix before this can land.**

**1. It conflicts, but only mechanically.**

```
$ git merge-tree --write-tree --name-only main refs/pr/486
CONFLICT (content): Merge conflict in src/voice/realtimeController.js
```

`CHANGELOG.md` and `docs/CURRENT-STATE.md` auto-merge. The controller conflict is the format pass plus code motion — the guarded region moved from ~1174 to ~1329 — not a semantic clash. `startEpoch`, `this.dc`, the `finally`, and the `sendVisualContextIfUseful` signature are all still there in the same shape, so the rebase should be a relocation rather than a redesign.

**2. The added lines fail `npm run format:check`.**

`scripts/format-runtime.json` now discovers all of `src` and `server` automatically (`ae99843`), so `realtimeController.js` is in the formatter gate. `format:check` is green on `main` at `a8125ad` (837 files), so this would be PR-introduced. Run through the repo's Prettier 3.9.6:

```js
    const isCurrentSession = () =>
      this.startEpoch === toolSessionEpoch && this.dc === toolSessionChannel;
```

and:

```js
    if (
      !imageUrl ||
      this.startEpoch !== captureSessionEpoch ||
      this.dc !== captureChannel
    )
      return false;
```

`npm run format` will do both. The test file is fine — `src/voice/gevRealtime.test.mjs` isn't in `format-scope.json`, so its long lines don't trip the gate.

**One design question, not a blocker.** `isCurrentSession` requires *both* `startEpoch` and `dc` identity. `startEpoch` alone covers `stop()`/`start()`; the `dc` check adds the case where the channel is replaced without an epoch bump. That's genuinely stricter and I think correct, but it means any mid-session channel swap that doesn't bump the epoch will now silently drop a legitimate in-flight tool result instead of answering it. `this.dc` is only assigned at `:230`, `:469`, `:558` and `:969`, so I could not find a path where that happens today — worth a one-line comment saying the `dc` check is there for a reconnect that reuses the epoch, so a future refactor doesn't read it as redundant and delete it.

The table-driven `resolve`/`reject` and `stopped`/`reconnected` tests are good, and asserting `deepEqual(newSent, [])` on a *fresh* channel is the right shape — it proves the old call never reaches the new conversation, which is the thing that actually matters.

What I verified: the epoch/`dc` lifecycle, the `finally` cleanup, the reservation path, the merge conflict, and the formatting (against the repo's own Prettier, with a green baseline on `main`). What I did not: I have not run the voice suite or exercised a real Realtime session.

PR #489, opened 2026-09-14
feat (cctv): live video feeds, starting with DelDOT

Danielslay86:
Title: cctv: live video feeds, starting with DelDOT

What this is

CCTV cameras that publish a video stream now play as continuous live video instead of refreshed stills. The video projects onto the monitor plane in the scene and plays in the panel card at the same time, from one decoder. Delaware ships as the first agency: a statewide DelDOT loader registers every active camera from their public catalog (about 290 cameras), so the feature is demonstrable out of the box.

This is a re-port of #189 onto the current architecture. That PR predated the CCTV reorg (vite.config.js proxies into server/providers/cctv/, ui.js into src/ui/), so it targeted files that no longer exist. I closed it and rebuilt the same reviewed code on the new structure. The pipeline itself has weeks of soak testing behind it against DelDOT's streams.

How it works

The server owns the upstream session. Two strategies, picked by URL:

.m3u8 upstreams: a Node puller polls the chunklist, re-resolves the master playlist when a session expires, and serves a proxy-owned playlist. No new dependencies.
RTMP and everything else: ffmpeg remuxes with -c copy (no re-encode) into local segments, respawning instantly when the agency drops the session. ffmpeg is optional; without it those cameras fall back to the stills path and nothing else changes.

The client plays the proxy playlist through hls.js. All media requests stay origin-pinned: sub-paths resolve only against the registered camera's origin, and a playlist reference that would resolve elsewhere is refused. Segments live in a short rolling window under the OS temp dir and get reaped when idle.

Why the playback-rate governor exists

DelDOT's encoders deliver roughly 52 seconds of content per 60 seconds of wall time, so no transport can play them continuously at 1x. The client holds playbackRate slightly under 1.0 based on buffer depth and eases back to 1.0 when the buffer is healthy. Cameras with honest clocks sit at 1.0 and never leave it. The stream ends up 20 to 30 seconds behind live, which is the tradeoff for never starving.

Testing

10 minute uninterrupted soak on a DelDOT camera (scenario now in TESTING.md section 6)
Camera switching, projection toggle, and stills/video coexistence all covered in that scenario
Session death and recovery exercised heavily: DelDOT restarts every stream on a 60 second timer, which turned out to be a good adversarial test bed

Known limitations

DelDOT rate limits per IP. Bursts of rapid camera switching can bring on 404s and faster drains until it cools off. Normal use does not hit this.
The roughly once a minute stream restart shows as a brief hitch. That is the agency's publish cycle, not a pipeline bug.

Not in this PR

More agencies. Caltrans exposes a streamingVideoURL field in the same catalog the existing loader already reads, so that is the natural next one. I would rather land this and follow up than grow the diff.
A WebGL render path for the panel card (sharper video at small sizes).


bilawalsidhu:
Thanks @Danielslay86 for the live video work. Most of it landed on main in #617: HLS cameras play as live video through one decoder shared by the monitor plane and the panel card, the server pulls and origin-pins the playlists, the buffer-based playback-rate governor keeps DelDOT playing, and the statewide DelDOT pack loads by default from DelDOT's official HTTPS HLS links. Your commits are in #617 with your authorship, the merge commit credits you as a co-author, and CHANGELOG.md credits you for #489.

What didn't land is the ffmpeg remux path for RTMP and other non-HLS streams. Segments stay in bounded memory, and RTMP-only sources aren't supported. I'm leaving this PR open for that part. If the merged version behaves differently from your soak tests, please say so here.

Danielslay86:
Closing this since it landed in #617 as part of the integration merge. Thanks for bringing it in and hardening it. The memory-only sessions and bounded leases are a better fit than the disk segment store I had.

Next up is Caltrans. Most of its cameras publish an HLS stream URL in the catalog the existing loader already reads, so it should run on the new pipeline without server changes. I'll open that as a small pack PR off current main.

PR #493, opened 2026-09-14
feat(vessels): add keyless APRS-IS live vessel source

beneduzi:
Adds an optional APRS-IS provider for the existing vessel layer. It reuses the
current vessel ingestion, selection, cards and tracking path; it does not add a
second visual layer.

## Why APRS-IS, not APRS.fi

APRS.fi does not expose a public WebSocket and its REST API requires a private
API key, so it cannot back a keyless live feed. APRS-IS is the documented,
receive-only amateur-radio network and is the correct source for a continuous
stream. No APRS.fi endpoint is called and nothing is scraped.

## What changed

- `server/providers/vessels/aprs-live.js`: server-side, receive-only TCP/TLS
  client with line framing, bounded record cache (2,000 records, 30-minute
  expiry), and reconnect with exponential backoff.
- `server/providers/vessels/aprs-parser.js`: dependency-free parser for AIS
  `!AIVDM` position reports (Class A types 1/2/3 and Class B type 18), maritime
  APRS position packets (`!`, `=`, `/`, `@`), and APRS telemetry.
- `/api/aprs-live` and `/api/aprs-live/track`: same-origin JSON endpoints.
  Returns 503 with `configured:false` when the source is not enabled;
  credentials never reach the browser.
- `src/sources/live/standalone.js`: `createAprsIsSource()` implementing the
  existing provider-neutral vessel contract, exported from
  `src/sources/live/index.js`.
- `src/data/aisLiveVessels.js`: env-driven provider selection
  (`VITE_VESSEL_SOURCE=aprs-is`); AISStream remains the default.
- Attribution in `src/data/dataCredits.js` and `DATA_SOURCES.md`, plus
  `docs/APRS-IS.md`.

## Viewport filter

Snapshot requests carry the camera's visible bounds, which the server converts
to an APRS-IS area filter (`a/latN/lonW/latS/lonE`). Antimeridian crossings are
split into two boxes. Updates are applied to the live connection without
reconnecting and are coalesced (change check plus a 5-second minimum interval),
so camera movement cannot flood APRS-IS. `APRS_IS_FILTER` remains the fallback
before the first viewport and whenever the viewport is absent or degenerate.

## Tracking

Positions accumulate in a bounded per-reference history (64 samples per vessel,
5,000 vessels). Selecting a vessel backfills its recent path through
`/api/aprs-live/track?reference=...`, reusing the existing trail path. The
reference is validated before lookup and is never interpolated into an APRS-IS
filter.

## Vessel telemetry

APRS telemetry reports (`T#sss,a1,a2,a3,a4,a5,bbbbbbbb`) are parsed and kept per
source callsign: latest sample plus a bounded recent history (32 samples).
Analog channels are exposed raw because their units live in the station's
separate `PARM`/`UNIT`/`EQNS` messages; no unit is invented.

The latest reading is rendered through the existing vessel card detail-line
contract rather than a dedicated renderer:

- selected card — a `TLM #n · A: … · D: …` detail line, alongside the existing
  type/speed/heading, destination and identity lines;
- ambient card — a compact `TLM #n` summary, only when the card has no other
  metrics to show, so decluttered labels stay small;
- vessel HUD — a monospace telemetry line below the identity row.

Telemetry is carried through the store on both create and in-place update, so a
refresh without a report keeps the last known reading. APRS rows are labelled
`APRS`/`CALL` while AIS rows keep `AIS`/`MMSI`, and callsign-only beacons use
the callsign as their identity.

## Configuration

`APRS_IS_ENABLED=1` (or `APRS_IS_HOST`), `APRS_IS_HOST`, `APRS_IS_PORT`,
`APRS_IS_CALLSIGN`, `APRS_IS_PASSCODE` (`-1` is receive-only) and
`APRS_IS_FILTER`. Credentials stay on the server.

## Verification

- `node --test server/providers/vessels/aprs-parser.test.mjs server/providers/vessels/aprs-live.test.mjs src/sources/live/aprsIs.test.mjs src/sources/live/contract.test.mjs` — 24/24, including an independent spec-based AIS encoder that decodes Class A and Class B position reports, viewport/antimeridian filter construction, history caps, telemetry parsing, and the endpoint contract.
- `node --test src/data/aisLiveVessels.test.mjs` — 84/84, covering telemetry formatting, APRS/AIS labelling, HUD output, and in-place telemetry propagation across refreshes.
- `npm test` — 3407 passed, 1 skipped, 0 failed.
- `npm run build` — passed (the worktree logs an unrelated Cesium asset-copy warning because `node_modules` resolves from the parent checkout).
- `prettier --check` on the changed files — clean.

## Limitations

- AIS multipart (type 5 static data) and long-range type 27 reports are not
  decoded.
- APRS beacons have no MMSI, so their identifier is the callsign.
- Telemetry analog channels are shown as raw values; resolving them to physical
  units would require consuming each station's `PARM`/`UNIT`/`EQNS` messages.


bilawalsidhu:
Thanks @beneduzi. Could APRS-IS be an additional vessel source alongside AISStream, instead of an env switch that replaces it? Then both can be on together. Tracking in #714.

beneduzi:
Yes It could, but they track the same vessels, APRS-IS tracks fewer then AISStream, some diferent, but there is a lot of duplicates, since I could not think a simple way to solve this colisions, I choose to display either or another. Any suggestions on how to solve this?

PR #455, opened 2026-09-14
feat(firms): add MODIS NRT active fires

beneduzi:
## What

Adds NASA FIRMS **MODIS_NRT** (combined Terra + Aqua) to the existing **FIRMS Active Fires** layer, which today merges the three VIIRS NRT sources (NOAA-20, NOAA-21, Suomi-NPP).

## Why

The Active Fires layer only surfaced VIIRS detections. MODIS adds a second, independent sensor family (Terra/Aqua overpasses, ~1 km pixels vs VIIRS ~375 m) and is the most requested fire source. The area-CSV parser in `src/data/firmsCsv.js` already tolerated the MODIS schema, so this is a source-level addition rather than new plumbing.

## Changes

- `server/providers/firms.js` — add `MODIS_NRT` to the sequential source list; update the module contract comment.
- `src/data/firmsCsv.js` — document both the VIIRS and MODIS headers, the categorical vs numeric `confidence`, and the `bright_ti4`/`brightness` + `bright_ti5`/`bright_t31` mapping. No parsing logic change.
- `src/data/firmsCsv.test.mjs` — MODIS parse coverage.
- `src/data/firmsProxy.test.mjs`, `src/tooling/environmentProviders.test.mjs` — update 3→4 source expectations.
- `DATA_SOURCES.md` — document MODIS alongside VIIRS (CC0/public domain, shared `FIRMS_MAP_KEY`, trailing 24 h clamp, 30 min cache).
- `docs/CURRENT-STATE.md`, `CHANGELOG.md` — runtime table + changelog.

MODIS shares the existing `FIRMS_MAP_KEY` and 30-minute proxy cache; sources are still fetched sequentially to respect the FIRMS transaction quota.

## Verification

Rebased onto current `main` (was 65 commits behind) and re-run against it:

- `npm test` — **3897 pass, 0 fail** (1 skipped; allocation microbenchmarks skip on non-Node-24).
- `npm run check:boundaries` — pass.
- `npm run build` — pass.
- `prettier --check` — clean on the files this PR touches.

Two files this PR edits (`server/providers/firms.js`, `src/data/firmsCsv.test.mjs`) had drifted from Prettier's formatting; both are formatted again, and the change is limited to this PR's own lines. Four other files (`CHANGELOG.md`, `DATA_SOURCES.md`, `docs/CURRENT-STATE.md`, `src/data/firmsProxy.test.mjs`) are already non-compliant on `main` and were deliberately left untouched to avoid unrelated churn.

## Notes

[...]

PR #497, opened 2026-09-14
feat(firms): keyless NOAA GOES ABI fire detections

beneduzi:
Makes the Active Fires layer work **without any API key**, by reading the NOAA GOES-R ABI fire/hot-spot product straight from the public GOES S3 buckets.

## Why GOES

The layer currently merges three VIIRS NRT sources through `/api/firms`, which needs a free `FIRMS_MAP_KEY`; without it the layer is empty. Keyless MODIS/VIIRS is not an option: `MOD14`/`MYD14` and `VNP14IMG` are distributed through NASA LAADS/LP DAAC and require an Earthdata Login token for programmatic download, even though the data are public. There is no AWS Open Data bucket for them.

GOES is different. `ABI-L2-FDCF` lives in `noaa-goes18`/`noaa-goes19`, which are AWS Open Data: anonymous access, `us-east-1`, not requester-pays. I verified against the live buckets while building this — the product is still produced, with granules landing every 10 minutes (`Last-Modified` of the newest granule was minutes old when I checked). GOES-16 and GOES-17 are decommissioned and produce no 2026 data, so only 18/19 are used.

The trade-off is geometric and worth stating: GOES is geostationary, so it images the full disk every ~10 minutes and catches fires far sooner than a twice-daily polar overpass — but at ~2 km at the sub-satellite point (coarser toward the limb) and only over the Americas and adjacent oceans. VIIRS is ~375 m. The providers are complementary, so **FIRMS still wins whenever its key is configured**.

## Data path

```
noaa-goes19.s3.amazonaws.com/ABI-L2-FDCF/YYYY/DDD/HH/OR_ABI-L2-FDCF-M6_G19_s…_e…_c….nc
        │  anonymous HTTPS, ~1.8 MB HDF5
        ▼
server/providers/goes-fires.js     list newest key → download → cache (10 min)
        ▼
server/providers/goes/fireGrid.js  HDF5 decode + ABI projection + detection
        ▼
GET /api/goes-fires                FIRMS-shaped rows
        ▼
src/layers/firms/source.js         keyless fallback on 503 no_key
        ▼
existing fire layer                adapt → cards → render (unchanged)
```

## Finding the newest granule

The `s<timestamp>` segment is the scan start and does **not** land on the 10-minute boundary (e.g. `s20262570000215` = 00:00:21.5Z), so the key cannot be derived by rounding the clock. The provider lists the current hour prefix and takes the last key, walking back an hour at a time when the prefix is empty.

## Decoding and detection

The granule is NetCDF-4/HDF5, read with `h5wasm` (WASM HDF5 — no native toolchain). Scale factors, offsets and fill values are read from the file rather than hardcoded.

Detection is `Power > 0`, and that criterion is **checkable against the granule's own aggregates**: on the 2026-09-14 00:00Z GOES-19 granule, `Power > 0` selects 83 pixels and filtering `DQF == 0` leaves exactly 79 — the value of the file's `total_number_of_pixels_with_fire_radiative_power`. Valid `Temp` pixels (59) match `total_number_of_pixels_with_fire_temperature` the same way.

`Mask` is a scene-classification bit field. Its class semantics are deliberately **not** reimplemented — the raw value is passed through, because guessing the bit meanings would silently mislabel detections.

The product ships no `Lat`/`Lon` variables, so the ABI fixed grid is inverted using `perspective_point_height`, the two axes and `longitude_of_projection_origin`. Note that the latter is in **degrees** while the formula works in radians; mixing them yields longitudes near −4300°.

Confidence: FDCF has no FIRMS-style confidence category. `DQF` maps to the vocabulary the layer already understands (`h`/`n`/`l`), documented as a DQF-derived tier. The string form matters — `src/data/firmsAdapt.js` interprets numeric confidence as the MODIS 0–100 scale and would divide a 0–1 value by 100.

## Client wiring

`createFirmsSource` falls back to `/api/goes-fires` **only** when `/api/firms` answers `503 no_key`, so a genuine FIRMS outage (500/502) is never masked by the fallback. Both providers return the same `fires` row shape, so the adapter, cards, viewport and rendering path are untouched. `getStats()` reports which backend answered via `provider`, and the layer's source label names both.

## Verification

- `node --test src/data/goesFireGrid.test.mjs src/data/goesFiresProxy.test.mjs` — 8/8. The grid test builds a **real HDF5 granule** with `h5wasm` and decodes it end to end, covering fill-vs-value scaling, DQF/mask passthrough, day/night and the scan-angle indexing; the proxy test drives the middleware through list → download → decode → JSON and covers cache hits, partial satellite failure and the 502 path.
- **Live smoke test**: listed and decoded an actual GOES-19 granule (1.78 MB) — **162 detections**, 155 with `DQF == 0`, coordinates landing in southern Brazil/Uruguay. 
- `npm test` — 3430 passed, 1 skipped, 0 failed.
- `npm run build` — passed.
- `prettier --check` on the changed files — clean.

## Note on the boundary checker

`h5wasm` must be imported as `h5wasm/node` (the root entry is a browser/MEMFS build that cannot open a real path, and neither build exposes `FS`). The checker required every `external` entry to be an exact dependency name, so it rejected the subpath. It now accepts a declared dependency **or a subpath of one** — the exact specifier still has to be listed so the bundler externalizes it, and the dependency it belongs to must still be declared for that runtime.

## Limitations

- Only the full-disk `FDCF` product; `FDCC` (CONUS) and the mesoscale products are unused.
- GOES coverage stops at the limb — no fires outside the Americas and adjacent oceans.
- Only the newest granule per satellite; no multi-scan temporal compositing or persistence filter.
- `Mask` class semantics are passed through undecoded.


bilawalsidhu:
Thanks @beneduzi. Active Fires with no key is well worth having, and the projection maths and HDF5 fixture tests are careful. We ran it against live granules, and a few things need fixing before it can ship:

1. Memory: one decode took the server from about 115 MB to 545 MB RSS, because Power, Temp, Area, Mask and DQF are read whole. Read `Mask` first, pick the fire pixels, then read only those indices from the other datasets.
2. False positives: detection is `Power > 0` and confidence comes from DQF, which was 0 for every live detection. Please take confidence from the FDC `Mask` codes (10/11/30/31 high, 13/33 nominal, 14/15/34/35 low) and drop or flag the cloud-contaminated 12/32.
3. `solar_zenith_angle` is a scalar in FDCF, not a per-pixel grid, so `daynight` always comes out `D`. Compute it from scan time and longitude, or drop it (the test fixture currently models it as a grid).
4. Add `AbortSignal.timeout` and capped reads to the S3 listing and granule fetches, and validate the disk cache the way `firms.js` does.
5. Label the fallback clearly: a GOES source line with its coverage (Americas, newest 10-minute scan), and keep the key hint visible so users elsewhere know how to get 24-hour global coverage.
6. Bump `h5wasm` to a current release and rebase on main.

Separately, your MODIS change (#455) is going into a fire PR we're putting together now, with your authorship kept.

beneduzi:
Thanks for the detailed review — I went through the requested changes and pushed the fixes.

- **Memory:** the decoder now reads `Mask` first, selects the fire pixels, and only reads `Power`, `Temp`, `Area`, and `DQF` for those indices using hyperslabs. The full telemetry grids are no longer materialized.
- **False positives:** detections are still based on positive fire power, but confidence now comes from the FDCF Mask codes:
  - `10/11/30/31` → high
  - `13/33` → nominal
  - `14/15/34/35` → low
  - `12/32` → dropped as cloud-contaminated
  `DQF` is retained as diagnostic data only.
- **Day/night:** `solar_zenith_angle` is handled as a scalar, and missing or invalid values now remain unknown instead of being reported as daytime. The ABI fixed-grid projection was also corrected and validated.
- **S3 and cache safety:** listing and granule fetches now use abort timeouts and streaming byte caps. Disk cache entries are versioned and validated, future timestamps are rejected, and cache write failures no longer turn a successful fetch into an error.
- **Fallback labeling:** the active fallback now reports:
  `NOAA GOES ABI · Americas · newest available ~10-minute scan`
  It also keeps the non-blocking hint that adding `FIRMS_MAP_KEY` enables global detections over the trailing 24 hours.
- **Dependency/rebase:** `h5wasm` was updated to `0.10.3`, and the branch was rebased on current `main`.

I also updated the affected tests and verified the full suite, focused GOES/FIRMS tests, package boundaries, formatting, and production build.

PR #568, opened 2026-09-15
feat(aqhi): add the Air Quality Health Index layer

rileygramlich:
## What

Adds a keyless Air Quality Health Index layer covering Canada, read live from Environment and
Climate Change Canada's MSC GeoMet OGC API. Around 134 stations across all five ECCC administrative
zones, coloured on the published 1–10+ health-risk bands.

It follows the current layer shape — a pure `model.js`, an injected snapshot `source.js`, and a
factory taking `{ source, overlayHost }` — with the standalone provider wired in `layerSources.js`
and the app instance in `app/layers/aqhi.js`. I used the earthquakes layer as the template since
it's the closest existing analogue: a keyless polled reference feed with no credential to broker.

AQHI pairs naturally with the fire layer already in here. FIRMS shows you where something is
burning; AQHI shows where the smoke ended up, which during a bad season is a few provinces away and
over the places people actually live.

## Two things worth a careful look

**AQHI is a published index, not a measurement.** ECCC reports whole numbers from 1, with anything
above ten as the open-ended "10+". The API hands you the underlying decimal — 1.08, 2.32 — and
publishing that as-is would invent a precision the index doesn't claim.

While writing the rounding I hit something worth flagging: `Number(null)` and `Number('')` are both
`0`, which then floors to a confident **AQHI 1** for a station that reported nothing at all. A
missing reading has to stay missing, so the empty forms are guarded before any coercion. There's a
test pinning it.

**The station catalog is load-bearing.** It's the only thing that supplies a reading its
coordinates, so observations are joined against it rather than trusted to be placeable. That also
sets the cadence: the catalog changes a few times a year and is cached twelve hours, while readings
poll every twenty minutes against an hourly publication. The observations query uses `latest=true`
— without it the collection returns the full history, roughly 2,000 rows nationally. An empty
catalog refresh can't blank a good one, and readings older than six hours are dropped rather than
shown as current.

## One honest caveat about coverage

**Quebec is thinly represented, and that's upstream rather than anything this layer does.** The
province runs its own Info-Smog program, so only a handful of Quebec stations appear in ECCC's
national AQHI feed. The reporting count also drifts hour to hour as stations report or fall silent
— a live run just now returned 124 of ~134 across four of the five zones. It's documented in
`DATA_SOURCES.md` so nobody reads the gap as a bug.

## Verification and budgets

- `npm test` green on Node 24.14.0 and 26.8.2 (allocation gate enforced on 24.14.0 via
  `GEV_REQUIRE_ALLOCATION_GATE=1`)
- `npm run build`, `npm run format:check`, `npm run check:boundaries` all clean
- `npm run test:track` 109 passed, 0 failed
- `node scripts/qa-aqhi.mjs` — 5 passed, 0 failed against the live feed: 124 stations, coverage
  across four ECCC zones, every reading in published integer form
- 24 unit tests across the model and the layer's ownership/lifecycle

The station cohort is registered in the overlay allocation gate. Measured across repeated
clean-process probes: median 3,788 B/frame, max 3,842, and 80.1 B/candidate at the ceiling — inside
the shared 154 B/candidate bound, so no per-source exception is taken. The 4,800 frame budget
leaves ~25% headroom. Measured standalone rather than stacked on the Phase-5 host, following the
submarine-cable precedent, since AQHI stations are sparse and continental.

## Licensing

ECCC Data Servers End-use Licence — open, attribution required, no endorsement implied. Registered
in `dataCredits.js` and documented in `DATA_SOURCES.md`.

---

Follow-on from #438 / #514 (Calgary cameras) — thanks for the rebase and cleanup on that one, the
component layout made this a much easier layer to slot in.


bilawalsidhu:
Thanks @rileygramlich. The model, tests and allocation-gate work are careful. Since this opened, main has a Weather category with a readout panel in the right rail, and air quality belongs there. We'd merge a version shaped like this:

1. A general "Air quality" layer in the Weather group with ECCC AQHI as the first provider: records shaped like `{ provider, scale, value, band }` so other networks can be added later, and "Canada only" stated in the coverage line.
2. A card in the WEATHER panel (`readout: true`; see "Adding a right-rail readout panel" in `docs/CODE-BOUNDARIES.md`) with the band legend and the highest readings in view.
3. In `update()`, add the abort controller and enabled guard the earthquakes layer uses and pass the `signal` through, so a late response after disable can't redraw.
4. Rebase on main: reference sources moved to `src/sources/reference.js`, and the allocation profile needs re-measuring against the current worker.
5. Leave the share-link token to us (#658); `h` is taken now.

Happy to review when it's rebased.

rileygramlich:
Thanks — that's a much better shape than what I had. All five points are in, force-pushed onto
current main.

**1. Generalized to a provider-backed layer.** It's now `air-quality` in the Weather group with
ECCC AQHI as the first registered provider. Readings carry
`{ provider, scale, value, band }` rather than a bare number — writing it out made the reason
obvious, since AQHI is 1–10+ on health risk while the US AQI is 0–500 on concentration
breakpoints and the EAQI is a five-step category. There's no shared numeric meaning to fall back
on, so the record names which index it's on. `AIR_QUALITY_PROVIDERS` is the seam a second network
registers against. Coverage reads "Canada only · ECCC AQHI network" on the provider, the row and
the card.

**2. WEATHER readout card.** `readout: true` with the band legend and the worst readings in view,
using `getRowControls`/`setRowControlsListener` the way the wind and cyclone cards do. Two edge
cases that seemed worth handling rather than leaving: an unresolvable view rectangle (space-level
or fully oblique camera) degrades to the worst readings overall instead of a blank card, and a
rectangle crossing the antimeridian selects correctly. Both are tested.

**3. Abort and enabled guard.** `update()` now follows the earthquakes shape — aborts any
in-flight request, passes `signal` through to the source, and refuses to touch the scene if the
request was superseded, aborted, or the layer was disabled while it was outstanding. `disable()`
and `destroy()` abort too. Three tests cover it: a response landing after disable, a superseded
refresh not clobbering a newer one, and `update()` while disabled.

**4. Rebased.** Source registered in `src/sources/reference.js`. Re-measuring the allocation
profile was worth doing — against the current worker it's median 3,900 B/frame and max 4,114,
where my earlier numbers were 3,788 and 3,842. Still well inside the 154 B/candidate ceiling at
85.7, but the old budget's headroom had quietly dropped from 25% to 16%, so it's now 5,200.

**5. Token left to you.** I've used `'3'` as a placeholder purely so
`validateLayerStateRegistry` passes — reassign it in #658 and I'll rebase, or change it in place.

Verified on Node 24.14.0 and 26.8.2: unit suite, `format:check`, `check:boundaries`, build, and
`npm run test:track` (109/0). `scripts/qa-air-quality.mjs` covers it live — 6/6, 123 stations
across four ECCC zones, and the readout list populated with the camera over Canada.

One caveat I'd rather state than have you find: **Quebec is thinly represented**, because the
province runs Info-Smog separately and only a few of its stations appear in ECCC's national feed.
It's noted in `DATA_SOURCES.md` so the sparse patch over Quebec doesn't read as a bug.

PR #180, opened 2026-09-04
fix(gbfs): refuse upstream redirects and enforce the body cap while streaming

dlucas489:
GBFS now refuses upstream redirects and enforces its 5 MiB response cap while streaming, preventing a validated feed from redirecting the proxy elsewhere or exhausting memory with an unbounded body.

The contribution is adapted to the current GBFS provider module. The 12-second deadline also covers body reads, and rejected downloads are cancelled. Development and preview use the same handler.

Validation: focused redirect, streaming, timeout, cancellation and dev/preview handler tests; full unit suite, setup doctor, formatting, package boundaries and production build. GitHub CI must pass before merge.

Original contribution by @dlucas489; provider adaptation and additional timeout/cleanup fixes by @samehkhamis.

Closes #30 and #31.

PR #581, opened 2026-09-15
fix(providers): bound the remaining upstream body reads

Mikejr314159:
Six proxied fetches still buffered upstream bodies with no size limit: FIRMS CSV, CelesTrak TLE groups, adsb.lol `/v2/mil` (both the OK and relayed-status paths), TomTom flow tiles, OpenSky `states/all` and adsbdb enrichment lookups. This routes each through the shared capped readers in `src/sources/httpBody.js`, following the pattern #180 established for GBFS.

`readResponseBytesCapped` gains the same optional `signal` as `readResponseTextCapped`, with the same cancel-on-abort plumbing and reader release, so the TomTom tile path can share its route deadline instead of reading `arrayBuffer()` unbounded; the existing GTFS-Realtime caller is unchanged. Caps are bounds sized well above measured payloads, not budgets (FIRMS 64 MiB per source, CelesTrak 16 MiB, OpenSky/adsb.lol/TomTom 8 MiB, adsbdb 1 MiB). Where a route already had a deadline, the same signal now covers the body read. Two polls had no deadline at all: the adsb.lol military poll gains a 12-second one and the OpenSky `states/all` exchange (including its Basic-auth retry) a 20-second one; both miss into their existing STALE paths.

Deliberately left alone: FIRMS map-key status, the OpenSky token exchange and CCTV media (#26/#28). `traffic-provider` and `firms-provider` now declare `common/http.js` and `src/sources/httpBody.js` in package-boundaries, matching the other groups; the new test is adopted into format-scope in a separate mechanical commit.

Validation: focused reader tests in `src/sources/httpBody.test.mjs` (exact bytes, chunk order, Content-Length rejection without pulling, streaming cap with cancel, exactly-at-cap, abort mid-stream never returns a partial body, pre-aborted signal, non-stream fallback); full unit suite (4,143 tests, 0 failures), formatting, import directions, package boundaries and production build. Run on Node 24.13.1, one patch below the declared `>=24.14.0` engine, so CI on the pinned runtime is the authoritative check.

Noticed while building a similar CesiumJS globe (a weather/storm tracker) and converging on the same capped-read design for its own proxies; this seemed like the useful piece to bring back.


Lob26:
Contributor review, not a maintainer verdict. I checked this against `main` at `a8125ad` because a batch of refactors landed after this branch was cut, and some of them move ground under it.

**The intent is right and still unmet** — `server/providers/aircraft/opensky.js:496` (`await upstream.text()`) and `server/providers/traffic.js` (`arrayBuffer()`) are genuinely unbounded at HEAD. Four notes, roughly in the order I'd address them.

**1. `readResponseBufferCapped` duplicates a helper that already exists**

`main` already exports `readResponseBytesCapped(response, maxBytes)` from `src/sources/httpBody.js:61`, re-exported through `server/providers/common/http.js`. It already does the early `Content-Length` rejection, the running cap, the `reader.cancel()` and the same `{ code: 'RESPONSE_TOO_LARGE' }` error. It is live — `src/sources/transitService.js:187` uses it.

The one thing it genuinely lacks is what this PR actually adds: the `signal` parameter and the abort plumbing. So the smaller change is to add `signal` to `readResponseBytesCapped` rather than land a second binary reader beside it. For TomTom, `Buffer.from(bytes.buffer, bytes.byteOffset, bytes.byteLength)` at the call site covers the `Uint8Array` → `Buffer` difference without a separate function.

**2. It will conflict — `common/http.js` became a re-export barrel**

```
$ git merge-tree --write-tree --name-only main refs/pr/581
CONFLICT (content): Merge conflict in server/providers/common/http.js
CONFLICT (content): Merge conflict in CHANGELOG.md
```

The implementations moved to `src/sources/httpBody.js`; `common/http.js` is now mostly a re-export list. The new function is added to what is no longer the implementation file, so the rebase is also a relocation.

**3. `npm run check:boundaries` will fail as written**

The PR adds `server/providers/common/http.js` to `traffic-provider` and `firms-provider` but not `src/sources/httpBody.js`. All ten existing groups that declare the former also declare the latter:

```
groups declaring common/http.js: 10
  live-providers, place-providers, space-providers, gbfs-provider, cctv-provider,
  radio-provider, overpass-provider, military-installations-provider,
  regional-provider, transit-provider
missing httpBody.js: none — every group pairs them
```

`scripts/check-package-boundaries.mjs` runs a real Vite build per group and rejects any module in the resulting graph that isn't declared (`Package boundary ${name} imports an unowned module`), and it deliberately does not let tree shaking excuse an unused import. Both new entries need `src/sources/httpBody.js` alongside.

**4. OpenSky gets a size cap but still has no deadline at all**

This is the one I'd most like to see folded in, since the PR already establishes the pattern. `states/all?extended=1` at `server/providers/aircraft/opensky.js:473` and the Basic-auth retry at `:488` carry no `signal` — not on the headers, not on the body. The PR gives adsb.lol a 12 s deadline for precisely this reason and the doc text explains why, but the primary worldwide aircraft source keeps none, so a hung OpenSky connection still stalls the poll indefinitely with or without this patch. `readResponseTextCapped(upstream, OPENSKY_STATES_MAX_RESPONSE_BYTES)` is also the only new call site passing no signal, which reads as an oversight rather than a decision.

**Smaller thing:** `src/data/readResponseBufferCapped.test.mjs` sits in `src/data/` but tests a module owned by `src/sources/`. `src/sources/httpBody.test.mjs` would keep it next to the implementation and follow item 1 naturally.

The tests themselves are the best part of this PR — the zero-`highWaterMark` `endlessResponse` harness that counts pulls proves the cap stops the read rather than just rejecting afterwards, which is the property that actually matters here. Worth keeping whichever way the helper question is resolved.

What I verified: the duplicate helper and its live consumer, the merge conflict, the boundary invariant (by inspection of all ten groups and of the checker), and the missing OpenSky deadline. What I did not: I did not run the suite, the build, or `test:track` against this branch, and I have not executed the branch at all.

Mikejr314159:
Thanks — this is a careful read and all four land. Reworked on top of `a30c423`:

1. **Duplicate helper** — dropped `readResponseBufferCapped`. `readResponseBytesCapped` now takes the optional `signal` with the same cancel-on-abort plumbing and reader release as the text reader; TomTom uses it with `Buffer.from(bytes.buffer, bytes.byteOffset, bytes.byteLength)` as you suggested. The existing GTFS caller is unchanged.
2. **Barrel relocation** — rebased; the change now lives in `src/sources/httpBody.js` and `common/http.js` is untouched.
3. **Boundaries** — `traffic-provider` and `firms-provider` declare `src/sources/httpBody.js` alongside `common/http.js`, matching the other ten groups. `npm run check:boundaries` passes.
4. **OpenSky deadline** — added `OPENSKY_STATES_TIMEOUT_MS = 20000` covering headers, the Basic-auth retry and the body under one signal; a miss serves STALE like the other failures. You're right that the signal-less call was an oversight, not a decision.

Test moved to `src/sources/httpBody.test.mjs`, same harness. Full suite, format, boundaries and build re-run locally.

PR #588, opened 2026-09-15
feat(layers): add live RainViewer weather radar layer

sandivd:
## Summary
Adds a live global weather radar (precipitation) data layer powered by the free RainViewer API.

- **Backend:** Adds `/api/weather-radar/metadata` and `/api/weather-radar/tile` server proxies with in-memory caching and rate limiting (`server/providers/weatherRadar.js`).
- **Frontend:** Implements `weatherRadarLayer` (`src/data/weatherRadar.js`) registered in `LAYER_STATE_REGISTRY` and data toggles.
- **Cesium Integration:** Drapes precipitation tiles on the Cesium globe with parent-level upscaling so the radar remains smooth and visible when zooming across regional and street levels without flickering.
- **Attribution & Compliance:** Registered static and dynamic RainViewer credits in `dataCredits.js` and `DATA_SOURCES.md`.

## Verification
- `node --test src/data/weatherRadar.test.mjs`: Passing.
- `node --test src/data/layerState.test.mjs`: Passing.
- `npm run check:boundaries`: Passed.
- `npm run format:check`: Passed.
- `npm run build`: Production build passes cleanly with no bundle errors.
- Verified in browser: Toggling Weather Radar renders live precipitation bands globally without flickering and stays visible when zooming in.


sandivd:
Saw the link to #617, looks incredible!

Noticed that #617's radar is CONUS (US-focused). If you'd like a keyless global provider to cover international regions alongside NOAA, the RainViewer proxy here provides worldwide coverage. 

Happy to adapt this provider to plug cleanly into the new weather architecture once #617 lands.

bilawalsidhu:
Thanks @sandivd for the RainViewer radar layer, and for the kind words on #617. It informed the radar there, which is now merged: NOAA MRMS reflectivity for the contiguous US on one history timeline with satellite clouds and lightning. #617 uses a different source, so the code here wasn't carried over.

You're right that radar outside the US isn't covered. If you'd like to bring RainViewer in as a global radar provider on top of the new layer, a PR against current main would be very welcome: the observed products share one timeline, panel and image proxy under `src/layers/weather`, and `docs/CODE-BOUNDARIES.md` describes the panel. Closing this PR in favour of #617.

PR #598, opened 2026-09-15
Feat/search locate zoom fix

Sibu05:
I added the Self-locate src/ui/selfLocate.js:1 + #self-locate-btn in index.html:43 (my_location) — click-only navigator.geolocation, ephemeral, never transmitted, toast on error. for a user to be able to locate their location i had problems with locating my current location as i would like to explore my nearby surroundings. and also added
Trackpad zoom src/ui/zoomSmoothing.js:1 + src/app/viewer.js:26 inertia (inertiaZoom 0.85 etc.) + rAF easing with trackpad/pinch vs mouse sensitivity.


bilawalsidhu:
Thanks! Against current main this duplicates existing shell methods and top-bar actions, and pinch zoom is already in. A small, single-purpose PR for self-locate on current main would be welcome.

PR #601, opened 2026-09-15
i18n foundation: locale core, parity gates, English-only string extraction

meirm:
## Summary

Foundation PR for interface localization, per the discussion on #312: the locale subsystem, parity gates, and a complete **English-only extraction** of all application-owned presentation strings. This PR changes no user-visible behavior — English output is byte-identical to `main`, enforced by tests. Locales ship as stacked follow-ups (#-es is the first, built on this branch); nothing here depends on them.

Following the #412 notes: there is **no automatic DOM translator**. Keys are static and semantic (`cockpit.readout.groundSpeed`), lookups are deterministic, and locale switching (when a locale ships) is a full reload — so first-observed-text caching and stale-async-translation states cannot occur by construction.

## What changes

- **`src/i18n/` core** — `locale.js` (normalization, `?lang=` → stored preference (`gev:locale:v1`) → browser language → default; guarded storage; `<html lang/dir>`), `index.js` (`t()` with `{named}` interpolation and `Intl.PluralRules` selection, Intl formatters, `applyDocumentTranslations()` for static markup, hash-preserving reload). Zero new dependencies.
- **English catalogs** — 912 keys across four append-only namespaces (`shell`/`cockpit`/`layers`/`setup`), semantically keyed, never renamed.
- **Extraction** — every user-facing English string at its final presentation boundary: `t()` call sites across the UI/layer/voice modules plus `data-i18n*` attributes across the HTML templates. Machine contracts (layer IDs, enums, share-link/hash parameters, API fields, voice tool schemas, provider names, attribution) are untouched by design; the keep-English boundary is documented in `docs/TRANSLATORS.md`.
- **Gates** — `catalog.test.mjs` (strict key/placeholder/plural-variant parity, generalized to any shipped locale set; CI fails on a missing translation), `markupCoverage.test.mjs` (every `data-i18n*` attribute resolves), `i18n.test.mjs` (resolution precedence, guards, fallback, interpolation, plurals, DOM application), `repairPass.test.mjs` (English byte-identity anchors).
- **Selector machinery** — the dock renders one button per *shipped* locale; with only English shipped this is a single EN button. Configurable via `GEV_DEFAULT_LOCALE`/`GEV_SECONDARY_LOCALE` (validated against shipped catalogs; unshipped values degenerate safely).
- **Docs** — `docs/TRANSLATORS.md` is the deliberate-design artifact: glossary conventions, keep-English boundary, and the add-a-locale recipe each follow-up PR executes.

Also fixes one latent markup bug found during extraction: duplicate `data-i18n` attributes on the first-run choice tiles wiped the sub-copy markup in **every** locale including English.

## How to test

```sh
node --test src/i18n/   # 40/40
npm test                # 5 known environmental failures + 1 Windows skip (all reproduce on pristine main)
npm run build           # green
```

Behavioral check: run the app — everything renders exactly as on `main` (the extraction is a pure refactor of string sourcing). `?lang=xx` for unshipped codes is ignored; stored preferences outside the shipped set fall through to English.

## Reviewer notes

- One sanctioned test-form change family: source-reading tests that pinned English literals in place now pin the `t()` call form — same strings, same surfaces, source-shape updated to the extraction.
- Commit structure: `core → extraction → docs` — three commits, squash-friendly.
- Non-goals: locales (stacked PRs), RTL (the `dir` machinery exists), locale-aware number formatting (`toLocaleString` sites documented as deferred).

Stacked follow-up: **#-es** (Spanish) is open against this branch — its diff is exactly the four `es` catalogs, the es plural test pins, and one measured es-scoped CSS wrap rule.


kenshinice-ai:
Not a review of this PR's architecture — just field notes from having taken the
interface all the way through one language, in case they are useful while the
locale core is still being shaped.

I maintain an out-of-tree Simplified Chinese layer for my own use. I'm aware
from #412 and #312 that an automatic DOM translator is not the path being
selected, and I'm not proposing one here. But whichever catalog design lands,
these four places will bite, because they are not "a string in a template":

**1. Three labels use their own text as animation state.**
`setSplitFlapText` treats `element.textContent` as the settled value
(`src/splitFlap.js:456`), and `settle()` bails when it no longer matches
(`src/splitFlap.js:424`). Any layer that changes the text of
`#global-loading-label`, `#traffic-sync-label` or `#cctv-sync-label` after the
app sets it will re-trigger the cascade on every ticker cycle. A catalog that
resolves the string *before* `setSplitFlapText` is called is fine; one that
rewrites afterwards is not.

**2. Two summaries are typed in character by character.**
`src/hud.js:674` and `src/ui/cctvPresentation.js:213` both do
`textContent = text.slice(0, i)` on an interval, so every frame is a different
string. Same rule: localize the source string, not the element.

**3. The HUD summary is composed from DOM text.**
`src/hud.js:630` reads `#hud-mode`'s `textContent` back to build the summary
line. Localize the mode chip without localizing the composition and you get one
translated word in front of a line of English telemetry.

**4. Four Material Symbols ligature names are also real UI labels.**
The icon font renders the element's *text*, so `draw`, `normal`, `radio` and
`adjust` must never be looked up as copy — while `Draw`, `Normal`, `RADIO` and
`ADJUST` are genuine labels a few lines away. `display-controls.html:94-95` is
the clearest case: the ligature and the label are adjacent lines in the same
component. Any case-insensitive key lookup will turn an icon into a word, and
`index.html`'s `icon_names` list is the authoritative set to exclude.

One non-obvious presentation note: the console's labels carry 1.4–1.6px / 0.16em
tracking (e.g. `src/ui/styles/layers.css:323`, `command-dock.css:130`). That
reads as precision in Latin small caps and as broken spacing in CJK, where every
glyph already has its own side bearings. A locale probably needs a tracking hook,
not just strings.

Happy to contribute the Simplified Chinese terminology once a catalog shape is
settled — ~400 reviewed entries plus the set of strings that should deliberately
*stay* English (MGRS/GSD/NIIRS-style imagery-intel abbreviations, unit symbols,
provider names and map attribution). That data is architecture-independent, so
just say the word and the format.

bilawalsidhu:
The direction is welcome: a catalog-based locale core with English as the reference. Please rebase onto current main (several files have been split since, including `applicationShell.js` and `realtimeController.js`) and we'll review. #602 can follow once this lands.

PR #605, opened 2026-09-15
feat (cctv): FL511 Florida statewide camera pack (stills)

Danielslay86:
Adds Florida to the CCTV mesh — FL511 (Florida DOT) statewide traffic cameras, stills.

## Source

Keyless FL511 ArcGIS FeatureServer (`services.arcgis.com/.../FL511_Traffic_Cameras`). One of the clean tier-1 catalog formats: JSON query, structured fields (ID, DESCRIPT, COUNTY, DIRECTION, LAT/LON, IMAGE). The statewide set is ~4,000 cameras, past the service's 2000-row page ceiling, so the loader pages on `resultOffset` until a short page ends it (hard-capped at 10 pages so a misbehaving service can't loop forever).

## How it fits the existing pattern

Mirrors the other open-data packs (DriveBC, NSW, Calgary): fetch catalog, validate coords, host-pin the image URL, normalize, distance-prioritize to a cap. Two image hosts (`images-dis.divas.cloud`, `snapshots.divas.cloud`) are allowlisted; a row whose IMAGE resolves elsewhere is dropped, same SSRF pin as the other packs.

`DIRECTION` is a real camera bearing here (N/S/E/W plus "NOT DIRECTIONAL"), so `directionToHeading` is trusted for the cardinals; "NOT DIRECTIONAL" falls through to the id-hash fallback like the headingless packs.

## Config

- `CCTV_FLORIDA_ENABLED=0` disables the pack.
- `CCTV_FLORIDA_MAX_SOURCES` caps it (default 500, anchored one point per region across the state so the cap spreads statewide rather than piling into Miami). Raise it toward ~4,000 to load the whole state.

## Not in this PR

Stills only. FL511's cameras also publish HLS video, but the stream URLs aren't exposed on this FeatureServer, so that's a possible follow-up if a stream endpoint is confirmed.

## Testing

Loaded the pack, confirmed 4,057 cameras register and the default cap of 500 spreads across the state (spot-checked Miami and Pensacola — both ends have cameras).


bilawalsidhu:
We'd like Florida in as a camera pack. FL511's terms limit its content to individual, non-commercial use, so please note that in the DATA_SOURCES row. To match the other packs: rebase onto current main, cap the catalog read, refuse redirects, add a Florida bounds check, a `dataCredits.js` entry, `.env.example` settings, unit tests and a CHANGELOG line.

Danielslay86:
Thanks. On it! I'll tag when complete.

Danielslay86:
@bilawalsidhu Ready for another look. Went through your list:

- Rebased onto current main
- Capped catalog read (2 MiB per page) and redirects refused
- Florida bounds check
- dataCredits.js entry, .env.example settings, CHANGELOG line
- DATA_SOURCES row and note carry FL511's individual, non-commercial terms
- Unit tests (11), including catalog wiring and the kill switch

Two bugs turned up along the way, both fixed: FL511 reuses its ID column across regions, which was collapsing 4,057 cameras to 2,717, so ids now come from the image channel; and the one-letter direction codes never parsed, so every camera had a fallback heading. The updated description has the details and gate results.

PR #618, opened 2026-09-16
feat(weather): thirty selectable layers on one globe, metered on purpose

monwiss:
Adds an optional **Weather** layer drawing Vaisala Xweather raster tiles, and a
`WEATHER` panel in the right rail that decides which of them draw and when they
are allowed to cost anything.

The globe gains observed radar, lightning, severe-weather warnings, tropical
cyclones, storm cells, surface analysis and a dozen analysis fields. Radar alone
is on by default, so the globe is unchanged until someone asks for more.

Branched from and merged up to `main` at `a65d9d8`.

## Why Xweather, and not NOAA / ECCC

**On the key first, because it is easy to mis-read.** Xweather is BYOK freemium:
15,000 accesses a month free, no credit card to obtain one. That is the same
shape as keys this project already carries — Live Vessels and Active Fires both
need one and have no keyless path either, and the flagship photorealistic globe
through the direct Google route needs *billing enabled*, which this README
already says. Against that, a free-tier key with no card is not the heaviest ask
here. What is genuinely new is the metering past the allowance, which is exactly
what the WEATHER panel exists to keep deliberate.

The free stack was not skipped, either — it was **built first, shipped in
sixteen commits on this branch, and then deleted**. The history is in this PR;
what follows is what measuring it showed.

### One free source could not do it

No single free source is global, observed and current at once, so the free
design was a ladder, each rung covering the one below:

| source | grid | runs | observed? | reaches |
|---|---|---|---|---|
| ECCC GDPS | 15 km | 2×/day | no | the globe |
| ECCC RDPS | 10 km | 4×/day | no | North America, north Atlantic, Scandinavia, Arctic |
| NOAA/NWS MRMS → IEM NEXRAD | ~1 km | 5 min | **yes** | United States |

Radar is the only honest one and covers a few percent of the planet. The models
fill the rest, and RDPS existed only because it is both finer *and* fresher than
GDPS where it reaches — a +3 h forecast from the 18Z run instead of +9 h from
12Z.

### Stitching them produced one incoherent globe

Three sources on one sphere meant three of everything, and they disagreed:

- **Coverage never closed.** RDPS advertises the whole northern hemisphere,
  which is the *envelope of a rotated grid* rather than its domain — believing
  it would have put empty tiles over most of the planet. Its real footprint had
  to be mapped by probing a field that is never blank where the model runs.
- **Different gradients per zone.** GeoMet's default style buckets the field
  into eight classes: measured over one 5,000 km extent, 8 distinct colours and
  a 42 km mean colour run, so 15 km cells merged into flat plateaus. Forcing
  `PRECIPPRTMMH-LINEAR` gave 714 colours and an 18 km run — the native grid —
  but that is one source tuned to match another, per source, forever.
- **Visible seams.** The CONUS cutout that stopped modelled rain reading as
  observed ran its southern edge at 24 N, just below Miami, so zoomed-in views
  there showed the model resume along a straight line, with holes wherever radar
  was silent inside the box.
- **Different refresh moments.** Radar inherited the model's hourly tick, so a
  five-minute product sat up to an hour stale; each tier ended up with its own
  cadence, which means adjacent regions of the same globe show different
  moments. Each also needed its own alpha to sit consistently over the basemap.

Every one of those was fixable, and each was fixed. The result was still a globe
assembled from parts that did not agree.

### And the parts themselves were forecasts

The defect that ended it was not rendering. GDPS runs twice a day and takes
about five hours to publish, so **the field valid *now* is a +3 h to +15 h
forecast — measured at +15 h**; RDPS is +3 h to +9 h. Both already pinned the
newest run available, so no layer work could close it: a forecast half a day old
and a live radar sweep disagree about where the rain is, which is exactly what
showed on zooming in.

The obvious escape was checked rather than assumed. **NASA IMERG** is global
*and* observed, but 7 h 20 m stale — rain travels 300–500 km in that time, so it
places worse than the model it would replace.

### Even as a precipitation source alone, it is the better one

Set aside the other twenty-nine layers and compare like with like. Against the
ECCC ladder, `radar-global` is finer (~1 km observed against a 15 km and a 10 km
model grid), global at that detail rather than only where a regional model
reaches, and coherent across zoom — one source, one palette, one moment, no
seams and no cutouts.

Its one real drawback is ours rather than the vendor's: upstream refreshes every
two minutes, but the default refresh is once a day to stay inside the free
allowance, so what you see is an observation from the last refresh. Even then it
compares well — a day-old *observation* of where rain was is a different and
smaller error than a *forecast* of where rain will be in three to fifteen hours,
and the panel's Refresh button closes it on demand for the cost of one tile set.

### What one observed source bought

Xweather `radar-global` is observed, global (ground radar with satellite fill
between networks) and refreshed upstream every two minutes. It replaced all four
placements, and took the stitching machinery with it: the rectangle covers, the
complementary-cover arithmetic, the cutouts, the zoom bands, the partition rule,
a layer-scoped WMS capabilities parser (the unfiltered GeoMet document is
~39 MB) and the concurrent multi-source reads. One palette, one refresh moment,
no seams.

Scope note: the layer is optional and off by default, the key is BYOK with a
free 15,000 accesses a month, and without one the row says so rather than
degrading quietly. Nothing else in the app changes behaviour because of it. If a
mandatory-key layer is outside what this project wants to carry, that is a
reasonable call and better made now than after review — the reasoning above is
offered so it can be made on the evidence.

## Why the panel exists

Every tile is billed against a monthly allowance, and the bill scales with **how
many layers are enabled, not how often they refresh** — each enabled layer
re-fetches on every camera move. One exploratory session of a single layer
fetched 405 tiles, about 2.7% of a month.

So the feature is a control surface, not a layer list:

- Layers are opted into one at a time; offering one is free, enabling one spends.
- Refresh is a button. A timer is available and off until chosen.
- The running monthly total sits beside the controls that move it.
- The layer set was chosen so that a daily refresh of everything enabled, for a
  whole month, fits inside the free 15,000.

Every layer offered bills at the base rate. The vendor also sells lightning and
air-quality products at five and ten times that; each has a base-rate equivalent
that is as good or better, so the surcharged variants are excluded and a test
names them.

## Notable decisions

**The proxy is mandatory, not an optimisation.** Xweather authenticates by
putting *both* halves of the credential in the tile URL path, so the browser can
never be allowed to build that URL. It fetches same-origin
`/api/xweather/tile/{layer}/{z}/{x}/{y}.png`. The layer name arrives from the
browser, so it is checked against the catalogue before the key is read —
forwarded on trust this would be an open proxy to every product on the account.
Log lines from that proxy are masked, because an upstream failure quoting the
request it attempted would otherwise write the secret to disk.

**One catalogue, three consumers.** `src/data/xweatherCatalogue.js` is the only
list of drawable layers. The Node provider derives its allowlist and each
layer's zoom ceiling from it; the browser layer derives its spec table; the
panel derives its labels and grouping. They cannot disagree.

**Fields are single-select.** Continuous fields paint every pixel (24–160 KB a
tile against an overlay's 116 bytes), so two would bill for one visible layer.
Sparse overlays stack freely above them, held there by paint rungs rather than
by refresh order.

**Zoom ceiling is a property of the layer.** Radar and the continuous fields are
sampled rasters and stop at level 9, where the service starts upsampling. The
symbol layers — lightning, warning polygons, storm cells, cyclone tracks — are
redrawn at each level at constant screen size and run to 12, because magnifying
a symbol scales the symbol rather than blurring a field.

**Coverage is on the row.** A layer drawing nothing over Europe because its
source stops at a border is indistinguishable, on screen, from a broken one.
Each such layer carries its coverage on the chip, checked against the coverage
Vaisala publishes, and forecasts are labelled as forecasts.

**Refresh means refresh.** Tiles carry a freshness floor — the moment new data
was last asked for — so a cached tile older than that is refetched. The cache
TTL is tuned for spend and is far longer than the weather stays still; without
the floor the button would re-request every tile and be served the same bytes.

**The panel follows its layer.** It is the layer's control surface, so it is
absent while the layer is off and opens with it when switched on, and the layer
stays *on* in the two states it is designed to explain — no key, and a photoreal
stack hiding the globe — rather than failing to enable and taking the
explanation with it.

**It does not draw in 3D, and that was a decision.** A photorealistic stack
hides the globe and every imagery layer with it. Draping the tiles on the 3D
tileset instead was built and measured before being withdrawn: Cesium picks the
draped level per model primitive rather than from the camera, so sharpness
oscillates as you descend and magnification turns lightning glyphs into blobs.
`docs/KNOWN-ISSUES.md` records the measurements, so the next person to notice
finds the answer rather than rebuilding it.

## Security notes for review

- **No credential reaches the client.** Neither half appears anywhere in `dist/`
  after a build; the browser only ever sees the relative `/api/xweather/*` route.
  Pinned by a test on the tile template.
- **One new outbound host**, `https://maps.api.xweather.com`, fixed in the
  provider — not composed from any client input. The layer segment is validated
  against the catalogue allowlist *before* the credential is read, so a probe
  for an unlisted product is a 400 and never a billable fetch.
- **Logs are scrubbed.** The credential sits in the URL path, so an upstream
  error quoting its own request would otherwise write the secret to disk. A test
  asserts the mask holds while keeping the message useful.
- **Supply chain shrinks**: one devDependency swapped (`vite-plugin-cesium` →
  `rollup-plugin-external-globals`), 3 packages added, 31 removed, net **−28**.

## Also in this PR

Two changes that are not the Weather layer, called out so they can be judged
separately:

- **`vite-plugin-cesium` replaced with an in-repo plugin** (`build/cesium.js`),
  dropping the deprecated `sourcemap-codec` chain while producing byte-identical
  build output. The upstream plugin is unmaintained and the layer needed a
  working Cesium build to land on. Happy to split this out if preferred.
- **The provider disk cache and the voice debug log are excluded from the dev
  server's file watcher.** The tile proxies write a file per tile; watching that
  directory costs a cold tile 3–9 s instead of ~150 ms, because the watcher's
  stat calls take the libuv threads `getaddrinfo` needs. Development only —
  there is no watcher in preview or production — but it affected every
  disk-cached provider.

## Without a key

There is no keyless mode for this layer, unlike every other source here. The row
reports `UNAVAILABLE · ADD XWEATHER KEY` and draws nothing, rather than showing
an empty globe that reads as "no weather anywhere".

## Verification

Run on the merged tree, at `main` `a65d9d8`:

| Gate | Result |
|---|---|
| `npm test` | 4217 tests, 4207 pass, 0 fail |
| `npm run test:track` | 109 pass, 0 fail — no console errors, no 5xx |
| `npm run build` | clean |
| `npm run format:check` | 936 source files clean |
| `npm run check:boundaries` | clean |
| `scripts/qa-l9-matrix.mjs` | B10b/B10c/B11b cover tiles, accounting, the allowlist refusal and the keyless path |

Exercised against the live API: layer selection and single-select fields; share
links round-tripping a multi-layer selection; the budget counter moving only
when a fetch actually happens; refresh refetching (23 requests, 23 `MISS`, +23)
where a cached pass is 23 `HIT`, +0; two dev servers on one checkout summing
their spend rather than overwriting it; the panel appearing and disappearing
with its layer row; `UNAVAILABLE · GLOBE HIDDEN IN 3D` under a photoreal stack.

Zoom ceilings were measured rather than assumed, by comparing the same ground at
levels 9 and 12 for both a sampled raster and a symbol layer.

## Merge notes

One collision with upstream: share-link token `n` is now the Directions layer,
so Weather takes `k`. Nothing had shipped with `n`. The registry holds 22 layers
on 22 distinct tokens.

Three pinned Realtime tool digests are re-derived, deliberately — `weather`
joins the two layer menus in `actionSchemas.js` and the common-name mapping in
`toolDescriptions.js`. The serialized schema length is unchanged at 26,257.

## Docs

`CHANGELOG.md`, `README.md`, `DATA_SOURCES.md` (license and attribution),
`docs/CURRENT-STATE.md`, `docs/CODE-BOUNDARIES.md` (new package exports),
`SECURITY.md` (allowlist, credential scrubbing, the LAN quota list) and
`.env.example`.


bilawalsidhu:
Thanks @monwiss, and sorry for the wait. The server side here is excellent: the key never reaches the browser, the metering is real, and the allowlist and tests are careful.

Since this opened, #617 landed its own weather layers (radar, satellite clouds, lightning, wind, cyclones), a WEATHER panel in the right rail and a shared history timeline. As written, this now builds a second weather system beside that one (the `weather` id, token `k` and `#weather-panel` all collide). We'd like to merge Xweather as an optional keyed source inside that structure instead:

1. With a key present, the Rain radar and Lightning cards gain a `Global (Xweather)` choice, plus one new warnings card. Start with those three products; more fields can follow.
2. Render through the existing weather layer's imagery host and 3D Tiles shells (`src/layers/weather/`) so it also works on Google 3D, and drive frames from the shared clock (`src/layers/weather/clock.js`), using absolute times if the tile API accepts them.
3. Keep your budget governor, and show the monthly count as a line on the card rather than a separate panel. Please add a per-client rate limit and a concurrency cap before paid fetches, and cap tile bodies with `readResponseBytesCapped`.
4. Split these out as their own small PRs: the dev-server watcher exclusion for `.gev-cache`/`.gev-logs` (worth having on its own), the `tileBudget.js` extraction, and the Cesium build-plugin swap.
5. Leave the share-link token to us (#658).

Happy to review the pieces as they come.

bilawalsidhu:
Thanks for the careful work here, especially the tile-budget governor. The keyless weather stack landed in #617, so this one is superseded. If you'd like to bring Xweather in as an optional keyed provider on top of #617's provider setup, a small follow-up PR would be welcome.

monwiss:
Thank you @bilawalsidhu for your project and the review of the PR! I'm on it with your suggestions, they should land shortly!

PR #635, opened 2026-09-16
ALPR analyst queries, cordon analysis, and a Live Trains layer

peoplezoo:
## What this adds

Three additions to the surveillance-awareness and movement layers, each its own commit and independently revertable. Happy to split into separate PRs if you'd prefer to review them individually — say the word.

### 1. ALPR: analyst queries + corrections loop (`81a4447`)
- `alpr-cameras` joins `ANALYST_LAYERS` and the `analyst_query` layers enum, so voice can answer "how many plate readers within 5 km?" / "show the Flock cameras in view" (filter hint: `manufacturer contains flock`). Counts announce the viewport-loaded caveat via the existing coverage-note path (`VIEWPORT_LOADED_LAYERS` generalized past flights).
- New **FIX ON OSM** row chip opens the selected camera's OSM node, so corrections flow back to the shared dataset instead of a private copy.
- Camera cards surface mapped `surveillance:zone`, `ref`, and `check_date`/`survey:date`.

### 2. ALPR: cordon analysis (`76ccffb`)
A **CORDON** row chip measures what share of the roads entering the town at view center pass a mapped reader, and shows where the gaps are. The admin boundary (`is_in` → area pivot, your proxy's 30-day boundary cache class) induces a cut of the road graph: every drivable way crossing the ring is an entry gate; split carriageways merge; a gate is covered when a mapped camera sits within 100 m of its road polyline. Gates render on the globe (coral = unmonitored, labeled), and the legend reports covered/total + majors — always with the mapped-cameras-are-a-floor caveat and a truncation warning when either fetch hits its cap. Crossing detection intersects segments with the ring directly, so a long sparse segment passing clear through a small town still counts. Pure math in `cordon.js` (Cesium-free, pinned in Node tests); fetch adapter and overlay renderer separated per the layer's owner split. No `poly:`/area-scans — bbox + `is_in`/pivot only, per the proxy validator.

### 3. Live Trains layer (`f9344cc`, motion in `fa80360`)
Keyless, CORS-open Amtrak GPS via the Amtraker community API, polled once a minute. Trains dead-reckon every frame along the line their last two *real* fixes define (no compass guessing), with carry-forward clamped to two minutes so quiet GPS leaves a dot resting near where the train was genuinely last seen. Holds the continuous render loop while enabled (aircraft-layer pattern), entities persist across polls, lateness uses the operator's own color (validated hex only), Predeparture trains are dropped so origin stations never paint phantom positions, and a malformed payload is a failure — never an authoritative empty map. Registered across catalog, panel, share-state registry (token `l`), credits, `DATA_SOURCES.md`, package boundaries, and both voice layer enums with spoken aliases.

## Validation
- Full suite: **4,161 tests, 0 failures** (25 new tests across `alprCordon.test.mjs`, `alprCameras.test.mjs`, `liveTrains.test.mjs`)
- `npm run check:boundaries` and `npm run format:check` clean; new modules registered in `scripts/package-boundaries.json`
- Voice schema pins re-derived per the in-test procedure for the deliberate enum/wording changes (`firstRunExperience`, `actionSchemas`, `radioMarkup` pins), with the changed tools named in the pin comments


bilawalsidhu:
Decision: yes to the ALPR analyst queries and to Live Trains; we'll pass on the cordon coverage verdicts. Could you split this into two PRs: ALPR queries, and Live Trains as a transit feed adapter? Note that token `l` is taken (tokens are being sorted out in #658). Thanks!

PR #637, opened 2026-09-16
Pattern Watch: circling-aircraft detection

peoplezoo:
## Pattern Watch — circling-aircraft detection

Adds a derived layer (no new network source) that reads the records the Live Flights and Military layers already hold, keeps a rolling 12-minute track per airborne aircraft, and flags the loitering signature: accumulated **signed** heading change of ≥2 same-direction revolutions over ≥4 minutes, whole track inside an 8 km circle. Displacements under 150 m carry no bearing, so GPS jitter on a slow aircraft can't sum into a fake orbit; straight flight cancels out. Flags render as a coral ground ring under the orbit with callsign, turn count, and duration. The legend states the ethics where the flags appear: a flag describes the flight path, never the mission — training, traffic watch, survey work, and airport holding stacks all circle.

### Scope
Rebased directly onto current `main`; this PR is a single commit containing only the Pattern Watch work. The ALPR/cordon and Live Trains changes are no longer part of this branch.

### Validation
Full suite green; 6 new tests pin the circling math (orbits flag either direction, straight/1.5-turn paths don't, jitter yields zero segments) and the derived-layer lifecycle. Boundaries + format clean; voice layer enums extended with schema pins re-derived against current main, and the Pattern Watch modules registered under the refactored package boundaries.


bilawalsidhu:
Hey @peoplezoo — this is awesome. Pattern Watch is exactly the kind of derived intelligence capability we want to encourage: it creates a genuinely new behavior from data already in the app, and the language distinguishing an observed flight pattern from inferred intent is especially strong.

We’ve just completed a substantial refactor on `main`, and this PR is currently stacked on #635. Would you be willing to rebase it directly onto current `main` and make the PR contain only the Pattern Watch work? That would let us review the feature cleanly without coupling it to the ALPR/cordon/Live Trains changes.

Please preserve the focused scope and the ethics/limitations language. Once it’s rebased and the gates are green, tag us here and we’ll take another look. Thank you for building this — it’s a very cool contribution.

peoplezoo:
Hello bilawalsidhu
.
Yeah I’ll get on it right now . Glad you like it
This is awesome!

peoplezoo:
@bilawalsidhu

 Rebased onto current main — the PR now contains only the Pattern Watch commit; the ALPR/cordon and Live Trains changes are no longer part of this branch.

Notes from the rebase:
- Re-derived the three schema pins (tool-schema length/sha, the two Realtime digests) against the current main baseline, and bumped the layer-count assertions 21 → 22.
- Registered the Pattern Watch modules under the refactored package boundaries (`application-components` and `application-layer-construction` ownership lists, plus its own `pattern-watch` group).
- The ethics/limitations language in the legend and module docs is preserved unchanged.

All gates green locally (unit suite, `check:boundaries`, `format:check`). Ready for another look — thanks!

peoplezoo:
@bilawalsidhu/gods-eye-view ***@***.***>  its ready
for your review, I also built a feat: convergence (rendezvous) detection
<https://github.com/bilawalsidhu/gods-eye-view/pull/681>

peoplezoo (commented):
@bilawalsidhu  all green

PR #649, opened 2026-09-17
fix: treat AIS "not available" speed and course as unknown

theerayutbo:
## What

AIS encodes speed over ground in 0.1-knot units and course over ground in 0.1-degree units, reserving the top code of each field for "not available" (ITU-R M.1371). SOG 1023 arrives from AISStream as `102.3` and COG 3600 as `360`. Both are finite numbers, so `numberValue` stored them as ordinary readings and every consumer — vessel cards, the HUD, voice analyst queries — treated a missing measurement as a real one.

`normalizedHeading` already does exactly this for true heading's own sentinel (511). This applies the same treatment to SOG and COG, in the same place, so the sentinels never enter the cache.

## Evidence

On a live AISStream feed of 3,062 vessels:

- exactly **3** records at `102.3` knots, and **no reading anywhere between 40 and 102.3**
- **404** records at exactly `360` degrees
- one Dutch rescue boat moored at its station (`KNRM ARBOR VITAE`, MMSI 244060365) carried both sentinels at once

An analyst summary over that snapshot reported a maximum speed of 102.3 knots — faster than any vessel afloat. After the change, on a live feed of 2,088 vessels: 0 sentinels, max speed 66.4 kn, max course 359.9°.

## How verified

- `npm test` — 4,150 passing, 0 failing, including 5 new tests in `src/data/aisStreamSentinels.test.mjs`
- `npm run build` — clean
- a live AISStream feed, before and after

The new tests cover both sentinels plus three guards that fail if the filter is too aggressive: a stopped vessel's `0`, and the highest values the fields can actually encode (102.2 kn and 359.9°). The tests were written first and observed failing against the old behaviour.

The test file lives under `src/` rather than next to the module, because `scripts/run-unit-tests.mjs` only discovers tests under `src/`; a test placed in `server/` is never executed by `npm test`.

## On `npm run test:track`

I could not establish a reliable baseline with this harness on my machine, so I am reporting what I saw rather than claiming a pass.

Three of six runs on this branch, and **two of three runs on unmodified `main`**, ended in a puppeteer harness error (`TargetCloseError`, `Execution context was destroyed`) before completing. Of the runs that finished: `main` gave 109/0 once; this branch gave 107/2 and 105/4.

Every failure on this branch was the same frame-capture gate — `frames=27/30` and `frames=22/30` with `timedOut=true` — plus the assertion derived from it (`trackedReadyFramesComplete && …`), whose own measured value was the correct `0`. No value, height or sampling-count assertion failed in any run.

The changed code path is the server's AIS websocket ingest, which this harness bypasses by stubbing `/api/ais-live` to a zero-row snapshot. I am running under software rendering (`--use-gl=angle --use-angle=swiftshader`) on a laptop, and a `test:track` result from a machine with hardware GL would be welcome.


samehkhamis:

PR #655, opened 2026-09-18
fix(traffic): prioritize on-screen roads in the dot budget

daikaginza:
Closes the first "next iteration candidate" under *Street traffic can be slow/uneven when panning across dense city blocks* in `docs/KNOWN-ISSUES.md`: prioritize currently visible road segments inside the active viewport before off-center segments.

## Changes

`MAX_DOTS` is a hard cap and every dot costs the same per-frame animation work whether or not it is on screen, but allocation spread that cap evenly across the whole fetched tile. Roads are fetched in a box centered on the look-at ground point (the C4 fix) while the frame the user sees is the camera rectangle, and at an oblique pitch the two are offset: measured on the running app at 650 m / −30° over midtown, 47 % of the fetched roads lay outside the camera rectangle with the cap already exhausted, and only 36.8 % of the dots landed on canvas. `allocateRoadDotBudgets` now takes the live camera rectangle — the one from `getViewBounds()` at render time, not the one the fetch was centered on, since a render can follow a cached tile or a late flow response. Roads whose bounding box meets that rectangle, grown by a margin ring, reach their ideal dot count first; the rest take every dot the visible tier leaves. Passes 2 and 3 are unchanged in kind and simply run per tier.

The ring is floored at `MIN_CENTER_SHIFT_KM`: `onCameraChanged` skips reloading — and therefore re-allocating — for any pan shorter than that, and a purely proportional ring is narrower than that pan on a small rectangle, which is exactly where the cap binds hardest. `VIEWPORT_PRIORITY_MARGIN` has to stay well under 0.5 for the same reason in reverse; at 0.4 the ring swallowed the whole tile in that measurement and the pass became a no-op.

The fairness *seed* does change, and I would rather state it than let it be found later: under a cap too small to give every road one dot, the visible roads are seeded first, so a road outside the frame can drop from one dot to none where it previously kept one. That is the intended trade, `docs/KNOWN-ISSUES.md` records it, and `model.test.mjs` pins it rather than hiding it. With no viewer, or a rectangle that cannot be used — a zero span, or the wrapped form Cesium reports across the antimeridian — allocation falls back to the previous demand-only order, byte for byte.

## How it was verified

- Gates, on the pinned CI matrix rather than only my machine: Node 24.14.0 and 26.x are both green. `doctor -- --json` reports `"ready": true`, `format:check` covers 918 files, `check:boundaries` and `build` pass, and `npm test` is 4170 tests / 0 failing / 1 pre-existing skip — under `GEV_REQUIRE_ALLOCATION_GATE=1`, so the Node-24-calibrated allocation probes ran as hard failures instead of being skipped. `npm run test:track` is 109 passed, 0 failed locally; it needs a browser, so CI does not run it.
- New `src/layers/traffic/model.test.mjs`, 11 tests, including one that runs the real Overpass parse rather than hand-built roads. Every one of the 11 is proven able to fail — I ran 12 mutations against the real source; 11 were killed by the specific test that claims to guard them, and the twelfth is recorded as an equivalent mutant with the argument for why nothing can kill it.
- Randomized property check, 50 000 allocations (48 251 with the cap binding): never exceeds the cap or a road's ideal; an uncapped budget reaches every road exactly; the total is conserved; the visible tier never comes away with fewer dots than before; and **with no rectangle the output is byte-identical to the previous algorithm**. The largest per-road difference inside a tier is 1 dot, from largest-remainder rounding.
- Cost is +1.5–2.3 ms per allocation at 2520 roads × 80 coordinates. There is one call site and it sits on the load path, not in the `preRender` loop.
- `scripts/qa-traffic-viewport-budget.mjs` drives the running app against a deterministic road fixture, reads every rendered dot back through the layer's own `getDetectableObjects()`, and classifies it against the live camera. It exits non-zero when its own preconditions fail — most importantly when no camera reached the cap, since allocation order cannot change such a capture. That is not hypothetical: the first reading I took was exactly that mistake, and it measured this change as a no-op.

| camera | at cap | on canvas before | after |
|---|---|---|---|
| **650 m, −30°, midtown** | **yes** | 36.8 % | **43.8 %** |
| 650 m, −35°, London | no | 40.0 % | 40.2 % |
| 1500 m, −55° | no | 55.1 % | 55.2 % |
| 1500 m, −78° | no | 74.1 % | 74.0 % |

Same 6000 dots, 419 more of them on canvas, and dots behind the camera fall 608 → 257. The three cameras below the cap move by at most 0.2 points against a measured re-run spread of 0.36 — unchanged, as intended.

Two bounds on that number, both stated in the harness header: occlusion is not tested, so the percentage is an upper bound on what is really visible; and the fixture grid is axis-aligned, where the allocator's bounding-box reachability test is very nearly exact, so real curving ways will land in the visible tier more often and the measured *gain* is an upper bound too. `--real-overpass` swaps in live road data at the cost of run-to-run comparability.

What I have not done is a human visual pass on a real GPU. Worth one before merge — and note that #656 makes that pass harder than it should be at exactly the 650 m oblique camera this change targets.

## Unrelated bug found while measuring (not fixed here)

`parseRoads` places an entire road from a single `scene.sampleHeight()` probe taken at its start, guarded only by `Number.isFinite`. When that probe lands before the tiles under it are loaded it returns a finite but absurd value, which passes the guard and plants every dot on that road kilometers underground.

This is not an artifact of my harness. On plain `main` (0d41b6b), with real Google 3D tiles and a real GPU (`ANGLE (NVIDIA Corporation, NVIDIA GB10/PCIe, OpenGL 4.5.0)`, `softwareRenderer: false`), arriving at midtown 650 m / −30° from above the activation ceiling:

```
midtown 650 m / -30:  layer reports 6000 dots | height min -16851 median -16829 max -16805 m | 6000 of 6000 below -1000 m
midtown 1500 m / -55: layer reports 2680 dots | height min   -31 median    -19 max     51 m |    0 of 2680 below -1000 m
```

The HUD says 6000 dots and the screen has none. `policy.js` already documents a milder form of the same single-sample root cause in `JAM_DOT_DEPTH_PUNCH` ("396 jam dots, zero visible"). The QA harness in this PR works around it by re-projecting each dot at ellipsoid height 0 and reporting `meanRenderedHeightM` so the contamination stays visible instead of being folded silently into the numbers.

I left it alone deliberately — `docs/MAINTAINER_WORKFLOW.md` asks contributions to avoid unrelated changes — and filed it separately as #656, with the console repro and the contrast case. Nothing in this PR depends on it, and nothing in this PR fixes it.


daikaginza:

On the root cause — this one is not the rate-limit family the other open reports share. It is a **mismatch between two shapes that were each individually correct**:

Roads are *fetched* in a square box centred on the look-at ground point. The frame is a *camera rectangle*. At a level pitch those two nearly agree, so nothing looks wrong. At an oblique pitch they stop agreeing badly: a large part of every fetched tile sits outside the frame, much of it behind the camera. The dot allocator was then doing exactly what it was written to do — spread `MAX_DOTS` fairly across the roads it was given — and fairness across the *tile* is the wrong objective when the budget is a *screen* budget. The result is a cap spent on geometry the user cannot see, and it is invisible in the code because neither half is wrong on its own.

So the change is not a new policy, it is giving the allocator the frame it was always implicitly serving: roads inside the camera rectangle plus a margin ring reach their ideal count first, and everything else takes what the visible tier leaves.

Three things I would put in front of a reviewer, because they are the parts I would want to argue with:

**There is a real trade, and it is in the fairness seed.** When the cap is too small to give every road one dot, the seed now goes to the visible roads — so a road *outside* the frame can drop to zero dots where it previously kept one. That is the intended direction, but it is a loss, not a free win, and it should be stated as one.

**The margin ring is floored at `MIN_CENTER_SHIFT_KM`, and that floor is load-bearing.** `onCameraChanged` skips reloading — and therefore skips re-allocating — for any pan shorter than that distance. A purely proportional ring is narrower than that pan on a small rectangle, which would leave a newly revealed edge holding nothing but seeds until something else triggered a reload.

**Nothing changes where there is no usable rectangle.** No viewer, a zero span, or the wrapped form Cesium reports across the antimeridian all take the previous path unchanged.

On measurement: `scripts/qa-traffic-viewport-budget.mjs` now **fails when no camera reaches the cap**, because allocation order cannot matter below the cap and a green run that never bound the budget would be measuring nothing. At the one camera where it does bind (650 m, −30° over midtown): **36.8% → 43.8%** of dots on canvas at the same 6,000 dots, and **608 → 257** dots behind the camera. The three cameras below the cap move by at most 0.2 points.

Full gates on the rebased tree: `npm test` 4191 tests / 0 fail, 13 allocation microbenchmarks under `GEV_REQUIRE_ALLOCATION_GATE=1`; `npm run test:track` 109 / 0; doctor, format:check, check:boundaries, build all clean.

PR #664, opened 2026-09-19
fix: render native select option lists in the dark palette

Udaydsmls:
## What

Native `<select>` option lists render in the platform's light palette, so the option text — which inherits `--text-primary` (`#e8eaed`) — comes out near-white on a white popup. Only the highlighted row is legible, because the browser forces its own highlight colours on that one.

Affects the HUD layout, Scenes, CCTV camera, Radio filter and Draw colour menus.

Screenshot 2026-09-18 171219

Screenshot 2026-09-18 171240

## Change

Declare `color-scheme: dark` on `select` and set explicit `option` / `optgroup` colours from a new `--menu-bg` token.

Declared on the controls rather than on `:root` deliberately: at root it would also reskin the scrollbars that `cctv.css`, `cockpit.css` and `bhote-koshi.css` tune explicitly.

Screenshot 2026-09-19 001706

Screenshot 2026-09-19 001714

## Verification

`npm test`, `npm run test:track`, `npm run build`, `npm run format:check` and `npm run check:boundaries` are green.

Computed styles on the running app confirm `color-scheme: dark` and the option colours on the affected selects. The popup is an OS-drawn window and does not appear in page screenshots, so it's worth opening a dropdown to eyeball it.


samehkhamis:

Issue #661, opened 2026-09-18
bug: Street Traffic reports road data unavailable even with working TomTom access


jimerb:
### What happened

The Street Traffic layer consistently ends in:

`UNAVAILABLE · OpenStreetMap · Road data temporarily unavailable`

<img width="720" height="277" alt="Image" src="https://github.com/user-attachments/assets/39fc0ced-1a4d-4459-85da-04452ce42b1a" />

This happens even when the camera is left completely still. It is reproducible in a city-level view, so it is not caused by panning or a city jump.

### What I verified

- The running app reports the TomTom provider as configured (`GET /api/tomtom/status` returns HTTP 200 with `hasKey: true`).
- A real TomTom traffic-flow tile request through the app returns HTTP 200 with data.
- The OpenStreetMap/Overpass road request fails independently in this environment:
  - `overpass-api.de` returns HTTP 406.
  - `lz4.overpass-api.de` returns HTTP 406.
  - `overpass.kumi.systems` and `overpass.private.coffee` time out.
- The local Overpass cache is empty, so there is no road snapshot available for fallback.

This indicates the TomTom credential is not the cause. Street Traffic needs OSM road geometry before it can match TomTom flow data to roads. When the Overpass mirrors refuse or time out, the layer becomes unavailable even though TomTom is responding.

The repository already documents slow or uneven street-traffic loading while panning in `docs/KNOWN-ISSUES.md`; **this report is more specific because the same failure occurs while stationary and the road-data dependency is independently failing.**

### Expected behavior

The layer should either recover through a healthy road-data source, clearly identify the Overpass provider failure and retry, or use a cached/local road snapshot so valid TomTom flow data can still be used. The current UI makes this look like a general Street Traffic or API-key failure and gives no actionable indication that the failing dependency is the public Overpass road-data service.

### Steps to reproduce

1. Start the local development server and open `http://localhost:4173`.
2. Zoom to a city-level view.
3. Enable the Street Traffic layer.
4. Leave the camera completely still and wait for loading to settle.
5. Observe `UNAVAILABLE · OpenStreetMap · Road data temporarily unavailable`.
6. Separately verify that `GET /api/tomtom/status` returns `hasKey: true` and that a TomTom flow tile request returns HTTP 200.

### How you installed

Terminal (git clone + npm)

### OS, Node version, browser

Windows; Node 24.19.0; Chrome (version unknown)

### Version or commit

0d41b6b

### Output of `npm run doctor`

```text
[...]

PR #665, opened 2026-09-19
fix(traffic): say which upstream declined a road load

daikaginza:
Refs #661 — deliberately not `Fixes`. This addresses the reporting half of that issue and leaves the fallback half open, so the issue should stay open after this merges.

## Changes

Street Traffic has two upstreams: OpenStreetMap for road geometry and TomTom for flow. When the road load failed, the row said `Road data temporarily unavailable` no matter what happened, so there was nothing in it to tell the two apart. The reporter of #661 checked their TomTom status endpoint and a live flow tile, both healthy, while the actual failure was the public Overpass mirrors declining their network with HTTP 406 — a status the code already had in hand and threw away at `src/layers/traffic/ingestion.js:70`.

- `roadRequestError(status)` in `src/layers/traffic/source.js` maps a refusal to the line the panel prints: `Overpass rate-limited` (429), `Overpass timed out` (504), or `Overpass refused the road query (HTTP <status>)`. A refusal whose code is unreadable — the layer takes its source by injection, so an adapter may not supply one — says `Overpass temporarily unavailable` rather than printing `HTTP undefined`. The first two are the words `alpr`'s Overpass source already uses — this is the existing vocabulary, not a new one. What the status adds is *which* upstream declined and *how*, which a single-source layer never needs to say.
- The thrown error is a `RoadRequestError` carrying `status` beside the message, following `LiveSourceError` in `src/sources/live/contract.js`, so a caller can branch on the code without parsing English.
- The catch in `loadRoadsForBounds` promotes a classified refusal and keeps `Road data temporarily unavailable` for everything else. That branch is on the error type, not on the text: a dropped socket throws whatever the platform felt like saying, and `Failed to fetch` on a panel row helps nobody.

Nothing about retries, backoff, caching or the mirror rotation changes. The proxy already rotates all four mirrors and only surfaces a refusal once none of them answered, so a status arriving in the browser means the proxy had nothing better to offer.

## How it was verified

Same fault, same instrument, once per side. The 406 is injected at the network boundary rather than waited for: the mirrors that refused my network on 18 September answer it again today, and a transient upstream ban is not something a reviewer can reproduce on demand. Everything below the injection point is the real stack — proxy, source, ingestion, controls, panel.

| | row rendered in the browser |
|---|---|
| before | `UNAVAILABLE · OpenStreetMap · Road data temporarily unavailable` |
| after | `UNAVAILABLE · OpenStreetMap · Overpass refused the road query (HTTP 406)` |

Read out of the DOM, not out of `getStats()`, because the claim is about what a person sees. At 420px the row wraps and the page does not scroll horizontally.

Before injecting anything I checked the premise the issue rests on: `overpass-api.de` and `lz4.overpass-api.de` returned 406 to this network for the application's User-Agent, for curl's default, and for a browser User-Agent alike, while `overpass.kumi.systems` and `overpass.private.coffee` timed out. So the refusal is not something the `OVERPASS_USER_AGENT` header provokes, and there was no header fix hiding here.

Tests:

- `navigation.test.mjs` drives the assembled layer with a source returning `{ok: false, status}` and asserts the row text for 406 and 429, plus a companion test that a plain thrown `Error` still yields the general line. Reinstating the old assignment turns the first red and leaves the second green, so it is the behaviour that is pinned and not the presence of a string.
- `source.test.mjs` covers the mapping, that `status` survives on the error object, and that an absent or unreadable code never reaches a row as `HTTP undefined`.

Gates on Node 24.14.0: `npm run build`, `npm test` (4148 tests, 0 failures, 1 pre-existing skip), `npm run test:track`, plus `npm run format:check` and `npm run check:boundaries`. The diff adds no console calls: the existing `[Data:Traffic] Fetch error:` warning still carries the underlying error object, unchanged.

## What this does not do

#661 also asks for a cached or local road snapshot so valid TomTom flow can still be drawn when Overpass is unreachable. That is a feature with its own design questions — which snapshot, how stale is too stale, what the row says while it is being used — and I have left it alone rather than smuggle it in behind a message change. This PR only makes the failure legible.


daikaginza:

On the root cause — this one is not the rate-limit family the other open reports share, it is a diagnosis failure:

The Street Traffic layer has **two independent upstreams**: OpenStreetMap, through `/api/overpass`, for road geometry, and TomTom for flow. Only one of them can be down at a time. But the failure path collapsed both into a single line — `Road data temporarily unavailable` — and that line is the wrong shape for the situation it describes. It reads like a configuration problem, and the only configuration this layer has is the TomTom key. So the reader goes and checks their key, which is the half that was working, and loses the afternoon to it. That is exactly the report in #661: *working TomTom access, still says road data unavailable.*

The real cause was upstream of the key entirely — the Overpass mirror refusing the query — and the panel had no way to say so.

So the change is not a new behaviour, it is the layer telling the truth about which half failed:

- `Overpass rate-limited` (429)
- `Overpass timed out` (504)
- `Overpass refused the road query (HTTP <code>)` for anything else
- `Overpass temporarily unavailable` when the adapter hands back a refusal with no status on it — because `HTTP undefined` on a panel row is worse than not naming a number

The status travels *beside* the message as well as inside it (`RoadRequestError.status`), so a caller can branch on the code without parsing English — the same shape `LiveSourceError` already uses in `src/sources/live/contract.js`. The first two strings match `alpr`'s Overpass source word for word; this is the project's existing vocabulary, not a new one.

Two things it deliberately does **not** claim:

- It does not say the road data is gone. Anything the browser did not classify keeps the original general line, because a raw platform `e.message` is not a sentence anyone should have to read off a panel.
- It does not say how many mirrors were tried. The proxy rotates mirrors behind `/api/overpass`; a refusal arriving in the browser means the proxy had nothing better to offer, and how it got there is not something this side knows. The message does not pretend otherwise.

Full gates on the rebased tree: `npm test` 4183 tests / 0 fail, 13 allocation microbenchmarks under `GEV_REQUIRE_ALLOCATION_GATE=1`; `npm run test:track` 109 / 0; doctor, format:check, check:boundaries, build all clean.

Issue #29, opened 2026-08-25
Harden server-registered CCTV URLs against SSRF


jsawyerdev:
Reviewed commit: 880a672b5e16ad3e41d318801d3a5203f9201923

Problem:
- CCTV source URLs can come from env JSON or a configured JSON file.
- The proxy accepts arbitrary `http:`/`https:` URLs from those registered sources.
- There is no private-address rejection, DNS pinning, or redirect hardening on these fetches.

Evidence:
- `vite.config.js:3493-3506` loads `CCTV_SOURCES_FILE`.
- `vite.config.js:3514-3522` loads `CCTV_SOURCES_JSON`.
- `vite.config.js:4074-4101` preserves `url` and `snapshotUrl` strings.
- `vite.config.js:4549` and `4374-4400` fetch registered HTTP(S) sources.

Suggested fix:
- Validate registered URLs at load time and fetch time.
- Reject loopback, link-local, private, multicast, and otherwise non-public destinations unless explicitly allowed.
- Resolve and pin DNS results for the actual connection target.
- Apply the same redirect policy used by the radio proxy.

This was found during an adversarial repository review; impact depends on how the dev/preview server is exposed.

Issue #22, opened 2026-08-25
Do not expose key-broker API routes from vite preview by default


jsawyerdev:
Reviewed commit: 880a672b5e16ad3e41d318801d3a5203f9201923

Problem:
- The OpenAI and Google proxy middleware is installed in both Vite dev server and Vite preview server.
- `vite preview` is easy to mistake for a deployable preview while it still brokers server API keys.

Evidence:
- `vite.config.js:5216-5223` registers OpenAI proxy middleware for dev and preview.
- `vite.config.js:5503-5510` registers Google proxy middleware for dev and preview.

Suggested fix:
- Require an explicit environment flag before enabling key-bearing routes in preview.
- Fail closed if required auth/rate-limit settings are absent.
- Keep local static preview separate from server-side API brokerage.

This was found during an adversarial repository review; impact depends on how the dev/preview server is exposed.

bilawalsidhu:
Consolidating into #16, which tracks access to the cost-bearing endpoints when the server is exposed beyond localhost.

PR #666, opened 2026-09-19
fix(cctv): aim the Street View fallback at the registered camera and meter it

### LINKED ISSUE #20: Rate-limit and validate Street View fallback in CCTV frame proxy

Problem:
- `/api/cctv/frame/:id` can fall back to Google Street View using client-supplied `lat`, `lon`, `heading`, `fov`, and `pitch`.
- There is no route-specific rate limit before calling Google Street View.
- Unknown camera IDs can still reach the fallback path when finite coordinates are supplied.

Evidence:
- `vite.config.js:4620-4628` accepts query parameters that override source coordinates.
- `vite.config.js:4459-4489` calls Google Street View using `GOOGLE_MAPS_API_KEY`.
- `vite.config.js:4653` invokes the fallback after upstream image failure.

Suggested fix:
- Apply mandatory quota controls before any Street View request.
- Require a known registered camera source for fallback, or remove client-supplied coordinate fallback.
- Validate coordinate and camera parameter bounds before contacting Google.

daikaginza:
Fixes #20.

## Changes

`/api/cctv/frame/:id` degrades upstream snapshot → Street View → synthetic card. Only the middle step leaves the machine on a metered account: Street View Static bills the operator's own Google key per image. That step read `lat`/`lon` off the query string, ran for any camera id whether or not the catalog served it, and nothing counted the calls.

Two things followed from that, both reproducible in one request:

- an id the catalog never served still bought a lookup — `GET /api/cctv/frame/no-such-camera?lat=…&lon=…` returned a Street View frame;
- a *registered* id could be aimed anywhere. `/api/cctv/frame/tln-103?lat=48.8584&lon=2.2945` asked Google for the Eiffel Tower while `tln-103` is a Tallinn camera at `59.436563, 24.752702`. That makes a registered camera a general-purpose geocoding lookup on someone else's key.

Neither path had a ceiling.

SECURITY.md already states the invariant this sits under — *"the CCTV frame proxy fetches only server-registered camera/frame URLs — clients cannot pass an upstream URL to fetch"* — and CONTRIBUTING repeats it for source packs. The Street View branch honoured it for the *URL* and not for the *destination*: the URL was fixed, but the client chose which place on Earth it returned a photograph of. This change makes the documented guarantee true for that branch too. I have left SECURITY.md's wording alone rather than edit a file thirteen open PRs are touching; say the word if you would like the bullet extended.

- `resolveStreetViewTarget(source)` in the new `server/providers/cctv/streetview.js` resolves the billable target from the **registered record**. The browser still sends `lat`/`lon` on every frame, but `frameUrlFor` in `src/layers/cctv/source.js` copies them off the camera record it got from `/api/cctv/sources`, so they were always a restatement of what the server already holds — which is what makes them safe to stop honouring. An unregistered id resolves to nothing, and so does a pose that is absent, non-numeric or off-globe. `Number(null)` is `0`, so "absent" is checked before coercion rather than after; otherwise a catalog row with no position buys a frame of the Gulf of Guinea.
- Admission is metered per client with `makeRateLimiter` and `clientKey`, the same primitives `military-installations` and `places/routes` use. The cap is 240/min per IP with a 720 aggregate backstop, sized from the layer's own demand rather than from a round number: the ambient card drain is capped at 16 cameras (`CCTV_AMBIENT_CARD_DRAIN_CAP`) and an active card re-requests every `ACTIVE_FRAME_REFRESH_MS` (10s), so one viewer whose cameras have all lost their upstreams legitimately sits at 96 fallback frames a minute. The 90/min used by the Overpass-backed routes would have cut into normal use here.
- `heading` now wraps into `[0, 360)` beside the existing `fov` and `pitch` clamps, because a bearing is periodic — 400 is 40, not "too far north".

A refused lookup returns the synthetic card the route already serves when no frame is available, not a 429: the response goes into an ``, which has no way to render a status code. The upstream-snapshot and synthetic steps are untouched, and a frame for a registered camera is byte-for-byte what it was.

A refusal is deliberately not logged. The one situation that reaches the cap in volume is the abuse loop, so a line per refusal would turn a flood into log spam on the operator's own terminal; the health map already reports the camera as `degraded`, which is the signal the panel reads. Happy to add a rate-limited warning line if you would rather have it.

## How it was verified

Same probe, same instrument, once per side, driving the real exported `cctvProxy()` handler — the catalog, the fallback chain and the SVG exit are the shipped code. Only `globalThis.fetch` is intercepted, at the network boundary, so the probe never spends a real lookup; `GOOGLE_MAPS_SERVER_API_KEY` is set to a dummy value so the fallback is reached at all.

| | `before` (`0d41b6b`) | `after` |
|---|---|---|
| unknown id + client coordinates | served `streetview`, **1** billable call | served `synthetic`, **0** |
| 300 such requests | **300** billable calls | **0** |
| registered `tln-103`, upstream down, request carrying `lat=48.8584&lon=2.2945` | served `streetview`, Google asked for **`48.8584,2.2945`** | served `streetview`, Google asked for **`59.436563,24.752702`** |

The third row is the regression check that mattered most: the fallback still works for a real camera, and it is the catalog's coordinates that reach Google.

`src/data/cctvStreetViewQuota.test.mjs` covers the decision surface — unregistered id, absent/non-numeric/off-globe pose, the `Number(null) === 0` case, bearing wrap, the per-IP window, the global backstop, and that a capped caller does not consume another caller's quota. Mutation-checked rather than assumed: removing the unregistered-camera guard, removing the coordinate bounds, and lowering the cap to 90 each turn exactly one test red, and restoring each returns 11/11.

One thing worth stating plainly, because it shaped the verification: an early revision used `normalizeHeadingDeg` in `cctv.js` without adding it to the import. Every unit test stayed green — they exercise `streetview.js` directly — and `streetViewFallback`'s `catch` swallowed the `ReferenceError` into a `null`, silently demoting every registered camera to the synthetic card. Only the route-level probe caught it. The committed tests cover the pure decision surface, matching how this repo already tests `media.js`; the integration claim rests on the probe above.

Gates on Node 24.14.0 with `GEV_REQUIRE_ALLOCATION_GATE=1`: `npm run doctor`, `npm run format:check` (918 files), `npm run check:boundaries`, `npm run build`, `npm test` (4170 tests, 0 failures, 1 pre-existing Windows-only skip; both allocation microbenchmarks ran), and `npm run test:track`.

Per CONTRIBUTING rule 3, `CHANGELOG.md` and `docs/CURRENT-STATE.md` both record the behaviour change; the `CURRENT-STATE` paragraph sits in **Places and CCTV request bounds**, beside the existing note that Places rejects out-of-range coordinates before its limiter.

`scripts/qa-cctv-v2.mjs`, the CCTV v2 subsystem harness, run on this branch and on `0d41b6b` for comparison: **48 passed / 0 failed** here against **47 passed / 1 failed** on main. The five `INCONCLUSIVE` results are identical on both sides — they need an on-screen gizmo arrow this GL stack does not produce, headless or headful — and main's one failure is the tiles-load-timing ground-sample check, which is network-dependent; I am not claiming this PR fixed it. Its "no console errors during CCTV v2 exercise" check reports clean, which is the rule-2 requirement.

## What this does not do

`heading`, `fov` and `pitch` are still taken from the query string. That is deliberate, and it is the one part of the issue's first bullet this PR does not close literally. They are a genuine refinement rather than a restatement — `src/layers/cctv/source.js` sends `camera.pitchDeg || -10`, so a catalog row with no pitch gets a sensible default from the client — and pinning them to the catalog would change what real frames look like. They cannot choose a location, cannot change what a call costs, and are each bounded before the request. If you would rather the pose came from the catalog too, say so and I will follow up; I did not want to change rendered output inside a security fix.

#29 and #22, from the same review, are untouched: SSRF hardening for server-registered CCTV URLs and preview-server exposure of key-bearing routes are separate problems in separate code paths, and folding them in here would make one change impossible to review.


daikaginza:

On the root cause, because it is shared with several of the other open reports rather than specific to this route:

These proxies were written with real care about the **upstream** — validate the target, cap the body, cache the answer. What none of them bounded was **the caller**. This route is the sharpest version of that, because the caller was not merely unbounded: it was *steering*. `lat` and `lon` came off the query string and went straight into a billable Google Street View lookup, so anyone who could reach the dev server could aim the operator's key anywhere on Earth and repeat it as fast as they liked.

So the fix is two separate things that happen to live in one route:

- **Aim.** The pose now comes from the registered camera in the catalog, never from the request. `lat`/`lon` are still sent by the browser on every frame; they are simply no longer read. An unregistered id, or a registered camera with no usable pose, cannot spend a lookup at all.
- **Rate.** A per-IP window with a global backstop, so one host cannot drain the key and a refused caller does not eat anyone else's quota.

One design note worth a reviewer's eye: **a refusal here is not a 429.** This route answers an `` element, which has no way to render a status code — a 429 would just be a broken image. The route already has an answer for "no frame available" (the synthetic card), so a refused Street View lookup takes that same exit. It is the one place in this batch where the limiter deliberately does not surface as a 429.

`heading` wraps rather than clamps, because a bearing is periodic: 400° is 40°, not "too far north". `fov` and `pitch` still clamp, because those are not.

On verification, and the honest limit of it: I checked the real door — dev server, actual HTTP — and confirmed the route's contract is unchanged (registered id, unknown id, and hand-written `lat`/`lon`/`heading` all still answer 200 with a renderable frame or the synthetic card). But **the metered branch itself cannot be exercised at the door without a billable Google key, and I am not going to spend someone's quota to produce a screenshot.** That branch is gated by `src/data/cctvStreetViewQuota.test.mjs` instead — 11 cases covering the unregistered id, the client-supplied coordinates, a finite-but-not-a-coordinate value, the missing pose, bearing wrap, the per-IP window, the global backstop, and one caller not consuming another's quota.

Full gates on the rebased tree: `npm test` 4191 tests / 0 fail, 13 allocation microbenchmarks under `GEV_REQUIRE_ALLOCATION_GATE=1`; `npm run test:track` 109 / 0; doctor, format:check, check:boundaries, build all clean.

Issue #648, opened 2026-09-17
feat: get OpenStreetMap data from planet dumps not Overpass


lonvia:
### Problem

This application is massively overusing overpass-api.de to download OpenStreetMap data. By our accounts you have used up around 20k CPU seconds in the last 24h. Given the resources we have have and the numbers of users, your fair share would be at around 300 CPU seconds. In addition, [changing the user agent after being UA blocked](https://github.com/bilawalsidhu/gods-eye-view/pull/420) has now resulted in a [very annoyed sysadmin](https://www.redbubble.com/i/t-shirt/Don-t-Mess-With-System-Administrators-We-Have-Admin-Rights-by-jaygo/38981176/z5wf).

So this is to let you know that you have been blocked again.

### Proposal

Please find a solution where you process OpenStreetMap data yourself and deliver them to your users in form of vector tiles or similar. Or you can run your own Overpass server. There are also [commercial offers](https://wiki.openstreetmap.org/wiki/Overpass_API#Public_Overpass_API_instances) to help you do that.

Simply dropping overpass-api.de and switching to another "free" Overpass server is not going to be a solution either. These other services have even less computing resources available than the main instance. So this inevitably leads to even more annoyed sysadmins.

### Data source, license, and attribution

OpenStreetMap data should be usually consumed from [the planet dump](https://planet.osm.org/pbf/) and, if frequently updated data is required, from [diffs](https://planet.osm.org/replication/). License is ODbL, attribution "OpenStreetMap".

### Will you build it?

No, this is a request for someone else

manishraj27:
I would like to work on this issue. Please Assign me

bilawalsidhu:
Thanks for telling us directly, and sorry for the load. You're right: every install shares one User-Agent, so usage adds up, and changing it after the block was the wrong response. We'll keep a stable identity going forward.

Street Traffic road queries caused most of the load. We're moving them to vector tiles, or to the user's own traffic provider, and ALPR cameras to a pre-built OSM extract. Public Overpass instances and fallback mirrors come out of the defaults; features without a replacement will say so, and anyone wanting live Overpass can configure their own or a paid endpoint.

We'll post here when the release is out. At that point we'd like your view on one thing: occasional area-outline lookups for annotations, heavily cached. Is there a level you'd consider acceptable, or would you prefer we drop public Overpass entirely?

We really appreciate the work you do keeping Overpass running, and we'd love for this open-source project to keep benefiting from it the right way.

[...]

PR #667, opened 2026-09-19
feat(overpass): support self-hosted endpoints via OVERPASS_EXTRA_UPSTREAMS (#648)

theGaryLarson:
> **Draft:** the code, tests and docs are complete, but I have not yet confirmed dots on screen end to end against a freshly imported local instance. Marking ready once that passes.

## Why

Closes the operator half of #648. The overpass-api.de maintainers report this app consuming ~20,000 CPU-seconds in 24h against a fair share of ~300, and asked heavy consumers to run their own instance rather than rotate through the remaining free ones. Today there is no way to do that short of editing `server/providers/overpass/constants.js`, so a user's practical options are "keep hammering a shared mirror" or "the Overpass-backed layers are broken". Every one of them — Street Traffic, Mapped Installations, the OSM webcam pack — needs Overpass geometry before it can render anything, so this is not a niche path.

This does not itself solve #648 (that wants planet dumps / vector tiles). It gives the people who are ready to self-host today a supported way to, and it makes the recommendation in that thread actionable.

## What

**`OVERPASS_EXTRA_UPSTREAMS`** — one or more endpoints, whitespace- or comma-separated, tried **before** the built-in list. Built-ins remain as fallback.

Extras go first deliberately: an endpoint someone configured is a better first try than one that just timed out for them, and each dead built-in costs a full `OVERPASS_TIMEOUT_MS` before the chain advances. Resolution is lazy and memoised on the raw env string — module evaluation happens before the dev server copies `.env` into `process.env`, so reading it at module scope always sees `undefined`.

Structural URL validation only. A localhost or RFC1918 address is *allowed* here, unlike `osmWebcamStillUrl` whose input is a world-editable OSM tag; this value comes from the operator's own environment, where a private instance is the entire point.

**The regional-extract hazard.** This is the part worth reviewing closely. An instance built from a state or country extract answers HTTP 200 quickly for queries anywhere on Earth, returning an empty element list outside its own region. Left alone the proxy caches that as a valid answer for 7 days, silently pinning every Overpass-backed layer to "no roads here" over regions that have plenty — it fails quietly, which is the worst shape. So an empty answer from a *configured* endpoint is treated as that endpoint declining: rotate to the next, and if nothing answers, surface a visible failure. Built-in planet mirrors keep the original semantics, where empty can legitimately be the truth.

The emptiness check is substring-first so multi-MB payloads are never parsed just to learn they are non-empty, and an unparseable body is explicitly *not* called empty — an unknown shape is the caller's problem, not grounds for rotating off a healthy endpoint.

**Test fix.** `coalesced outage callers both receive last-good data` asserted `fetches === 4`, a hardcoded mirror count that any operator with their own endpoint would fail. It now counts the resolved chain; the invariant under test is one shared sequence per outage, not how many mirrors exist.

## Ships with

- **`docker-compose.overpass.yml`** — parameterised by `OVERPASS_PLANET_URL`, defaulting to a single US state: small enough to prove the setup works before committing to a continent. Note `OVERPASS_PLANET_PREPROCESS`: Geofabrik no longer publishes `.osm.bz2` for US extracts and this image expects bz2, so the `.pbf` is converted in place with `osmium cat`. That line is load-bearing, not boilerplate.
- **`docs/SELF-HOSTED-OVERPASS.md`** — prerequisites, verification, region-switching, troubleshooting, and measured disk/time per region:

| Region | Download | Disk after import | Import time |
|---|---|---|---|
| Washington state | 0.3 GB | ~2 GB | ~15 min |
| California | 1.2 GB | ~7 GB | ~45 min |
| Great Britain | 2.0 GB | ~12 GB | ~1 hr |
| United States | 11.3 GB | ~60 GB | ~6 hr |
| North America | 18.1 GB | ~85 GB | ~8–12 hr |
| Europe | 32.6 GB | ~150 GB | ~16–24 hr |
| Planet | 88.3 GB | ~400 GB | multiple days |

Downloads are measured from Geofabrik/planet.osm.org; disk and time are estimates that scale with storage speed, and assume `OVERPASS_META: 'no'` as shipped. Because traffic only activates below `ACTIVATION_ALTITUDE` (8 km), coverage only needs to include the places a user actually descends into — starting regional and re-importing larger later costs one extra download and nothing else.

## Credit

The `OVERPASS_EXTRA_UPSTREAMS` idea and its parser are from @Schnix84's #104, which was closed pending a maintainer batch. That PR predates the `vite.config.js` → `server/providers/overpass/` split, so this is a port to the current layout rather than a rebase, plus the regional-extract handling and the compose/docs.

## Testing

- 7 new tests in `src/data/overpassUpstreams.test.mjs`: parsing, ordering, dedup, private-address allowance, empty-rotates-for-extras, and empty-is-still-data for built-ins
- `src/overpassProxy.test.mjs` 11/11
- `npm run format:check` and `npm run check:boundaries` clean
- No behaviour change when `OVERPASS_EXTRA_UPSTREAMS` is unset: the resolved chain is exactly the current built-in list


theGaryLarson:
## Verified end to end

Ran a real Washington-state import and drove the app against it. Results:

| Check | Result |
|---|---|
| Direct query, downtown Seattle bbox | 1633 elements, **0.56 s** |
| Same query through `/api/overpass` | 1633 elements, **3.1 s** cold, `X-Overpass-Cache: HIT` warm |
| Serving endpoint | `X-Overpass-Upstream: http://localhost:12345/api/interpreter` |
| Street Traffic layer | `mode: "live"`, `LIVE · TomTom flow · 82% cov`, **2617 dots**, `error: null` |
| Flow buckets | free 2140 · slow 12 · jam 0 · sim 465 |

For contrast, the same query against the public mirrors from this machine takes **45 s to fall through all four and return an Apache 406 error page**.

Named roads came back correct for the bbox — Montlake Boulevard NE, Ship Canal Bridge, Mount Baker Ridge Tunnel — so this is real geometry, not a shape that merely parses.

### Two first-run defects this surfaced

Both are in the compose file and both only appear on a genuine first import, which is why they are worth calling out rather than folding in silently:

1. **`/db` is the `overpass` user's home directory**, so a fresh named volume inherits its `0700` mode. The import runs as `overpass` and reports success, but the CGI process answering queries runs as a different user and cannot traverse `/db` to reach the dispatcher socket. Every query fails with `runtime error: open64: 13 Permission denied /db/db//osm3s_osm_base` while the logs say initialization completed — a healthy-looking container serving nothing.
2. **Overriding `entrypoint` clears the image's CMD**, which the image's own entrypoint needs as its argument. Without restating it the container exits 0 and restarts forever with empty logs.

Fixed in 676daf0 and documented in the troubleshooting section.

### Note on `scripts/qa-traffic-baseline.mjs`

It cannot validate a regional instance: it primes a hardcoded **Austin** viewport, which is outside a Washington extract, so the empty-rotates-as-refusal guard correctly declines and the harness times out waiting for dots. That is the guard working as designed rather than a regression — but it does mean the harness assumes planet-wide Overpass coverage, which may be worth parameterising separately.

Marking ready for review.

PR #669, opened 2026-09-19
docs(attribution): credit adsbdb and carry its route-data terms

### LINKED ISSUE #256: data-attribution: api.adsbdb.com is missing from DATA_SOURCES.md and DATA_CREDITS
**What**

`api.adsbdb.com` is a live, user-visible data source in this app, but it is registered in neither attribution surface. `DATA_SOURCES.md` and `src/data/dataCredits.js` each contain **zero** occurrences of `adsbdb`, which contradicts the rule both files state for themselves:

- `DATA_SOURCES.md` § In-app attribution: *"When you add a new data source, add its license and attribution to this file **and** append an entry to `DATA_CREDITS` in `src/data/dataCredits.js` so it surfaces in the app."* The same paragraph enumerates the credited layer sources — adsb.lol, TeleGeography, OSM datacenters/dams/roads, NASA FIRMS, CelesTrak, USGS, City of Austin, GBFS, Radio Browser, OpenSky, AISStream. adsbdb is not among them.
- `CONTRIBUTING.md` § Pull requests item 4: *"If you add or change a data source, update `DATA_SOURCES.md` with its license and attribution."*

adsbdb and adsb.lol are different operators; the existing `adsb.lol` credit (ODbL 1.0) does not cover it.

**Where it is used** (verified on `main` @ `759652207fd1279ece97f0f19af566feb9a82146`)

- `vite.config.js:2390` — `adsbdbProxy()`, registered as the `/api/adsbdb` middleware at `:2475` and in the plugin list at `:7749`. Upstream calls to `https://api.adsbdb.com/v0/callsign/…` and `/v0/aircraft/…` at `:2447`–`:2448`.
- `src/data/flights.js:823` / `:847` — consume `/api/adsbdb/type/…` and `/api/adsbdb/route/…`, writing `registration`, `typeCode`, `typeName`, and `classifyAircraft`-derived `klass` onto the flight record, plus `airline`/`route` for the tracked contact.
- That data is not incidental: `klass` drives billboard presentation and the GLB model resync (`_applyFleetBillboardPresentation` / `_syncModelToClass`), and the route feeds the cockpit `FROM → TO` strip and destination chevron (`docs/CURRENT-STATE.md:1208`, `:1569`, `:2391`). So the source is already documented as an internal runtime data flow — it is only missing from the two files that exist to carry license/attribution.

**Why there is no patch attached**

I could not verify adsbdb's license or terms, and inventing the required attribution string is precisely the claim that should not be guessed. The minimal fix is a row plus a "Notes on the live sources" bullet in `DATA_SOURCES.md`, a matching `DATA_CREDITS` entry, and — if the terms constrain it — a decision recorded on the on-disk cache. Whoever knows the intended terms can land it in one commit.

Worth noting for that decision: the proxy writes responses to `.gev-cache/adsbdb.json` (`vite.config.js:2392`), and `flights.js:770`–`:774` deliberately bounds fan-out because "adsbdb is a free community API". Every other source with a caching or politeness posture has it written down in the "Notes on the live sources" list (e.g. TomTom's ≤120 s TTL, OpenSky's non-commercial caveat, LL2's courtesy credit). adsbdb's is recorded nowhere.

**Verified before filing**

- `DATA_SOURCES.md` (blob `faa351dc…`) and `src/data/dataCredits.js` (blob `b35da9e0…`) at `main`: 0 matches for `adsbdb` (case-insensitive) — `adsb.lol` appears 8× and 4× respectively.
- No issue or PR in any state names adsbdb in its title or body (GitHub search, `repo:bilawalsidhu/gods-eye-view adsbdb` → 0 results), and no open PR adds an adsbdb credit. The one PR diff that mentions the host (#51's Dockerfile comment listing the `configureServer`-only proxies) is prose and touches neither attribution file.
- Docs/registry inconsistency only — no runtime impact and nothing security-sensitive.

daikaginza:
Fixes #256.

## Changes

`api.adsbdb.com` is a live, user-visible source — it supplies the aircraft type, model name and registration on enriched flights, and the airline and origin/destination pair behind the tracked contact's `FROM → TO` strip — and it was registered in neither attribution surface. On `0d41b6b`, `DATA_SOURCES.md` and `src/data/dataCredits.js` each contain **zero** occurrences of `adsbdb`, which is what both files and CONTRIBUTING's fourth pull-request item say must not happen. `adsb.lol` is a different operator; its ODbL credit does not cover this data.

@jsawyerdev left the patch off #256 on purpose — *"I could not verify adsbdb's license or terms, and inventing the required attribution string is precisely the claim that should not be guessed."* That was the right call, so this PR takes the terms from the source instead of inventing them. From the [adsbdb README](https://github.com/mrjackwills/adsbdb):

- aircraft data comes from **PlaneBase**;
- the ICAO-to-N-number conversion is **Guillaume Michel**'s;
- the flight route data *"is the work of David Taylor, Edinburgh and Jim Mason, Glasgow, and may not be copied, published, or incorporated into other databases without the explicit permission of David J Taylor, Edinburgh."*

No licence and no rate limit are published for the service itself, so neither is claimed here.

- **`DATA_SOURCES.md`** — one row in the live-sources table, and a bullet in *Notes on the live sources*. The note records the politeness posture the issue observed was written down nowhere: at most 4 requests in flight, dispatches dripped to roughly 5/s, each key asked at most once per session, route lookups reserved for the tracked aircraft, and a 24-hour per-key disk cache including negative answers. It also records why that cache is safe to keep — `.gev-cache/` is gitignored, so the route data stays on the operator's machine and is never bundled or redistributed, which is what its terms require.
- **`src/data/dataCredits.js`** — the matching `DATA_CREDITS` entry, so the credit reaches the in-app **Data attribution** popover, and the `adsbdb` name added to the enumeration in *In-app attribution* that had listed every other credited source.
- **`src/data/dataCredits.test.mjs`** — new. Pins the adsbdb credit and the route-data restriction, plus a structural check that every entry has a unique key and something to render.
- **`CHANGELOG.md`** — one entry, following #511, the closest precedent here: a documentation-only attribution change that still earned a changelog line.

The test is the part worth arguing for. This attribution is a requirement, not a preference: if a later refactor drops the two names or the restriction, that should fail a gate rather than go unnoticed until someone reads the popover.

## How it was verified

The claim is that the source *surfaces in the app*, so that is what was measured — the real page, not the module.

Loaded the app in Chromium, expanded the **Data attribution** popover and read the rendered line back out of the DOM. It is one of 42 credit lines, and reads:

> Aircraft type, registration & flight routes: adsbdb · aircraft data from PlaneBase · route data is the work of David Taylor, Edinburgh, and Jim Mason, Glasgow, and may not be copied, published, or incorporated into other databases without the explicit permission of David J Taylor, Edinburgh

The `before` side needs no browser: the registry held no `adsbdb` entry at all, so there was nothing for the popover to render.

The new test is mutation-checked rather than assumed green — stripping the restriction from the credit, and renaming the entry's key, each turn it red; restoring returns 2/2.

Gates on Node 24.14.0 with `GEV_REQUIRE_ALLOCATION_GATE=1`: `npm run doctor`, `npm run format:check` (917 files), `npm run check:boundaries`, `npm run build`, `npm test` (4161 tests, 0 failures, 1 pre-existing Windows-only skip; both allocation microbenchmarks ran) and `npm run test:track`.

## Notes for review

Five judgement calls worth your eye:

- The note says the disk cache holds each key for **24 hours**, which is `TTL_MS = 24 * 3600_000` in `server/providers/aircraft/enrichment.js:11`. `src/layers/flights/policy.js:216` describes that same cache as keeping entries "on disk forever" — the comment predates the TTL. I took the number from the constant and left the comment alone rather than widen this PR into the flights layer, but it is worth a one-line fix by someone.
- This does not collide with #308, which bounds the same cache. That PR adds an entry ceiling and a server-side limiter *on top of* the 24-hour TTL; it does not change the TTL, the 4-concurrent fan-out or the 200 ms drip, so every number written here stays true whichever merges first. Nothing in this PR touches the proxy.
- `docs/CURRENT-STATE.md` is deliberately untouched. It already describes the adsbdb middleware, its caching, and the cockpit `FROM → TO` strip in five places; no runtime capability changes here, only the attribution surface. #511 — the documentation-only attribution change this most resembles — touched `CHANGELOG.md` and not `CURRENT-STATE.md`, and that is the shape followed.
- The table row is not padded to the column widths of its neighbours. The table already mixes padded and unpadded rows, and re-aligning it would bury a one-line addition in a whole-table diff.
- `scripts/qa-l9-matrix.mjs`'s C12 credit expectations are untouched. adsbdb enriches the flights layer rather than being a layer of its own, so `flights: /OpenSky/i` still describes that layer's required credit correctly; the adsbdb credit is registered statically and is present regardless of which layers are on.


samehkhamis:
Thank you for the contribution and for correcting the missing adsbdb attribution!

daikaginza:
Thank you — and thanks especially for not just merging it but fixing it.

Both corrections landed. The Guillaume Michel one stung in a useful way: I'd documented all three sources in `DATA_SOURCES.md` and then rendered only two. That's exactly the gap where writing something down feels like having done it, so I've started checking what the credit actually renders rather than what the doc says.

The other one I've already borrowed from you. Replacing my "what the route data's terms require" with the published fact plus "Deployers are responsible for obtaining any permission their use requires" is a better shape for anything licence-adjacent — state what's published, don't assert compliance on the project's behalf. I used that framing in my next PR.

One loose end from this one, if you want it: `src/layers/flights/policy.js:216` says the proxy "caches per-key on disk forever ... so repeat sessions never re-hit adsbdb", but `server/providers/aircraft/enrichment.js:11` sets `TTL_MS = 24 * 3600_000` and line 39 enforces it. Both halves of that sentence are off — it's a day, not forever, and sessions a day apart do re-hit. Happy to send a one-line comment fix, or to leave it alone if you'd rather keep the queue short.

PR #677, opened 2026-09-20
fix(tomtom): size the daily tile budget to the real free allowance

GGPOShadows:
## Problem

`TOMTOM_DAILY_TILE_BUDGET` defaults to **40,000 per day**, but TomTom's free allowance is granted **per month, not per day**: 200,000 Traffic Flow & Incidents tile requests per month ([pricing](https://docs.tomtom.com/pricing/)).

At the current default a user exhausts their entire monthly allowance in **five days**, and the traffic layer then falls back to stale/no tiles for the remaining ~25 days of the billing period.

The docs also contradict each other on this figure:

- `server/providers/traffic.js` describes the free tier as `~50k/day`
- `.env.example` cites the correct `200K tile requests/month` two lines below quoting the `40000` default
- `DATA_SOURCES.md` repeats `default 40,000`

`~50k/day` would be ~1.5M/month, about 7.5x the actual allowance.

## Changes

- `DEFAULT_DAILY_BUDGET` 40,000 -> 6,000. A 31-day month lands at 186,000, inside the 200K allowance with headroom.
- Corrects the stale free-tier figures in `server/providers/traffic.js`, `.env.example` and `DATA_SOURCES.md`.
- Updates `CHANGELOG.md` and `docs/CURRENT-STATE.md` per CONTRIBUTING.

This stays an **application-side safety ceiling, not a billing cap**, and the surrounding text still says so. TomTom's free tier needs no card, so an overrun cuts the layer off rather than generating a charge — but a layer that dies three weeks into every month is worth avoiding by default. Anyone who wants the old behavior can still set `TOMTOM_DAILY_TILE_BUDGET=40000`.

## Verification

Run on Windows 11, Node 24.19.0, with a real TomTom key configured.

- `npm test` — 4,149 passed, 0 failed
- `npm run build` — clean, only the pre-existing chunk-size warnings
- `npm run test:track` — **109 passed, 0 failed**, no console errors, no HTTP 5xx
- Live check: `/api/tomtom/status` reports `{"hasKey":true,"dailyCount":0,"budget":6000,...}` with the default, and still honors an explicit `TOMTOM_DAILY_TILE_BUDGET` override.

## Note

I checked open PRs before filing and found none touching the tile budget or the free-tier figures. Happy to adjust the number if you'd rather target a different safety margin.


samehkhamis:
Thank you for the contribution, @GGPOShadows!

PR #678, opened 2026-09-20
fix(security): ignore the whole dotenv ladder, not just .env

GGPOShadows:
## Problem

`.gitignore` lists only the bare filename `.env`, but the app reads **four** credential files, not one:

- `scripts/setup-doctor.mjs:92` walks ``['.env', '.env.local', `.env.${mode}`, `.env.${mode}.local`]``
- `server/standalone/vite.config.js:11` calls Vite's `loadEnv(mode, root, '')`, which reads the same ladder

gitignore matches literally, so three of those four are committable today. Verified on a clean checkout of `main`:

```
.env                    ignored
.env.local              COMMITTABLE
.env.production         COMMITTABLE
.env.production.local   COMMITTABLE
```

A key placed in `.env.local` is **fully live** for the app. `npm run doctor` reports it as configured, the provider proxies use it, and the browser-exposed defines pick it up. It is also staged by `git add -A`:

```
$ printf 'OPENAI_API_KEY=sk-proj-…\n' > .env.local
$ git add -A --dry-run
add '.env.local'
```

Two things make this sharper than a generic hygiene nit:

1. The in-app **Provider Settings** panel deliberately trains people to keep real keys inside the checkout rather than in a shell profile, so the odds of a key landing in a dotenv file here are high by design.
2. The repository has ~7.7k forks, all of which inherit the gap.

## Changes

Replaces the single `.env` entry with:

```gitignore
.env
.env.*
!.env.example
```

The negation keeps the tracked `.env.example` template exactly as-is — confirmed still tracked and still not ignored after the change.

Deliberately scoped to the dotenv ladder only. Other credential-bearing artifacts (a repo-local `OPENSKY_CREDENTIALS_FILE`, `*.pem` / `*.key`) are a separate judgement call and are left for a follow-up rather than widened into this PR.

## Verification

Windows 11, Node 24.19.0.

- Repro above no longer stages: `.env.local` is ignored after the change
- All of `.env.local`, `.env.production`, `.env.production.local`, `.env.development.local`, `.env.test` now ignored
- `.env.example` still tracked and not ignored (`git ls-files --error-unmatch .env.example` passes)
- `npm test` — 4,145 passed, 0 failed
- `npm run build` — clean
- `npm run test:track` — **109 passed, 0 failed**

## Note

I checked open and closed PRs and issues before filing and found nothing covering this. No behavior changes, so no `CHANGELOG.md` / `docs/CURRENT-STATE.md` entry — happy to add one if you'd prefer it recorded.


samehkhamis:
Thanks @GGPOShadows — this closes a real credential-commit gap with a clean, focused fix. Merged!

Issue #8, opened 2026-08-25
Render Performance Suggestions


WillBeebe:
Super cool project, thank you!

Explored the repo — it's a CesiumJS globe app that's already had several documented perf passes (render governor, throttled preRender ticks, billboard collections). Remaining high-value wins, ranked:

## Top 5 rendering-performance improvements

1. **Remove/gate `preserveDrawingBuffer: true`** — src/main.js:122
   Forces the browser to preserve the full-screen WebGL drawing buffer every frame, defeating the zero-copy buffer-swap fast path. It's only needed for screenshots/recording — gate it to those modes (or use `scene.canvas.toDataURL()` immediately after render).

2. **Drop `msaaSamples: 4` to 2 or make it adaptive** — src/main.js:119
   4× MSAA over Google Photorealistic 3D Tiles is one of the biggest fixed GPU costs per frame; brutal on integrated GPUs. Expose as a quality setting (the app already has a perf-investigation precedent for this pattern).

3. **Cut the 58 `backdrop-filter: blur()` rules** — style.css (58 occurrences)
   Every blurred HUD panel forces the compositor to read and blur the WebGL canvas beneath it *every frame*, and panels stack. Replacing with pre-baked translucent gradients/opaque backgrounds on the heaviest surfaces (command dock, trays, chips) is likely the single largest paint-cost win.

4. **Replace the compass-tape `innerHTML` rebuild** — src/ui.js:1450-1457
   On heading change it re-parses/rebuilds 7 `<span>`s via `innerHTML` string concat → parse + layout + GC churn during exactly the most interaction-heavy moments (camera motion). Build the strip once and slide it with a `transform: translateX` update (compositor-only).

5. **Split the monolithic render-blocking CSS** — index.html:8 (`style.css`: 247 KB / 9,163 lines) + 3 synchronous Google Fonts stylesheets (index.html:11-13)
   All of it blocks first paint even though much of the UI (cockpit mode, trays, modals, per-style themes) is hidden initially. Inline critical above-the-fold CSS, lazy-load the rest, and self-host/preload the two font families.

Honorable mentions: cap the world-overlay canvas backing store DPR (~19 MB at DPR 2, src/overlays/worldOverlay.js:1102) and coalesce the scattered HUD poll timers (250 ms/500 ms/60 ms in src/hud.js:216-238, src/ui.js:9188-9216) into one ticker.

demolerbi-cmyk:
> Super cool project, thank you!
[...]

PR #680, opened 2026-09-20
feat(perf): opt-in render-quality presets for weak GPUs

GGPOShadows:
Follow-up to [the profiling I posted on #8](https://github.com/bilawalsidhu/gods-eye-view/issues/8#issuecomment-5748116872). You asked there for current-main profiling and a measured frame-time comparison before taking perf changes; this implements only the part that measured.

## Problem

The viewer is constructed at `msaaSamples: 4` and never sets `resolutionScale`, so it renders at full device pixel ratio with 4x MSAA. On a discrete GPU that's the right default and this PR leaves it alone. On integrated graphics it's the difference between a usable globe and a slideshow.

## Change

`src/app/renderQuality.js` — a pure preset table plus a resolver over `location.search`, following the existing query-flag convention (`?detectDebug=1`, `?trafficDebug=1`). `?quality=balanced` and `?quality=performance` set `scene.msaaSamples` and `viewer.resolutionScale` together. Both are runtime-settable in Cesium, so the preset is applied just after construction and the WebGL context is never rebuilt.

**The default is unchanged.** `high` reproduces the construction values exactly, so nothing moves unless a user opts in. A test asserts that equivalence so the guarantee can't silently rot.

| `?quality=` | msaa | scale | measured |
|---|---|---|---|
| *(none)* / `high` | 4 | 1.00 | 17.9 fps |
| `balanced` | 2 | 0.85 | 22.6 fps |
| `performance` | 1 | 0.60 | 35.1 fps |

## Measurement

Intel UHD 770 (ANGLE/D3D11, i7-12700T), 1264x705 canvas, keyless Esri basemap, no layers enabled, parked camera, warm tile cache.

Frames are counted from `scene.postRender` and every arm asserts the canvas dimensions are unchanged. Worth stating why: an earlier unverified pass of mine reported a clean 60 fps for one arm — rendering had silently stopped and rAF was just free-running. Any perf number not tied to a verified render count and canvas size shouldn't be trusted, including mine.

End-to-end verification drove the real app at each `?quality=` value and asserted the viewer actually adopted the preset:

```
PASS  no param (default)       msaa 4 scale 1     canvas 1264x705  16.5 fps
PASS  ?quality=high            msaa 4 scale 1     canvas 1264x705  15.6 fps
PASS  ?quality=balanced        msaa 2 scale 0.85  canvas 1074x599  22.6 fps
PASS  ?quality=performance     msaa 1 scale 0.6   canvas  758x423  35.1 fps
PASS  ?quality=nonsense        msaa 4 scale 1     canvas 1264x705  15.7 fps
```

## What is deliberately NOT here

Two items ranked above MSAA in #8 measured as noise on this hardware, so I left them out rather than bundle unmeasured changes into a perf PR:

- **`preserveDrawingBuffer`** — flipped to `false` and re-ran the full matrix: 17.9 → 18.2 fps with overlapping samples, across two independent runs. The reasoning is sound in general; on this driver path it isn't costing anything measurable.
- **`backdrop-filter`** — +6% in one run, −1% in the next. Also tested hiding overlay containers one at a time; no single element moved the needle.

Also worth recording: frame time fits roughly **22 ms fixed + 40 ms per megapixel** across scales 1.0/0.75/0.5/0.35/0.25. That fixed term is outside JS (main thread idle ~72%, Cesium render phase ~11 ms) and bounds this machine near ~44 fps regardless of preset. The presets buy back the resolution-dependent half; I could not reduce the rest.

## Robustness

Resolution is total and fails soft. Unknown, empty or malformed values fall back to `high`; prototype keys (`__proto__`, `constructor`) can't smuggle a non-preset through the lookup (`Object.hasOwn`, not `in`); a viewer without a scene is left alone rather than taking down startup over a display preference. 12 tests cover these.

## Verification

Windows 11, Node 24.19.0.

- `npm test` — 4,157 passed, 0 failed (12 new)
- `npm run build` — clean
- `npm run test:track` — **109 passed, 0 failed**
- `npm run format:check`, `npm run check:boundaries` — clean
- `CHANGELOG.md` and `docs/CURRENT-STATE.md` updated per CONTRIBUTING

New module registered in both `package-boundaries.json` groups that own `src/app/viewer.js`, and the new test in `format-scope.json` (test files are excluded from runtime auto-discovery).

## Notes

n=1 machine, so the fixed-cost floor in particular may not generalise — the preset *values* are a judgement call and I'd happily take different numbers. A DISPLAY-rail control is the obvious follow-up; I kept this to the mechanism plus the evidence rather than also touching the shell, share-link state and the structural invariant tests in one go. Happy to do that next if you want it.


GGPOShadows:
Rebased onto current `main` after #677/#678 landed (thanks for the quick merges). Two conflicts, both header-only: the new `renderQuality` import now sits above the trackpad pinch-zoom block from #284 in `src/app/viewer.js`, and the changelog entry is ordered above the newer merges. Re-ran the gates on the rebased tree: `npm test` 4,183 pass / 0 fail, `npm run test:track` 109 / 0 / 0, build, format and boundaries clean. Still one commit.

bilawalsidhu:
We'd like this in, thanks for measuring it properly. Asks:
- Expose it in the Display panel, not only as a URL flag: Performance / Balanced / High, plus an Ultra tier if there's a meaningful step up on discrete GPUs (for example higher MSAA or supersampling).
- Make Balanced the default, and persist the choice like the other Display settings. Keep `?quality=` as an override.
- Rebase onto current main.

PR #683, opened 2026-09-21
fix(server): throttle the cost-bearing proxies by default

daikaginza:
Refs #16.

## Changes

The per-IP caps on the endpoints that spend a server-held key were opt-in, so an unconfigured server ran them unlimited. This turns them on, using the caps the Pinokio build already ships.

What made this worth doing is not the issue text but what the rest of the server already does. An always-on per-IP cap is the established pattern here — six modules apply one with no env var involved, and every one of them fronts a free upstream:

| Route | Upstream cost | Per-IP cap on `de2a419` |
| --- | --- | --- |
| `/api/overpass` | free | 90/min |
| `/api/military-installations` | free | 90/min |
| `/api/route` (OSRM) | free | 60/min |
| `/api/weather-effects` | free | 45/min |
| `/api/geocode` (Nominatim) | free | 30/min |
| `/api/regional-brief` | free | 30/min |
| `/api/openai/hud-summary` | **billed — OpenAI** | opt-in, off by default |
| `/api/realtime/token` | **billed — OpenAI** | opt-in, off by default |
| `/api/google/nearby-places` | **billed — Places** | opt-in, off by default |
| `/api/google/text-search` | **billed — Places** | opt-in, off by default |

To be precise about what this table does and does not say: plenty of other free routes carry no per-IP cap either, and several bound themselves by other means instead — `/api/tomtom` keeps a persistent daily tile budget, `/api/adsbdb` a disk cache and a bounded fan-out, `/api/opensky` request coalescing. The claim is narrower than "everything else is capped": it is that an unconditional per-IP cap is an existing, repeatedly-used pattern in this server, and the four routes that bill the operator were the ones not following it. This makes them follow it rather than invent a policy.

- **The caps are yours, not mine — and they are the ones already shipping.** `pinokio/_ENVIRONMENT` sets `GEV_RATELIMIT_OPENAI_PER_MIN=30` and `GEV_RATELIMIT_GOOGLE_PER_MIN=120`, so every Pinokio install already runs with exactly these numbers. Adopting them as the code defaults leaves the packaged app's behaviour untouched and changes only unconfigured servers — from unlimited to what the product already runs with.
- **A drift this PR also closes.** `.env.example` had come to suggest `60` for Google while the shipped build used `120`. The file now states the defaults and keeps `60` as the tightening it recommends when `HOST` is not localhost, and a test reads `pinokio/_ENVIRONMENT` and pins the code defaults to it, so the two cannot separate again unnoticed.
- **Measured against real demand, not guessed.** An open tab asks for a HUD summary every 15s (`HUD_SUMMARY_INTERVAL_MS`) — 4/min — and a voice session mints one token. 30/min clears that more than seven times over. Places calls are user-driven rather than polled — a search, a nearby lookup, an installation probe — so 120/min is far above anything the app generates.
- **A typo no longer disarms the guard.** Unset or blank takes the default; a positive integer overrides it; exactly `0` disables the limiter. A value that cannot be read as a number — `3O`, `60/min`, a negative — now takes the default instead of meaning unlimited. Opting out stays possible, but only by saying `0`, which nobody types by accident.
- **Keyless paths are untouched.** Keyless HUD summary and keyless Places still resolve before the limiter, so a server with no key spends no budget on them.
- **Two renames.** `makeOptInRateLimiter` → `makeCostRateLimiter`, `enforceOptInRateLimit` → `enforceRateLimit`. After this change the old names state the opposite of what the functions do, and a name that lies is the same defect as a stale comment.
- Docs follow the behaviour: `.env.example`, `SECURITY.md` ("opt-in" → on by default, with both values), `README.md`, `docs/CURRENT-STATE.md` and `CHANGELOG.md`.

## How it was verified

**Before and after, on the real exported handler.** One LAN peer (`198.51.100.7`), 40 requests to `/api/realtime/token`, upstream `fetch` stubbed so the run costs nothing — everything else is the shipped code path:

| | `upstream/main` | this branch |
| --- | --- | --- |
| tokens minted | **40** | 30 |
| requests refused `429` | 0 | **10** |
| calls that actually reached OpenAI | **41** | 30 |
| first refusal at | never | request **31** |
| `Retry-After` | — | `5` |

The pristine side was measured in a detached `git worktree` at `de2a419`, not by reverting in place.

**Tests.** Eight new cases in `src/costEndpointThrottle.test.mjs`, including two that drive the real route handlers — the realtime token route to its 31st request, and the Places route to its 121st — and assert that the refusal is a `429` with `Retry-After: 5`, that it never reaches the provider, that it does not leak the server-held key, and that the Places refusal keeps the `places: []` shape every other error on that route carries. They are mutation-checked rather than assumed green: restoring "unset means unlimited" turns three red, letting an unreadable value disable the guard turns one red, and reverting returns 8/8. The format gate was itself checked by injecting bad formatting and confirming it failed on the new file.

**Gates** on Node 24.14.0 with `GEV_REQUIRE_ALLOCATION_GATE=1`: `npm run doctor`, `npm run format:check` (919 files), `npm run check:boundaries`, `npm run build`, `npm test` (4183 tests, 0 failures, 1 pre-existing Windows-only skip, both allocation microbenchmarks ran) and `npm run test:track` (109 passed, 0 failed).

The concern worth checking before merging a default-on limiter is whether it throttles the project's own harnesses, since `clientKey` is the socket peer and everything local shares one bucket. It does not: `track-regression.mjs` stubs `hud-summary` and `nearby-places` through `page.route`, and `qa-l9-matrix.mjs` makes three real token requests against a cap of thirty.

Two other things I checked rather than assumed. The cap reaches the preview server as well as the dev server — `openAiRealtimeProxy` registers through one shared `install()` from both `configureServer` and `configurePreviewServer`, so there is no unthrottled second registration. And `createRealtimeTokenHandler` has exactly one caller, so no path mints a secret around the limiter.

## Notes for review

Four things I want your eye on.

- **`Refs`, not `Fixes`.** #16 consolidates #17–#19 and its remaining scope — an auth boundary, an explicit local/LAN/remote access policy — is a product decision, not something a contributor should settle. This is only the "make rate limiting default-on" item. The cross-site half is already handled by #242.
- **Three of #16's references have gone stale** since 25 August, which may matter for whoever closes it. The endpoint is no longer in `vite.config.js:5045-5213`; it lives in `server/providers/openai/`. `vite.config.js:427-479` no longer holds the limiter. And the consolidated scope names "Google geocoding" — `/api/geocode` is keyless Nominatim now, already capped at 30/min, so that item appears to be resolved by other means rather than outstanding.
- **`globalMax` stays at 20× the per-IP cap** — 600/min across the OpenAI routes, 2400 across the Places routes — which is looser than the ~3× the always-on limiters use (30→90, 45→120, 60→200, 90→300). I left it alone deliberately: it is the multiplier the helper already applied, so anyone who had set these env vars keeps exactly the backstop they had. Tightening it to the house ratio is a one-line change if you want the consistency, but it would alter behaviour for existing opt-in users, which felt like your call rather than mine.
- **One cost endpoint is outside this PR.** `/api/cctv` resolves `googleServerApiKey()` for the Street View fallback and has no throttle on `main` either; that one is #666, to keep these reviewable apart.

Rebased onto `de2a419`, so it sits on top of #355 rather than beside it. The two are complementary in `realtime.js`: #355 stops upstream error text reaching the client, and the `429` added here is generated locally and already generic, so nothing this PR emits carries upstream detail. All gates above were re-run after the rebase, not carried over from before it.


daikaginza:
Rebased onto `main` (the conflict was only my `CHANGELOG.md` entry meeting the newer ones; kept both), and pushed one follow-up commit — details below.

On the root cause, because it is shared with several of the other open reports rather than specific to these routes:

Every proxy in this family was written with care about the **upstream**. What none of them bounded was **the caller**. Here the guard was actually *written* — `makeOptInRateLimiter` exists, it is correct, and it is wired into both cost endpoints. It just defaulted to `null`, so on an unconfigured server the code path that enforces anything never ran. A guard that is off by default is indistinguishable, at runtime, from a guard that was never written; the only difference is that it reads as solved.

That is what makes this one worth calling out separately from the others: the other routes are missing a limiter, and this one *has* one and still had no limit.

Two details in the diff worth a reviewer's eye:

**The default is not invented.** `.env.example` already recommended 30 and the Pinokio build already ships `GEV_RATELIMIT_OPENAI_PER_MIN=30`, so 30 is the number this project had already chosen twice — the packaged app is unaffected and only an unconfigured server changes behaviour. `0` remains the documented, explicit way to run unthrottled.

**A typo must not disarm the guard.** `3O` (letter O), `60/min`, `thirty`, `-5` now fall back to the default rather than to unlimited. Reading an unparseable value as "no limit" is the opposite of what an operator was reaching for when they typed it.

Checked at the real door, not only in the suite — dev server, actual HTTP, no key configured:

```
GET  /api/realtime/token     -> 30 x 503 (keyless, as expected)
request #31                  -> 429 + Retry-After: 5
```

**And that probe found a gap in my own work, which is the follow-up commit.** 40 POSTs to `/api/openai/hud-summary` were all answered and none spent budget. My first read was that the limiter was missing on that route. It is not — with no key the request resolves on the keyless local path, which costs nothing upstream and is deliberately left unthrottled so an unconfigured install keeps its free HUD. Correct by design. But this PR edits `hud-summary.js` and the suite asserted nothing about it: it covered `/api/realtime/token` and the Google places route only. "Correct and untested" is one edit away from "wrong and untested", so I added a case that gates both halves — with a key, exactly the cap then 429 with `Retry-After` and no upstream call past it; with the key removed, the same exhausted client still gets its free answer instead of a 429.

Mutation-verified: removing the limiter from the route turns it red, and so does moving it *above* the keyless short-circuit.

Full gates: `npm test` 4190 tests / 4189 pass / 0 fail / 1 skip, 13 allocation microbenchmarks under `GEV_REQUIRE_ALLOCATION_GATE=1`; `npm run test:track` 109 / 0; doctor, format:check (919 files), check:boundaries, build all clean.

PR #685, opened 2026-09-21
fix(server): stop track backfill from spending the OpenSky credit budget

### LINKED ISSUE #259: OpenSky track-backfill proxies (/api/opensky-track, /api/adsblol/trace) bypass the shared credit/cooldown governor

daikaginza:
Fixes #259.

## Changes

`/api/opensky-track` spends 4 OpenSky credits per call against the same account as `/api/opensky` — the file's own doc comment says so. But it participates in none of the credit governor that proxy maintains, and its only throttle is a 60-second per-`icao24` result cache. That cache bounds **memory, not spend**: `icao24` is six hex digits, so a caller varying it produces a fresh key and a fresh live call every time.

The governor this route was missing is not something I invented. `server/providers/aircraft/opensky.js:28-39` records the incident that produced it:

> The global `/states/all` this proxy fetches costs 4 CREDITS per call against OpenSky's ~4000/day authenticated budget — a day with the app open burned the whole quota in ~8h and the layer then hard-died until the daily reset ("rate limited for 48h" owner report; auth itself was fine).

Three levers were added there. This route had none of them, while spending the same budget.

- **A per-client limiter, always on.** 30 requests/minute/client, 90 global, on both track routes. That is the band `/api/geocode` and `/api/regional-brief` already use, and it is sized against real demand rather than guessed: a backfill is issued once per aircraft the operator *selects* (`_backfillTrail`, reached from the tracked-contact path in the flights, military and vessels layers), and re-selecting the same aircraft inside a minute is served from cache. 30/min is a target switch every two seconds, sustained.
- **The cooldown is now shared.** `opensky.js` exports `openSkyCooldownRemainingMs()` — a read-only view of the existing `_openskyCooldownUntil`; only that module still writes it. While the account is cooling, `/api/opensky-track` does not mint a token and does not call upstream: it serves any cached track (lever 3, serve-stale — a cached body costs nothing further) and otherwise answers `429` carrying the remaining cooldown.
- **adsb.lol gets the limiter but not the governor.** It is free and unmetered, so there is no credit budget to protect; the limiter is there for the operator's IP reputation with a community service, the same reason `/api/overpass` carries one. Saying this explicitly because the two routes now look asymmetric on purpose.

Client behaviour is unchanged. `_backfillTrail` is fire-and-forget and already treats any non-OK response as "keep the locally accumulated trail" — a `429` is a path it has always had.

## How it was verified

**Six cases in `src/openSkyTrackGovernor.test.mjs`**, driving the real exported route handlers with `fetch` stubbed per test and restored after, so a case that reaches upstream fails loudly rather than making a request:

- 30 lookups with a **different `icao24` each time** pass; the 31st is refused with `429` + `Retry-After`, and upstream saw exactly 30 calls. This is the issue's core claim — that varying the key walks past the cache — asserted directly.
- A second client address still has its own quota, so one noisy host cannot take the layer away from everyone.
- `/api/adsblol/trace` is metered too.
- A malformed `icao24` is still rejected with `400` before any upstream call.
- `openSkyCooldownRemainingMs()` reads 0 until armed, and a past cooldown never reads as active.
- **End-to-end across both modules:** a stubbed `429` from OpenSky's `/states/all` arms the real governor through the real `/api/opensky` handler; a subsequent cold `/api/opensky-track` is then refused with the remaining cooldown and **zero** upstream calls.

**Mutation-checked rather than assumed green.** Disabling the limiter turns 3 of 6 red; keying the limiter globally instead of per-client turns 1 red; ignoring the cooldown turns 1 red. Reverting each returns 6/6.

**Gates** on Node 24.14.0 with `GEV_REQUIRE_ALLOCATION_GATE=1`: `npm run format:check` (919 files), `npm run check:boundaries`, `npm run build`, `npm test` (4167 tests, 0 failures, 1 pre-existing skip, all 13 allocation microbenchmarks ran), and `npm run test:track` (109 passed, 0 failed).

The limiter does not throttle the project's own harness: `test:track` is 109/109 with it on.

## Notes for review

- **`scripts/package-boundaries.json` needed two lines.** `check:boundaries` refused the first build because the `live-providers` package does not own `server/providers/common/rate-limit.js` or `src/sources/rateLimit.js`. I declared both, copying exactly what `place-providers` already declares. Flagging it because it widens a package's surface, which is your call rather than an implementation detail.
- **The issue's line references have gone stale.** It cites `vite.config.js:4956-5043`; that code now lives in `server/providers/aircraft/tracks.js`, and the governor it compares against is in `server/providers/aircraft/opensky.js`. The finding itself reproduced exactly as described.
- **What this does and does not bound.** A per-minute cap slows a determined abuser; it does not by itself make the daily credit budget unreachable. The cooldown is what actually protects the budget, because it engages the moment OpenSky says the account is out. I would rather state that plainly than imply the limiter alone closes the issue.
- **`globalMax` is 3× the per-IP cap here** (90), matching the always-on limiters in this repo rather than the 20× the opt-in helper uses.


daikaginza:

On the root cause, since it turned out to be shared with a few of the other open reports:

Each of these proxies was written with real care about the **upstream** — validate the target, cap the body, cache the answer, honour a cooldown. What none of them bounded was **the caller**. And the moment a cache key comes from the caller, the cache stops being protection and becomes amplification: every distinct key is a fresh miss, so the cache bounds memory while doing nothing at all about spend.

That is exactly this route's shape. `icao24` is six hex digits chosen by the request, the 60 s per-icao cache is keyed on it, and each miss costs 4 OpenSky credits against the same daily account `/api/opensky` is carefully protecting. The cooldown lever already existed a file away — it just wasn't visible to this module, so the route kept minting tokens and spending credits while the account was in the penalty box.

So the fix is two lines of intent: honour the existing cooldown (serve stale, never spend), and put a per-IP ceiling on frequency, which is the one thing the cache could never do.

Checked at the real door rather than only in the suite — dev server, actual HTTP:

```
bad icao24                     -> 400   (validation intact)
30 requests, one icao24        -> 200   (1 upstream call, 29 from cache)
request #31                    -> 429 + Retry-After: 10
/api/adsblol/trace right after -> 429   (shares the same 30/min budget)
```

Worth flagging that last line explicitly since the diff doesn't say it out loud: **both track routes share one 30/min budget per client**, deliberately — a backfill usually tries OpenSky and falls back to adsb.lol for the same aircraft, and I would rather the pair be metered together. Happy to split them into two budgets if you would prefer that.

Full gates on the rebased tree: `npm test` 4186 tests / 4185 pass / 0 fail / 1 skip, 13 allocation microbenchmarks under `GEV_REQUIRE_ALLOCATION_GATE=1`; `npm run test:track` 109 passed / 0 failed; doctor, format:check (919 files), check:boundaries, build all clean.

PR #686, opened 2026-09-21
fix(server): bound the terrain-heights cache and its request rate

### LINKED ISSUE #260: /api/terrain/heights: unbounded in-memory and on-disk cache growth, no rate limiter, no coordinate bounds validation

daikaginza:
Fixes #260.

## Changes

The shape of this endpoint is what makes it worth bounding: **every distinct 5-decimal point is its own cache key, its own permanent line in `.gev-cache/terrain-heights.json`, and its own upstream call.** Nothing limited the request rate, and coordinates were validated for being *finite* rather than for being *on Earth* — so `0,999` was a perfectly good cache key.

Four bounds, each following a pattern already in this repo rather than a new invention:

| Bound | Precedent followed |
| --- | --- |
| WGS-84 range rejected with `400` at parse | — (the gap itself) |
| Point cache ≤ 50,000 entries, oldest-first eviction | `trackBackfillProxies`'s `TRACK_CACHE_MAX` |
| Cache file above 24 MB not read back | `rocketLaunchesProxy`'s `maxDiskCacheBytes` |
| 90 requests/min/client, 270 global | `/api/overpass`, `/api/military-installations` |

**Both numbers are measured, not guessed.**

- The entry ceiling: this checkout's `.gev-cache/terrain-heights.json` held **6,189 entries in 1,039,666 bytes** after ordinary use — ~168 bytes/entry. 50,000 entries is eight times that working set and ~8 MB on disk, comfortably inside the 24 MB file ceiling.
- The rate cap: I instrumented the route and ran the full `npm run test:track` suite, the heaviest exercise this app has. It served **10 requests, peaking at 8/min**. 90/min is an order of magnitude above that.

Eviction is safe here in a way it would not be for a live feed: terrain does not move, so dropping an old point costs one refetch and never a wrong answer.

Range is enforced in `parseTerrainPoints` — at the edge, so no later stage has to repeat it. The poles and the antimeridian still parse; a batch containing one out-of-range point is rejected whole rather than silently trimmed.

## How it was verified

**Seven cases in `src/terrainHeightsBounds.test.mjs`**, driving the real exported route handler with `fetch` stubbed per test and restored after. Each of the four bounds is gated as *behaviour*, not asserted as a constant:

- Off-globe coordinates (`0,91`, `181,0`, `1e9,1e9`, `0,90.00001`) do not parse; the poles, the antimeridian and an ordinary point still do.
- An out-of-range point returns `400` with **zero** upstream calls, and one hidden inside a valid batch rejects that batch.
- **Eviction is proven by refetch:** with the ceiling set to 3, four distinct points cost four lookups; the newest is then served from cache at no cost, while the oldest has been evicted and costs exactly one refetch.
- **The disk ceiling is proven both ways:** the same cache file is skipped under a ceiling that excludes it (so the point is a miss) and *used* under one that admits it — which is what shows the first result came from the size guard rather than a broken path.
- The 91st request in a minute is refused, and a second client still has its own quota.

The ceilings are constructor options precisely so they can be driven like this; production constructs the proxy with no arguments.

**Mutation-checked.** Removing the range check reds 3 of 7; disabling eviction reds 1; removing the disk ceiling reds 1; disabling the limiter reds 2. Restoring each returns 7/7.

**Gates** on Node 24.14.0 with `GEV_REQUIRE_ALLOCATION_GATE=1`: `npm run format:check` (919 files), `npm run check:boundaries`, `npm run build`, `npm test` (4168 tests, 0 failures, 1 pre-existing skip, all 13 allocation microbenchmarks ran), `npm run test:track` (109 passed, 0 failed).

## Notes for review

Three things I want your eye on, two of which are changes to *your* test files that I would rather flag than bury.

- **`src/proxyErrorResponses.test.mjs` needed two lines.** That harness evaluates the extracted proxy source in a `new Function` scope, so a new import in the production file is an undefined identifier there. I added `makeRateLimiter` and `clientKey` to its `deps` — the real ones, so a fixture request is throttled exactly as production would be. The file is not in `format-scope.json`, so I edited it by hand and did not run the formatter over it.
- **The same harness dictated a signature.** It slices each proxy to the first closing brace in column zero, so a multi-line destructured parameter list cuts the function in half. The options are therefore destructured in the body, with a comment saying why — otherwise the next person will "tidy" it back and get a confusing `SyntaxError`.
- **One bug found by running the full suite, not by the new tests.** My first version called `fsp.stat` unconditionally before `readFile`. `environmentProviders.test.mjs` mocks `readFile` but not `stat`, so a missing file made `stat` throw and silently skipped the legacy-cache migration entirely. Fixed by treating "cannot measure it" as distinct from "too large": without stat, `readFile` decides, exactly as before. Worth knowing that path had no direct coverage of its own.
- **What the rate cap does and does not do.** 90/min still permits a lot of distinct points over a day. The entry ceiling is what actually bounds memory and disk; the limiter bounds the burst and the load on Re:Earth. I would rather say that than imply the limiter alone closes the growth issue.
- **The issue's line references have gone stale** (`vite.config.js:2242-2250` etc.); this code now lives in `server/providers/terrain.js` and `src/data/terrainHeightsProxy.js`. Every finding reproduced.


daikaginza:

On the root cause, because it turned out to be shared with several of the other open reports rather than specific to this route:

Every proxy in this family was written with real care about the **upstream** — validate the target, chunk the batch, cap the body, retry with `Retry-After`, keep the stale answer when the fresh one fails. What none of them bounded was **the caller**. And once a cache key comes from the caller, the cache stops being protection and becomes amplification: every distinct key is a guaranteed miss.

This route is the clearest case of it, because the key is a *coordinate*. `-97.74000,30.26000` and `-97.74001,30.26000` are two different keys, two different cache lines, two different upstream calls — and there are as many of them as a caller cares to type. The per-point cache is genuinely good engineering for the real client, which asks about the same ground repeatedly; it just has no opinion about a caller that never asks twice.

The missing coordinate validation is the same blind spot one level down: the parser checked that each number was *finite*, which is not the same as being *on Earth*. `0,1e6` was a perfectly good cache key before this change.

Two notes on the diff:

**The ceilings are measured, not picked.** This checkout's own `.gev-cache/terrain-heights.json` held 6,189 entries in 1.0 MB after ordinary use (~168 B/entry), so 50,000 is eight times a real working set and ~8 MB — inside the 24 MB file ceiling `rocketLaunchesProxy` already applies. For the limiter, a full `npm run test:track` run — the heaviest exercise this app has — drove 10 requests peaking at 8/min, so 90/min is an order of magnitude of headroom.

**`evictOverflow()` sorts by fetch time rather than evicting in insertion order.** That is deliberate: entries are refreshed in place, so Map insertion order stops tracking age. The cost is an O(n log n) pass on overflow instead of O(1) — bounded in practice because the limiter caps the route at 90 requests/minute, but I am happy to switch to delete-then-reinsert (which restores O(1)) if you would rather have that.

Checked at the real door rather than only in the suite — dev server, actual HTTP. Worth noting the wire format is `lon,lat`, which I got backwards on my first probe and had to correct:

```
points=-97.74,30.26   -> 200 {"lon":-97.74,"lat":30.26,"elevation":139.08,...}
points=0,90           -> 200   (the pole is a legal latitude)
points=0,90.00001     -> 400
points=-181,0         -> 400
points=0,-91          -> 400
90 requests           -> served
request #91           -> 429 + Retry-After: 10
```

Full gates on the rebased tree: `npm test` 4187 tests / 0 fail, 13 allocation microbenchmarks under `GEV_REQUIRE_ALLOCATION_GATE=1`; `npm run test:track` 109 / 0; doctor, format:check, check:boundaries, build all clean.

One process note in case it is ever useful to you: my first rebased `test:track` reported 4 failures, all of them `HTTP 504 (Outdated Optimize Dep)` on `node_modules/.vite/deps/egm96-universal.js`. That was my own harness carrying one branch's Vite pre-bundle into the next checkout, not anything in this diff — with a cold `node_modules/.vite` it is 109/0. Mentioning it because the failure text looks alarming and has nothing to do with the code.

PR #687, opened 2026-09-21
fix(terrain): cap the browser-side terrain height cache

### LINKED ISSUE #327: Unbounded In-Memory Cache Growth in `terrainHeights.js` (Memory Leak)
# 🚨 Engineering Incident & Pull Request Document

<div align="center">

# FIX: Unbounded In-Memory Cache Growth in `terrainHeights.js` (Memory Leak)
**Repository:** `bilawalsidhu/gods-eye-view` · **Subsystem:** Geospatial Height Datum & Terrain Engine · **Target File:** `src/data/terrainHeights.js`

[Severity: High](src/data/terrainHeights.js)
[Subsystem: Height Datum](src/data/terrainHeights.js)
[Status: PR Ready](https://github.com/bilawalsidhu/gods-eye-view)
[Coverage: Verified](src/data/terrainHeights.test.mjs)

</div>

---

## 📋 Executive Overview

| Metric | Specification |
| :--- | :--- |
| **Bug Identifier** | `GEV-BUG-002` (Critical Memory Leak during Long-Running Tracking Sessions) |
| **Component Impacted** | Client-side Ellipsoidal Elevation Resolver (`resolveEllipsoidalGround`) |
| **Primary Code Targets** | `src/data/terrainHeights.js:L47-L88`, `src/data/terrainHeights.js:L213-L232` |
| **Operational Impact** | Monotonic V8 heap expansion, progressive FPS drops, browser tab freezing / out-of-memory crash |
| **Resolution Type** | O(1) Least-Recently-Used (LRU) Pruning & Capacity Ceiling Insertion |

---

## ⚡ 1. Problem Statement

> [!WARNING]
> **Operator Hazard:** In surveillance consoles left running for several hours (or continuously tracking dense air traffic), the browser tab consumes over 1.2 GB of heap memory before freezing or crashing due to unbounded in-memory coordinate caching.

```
                    CONTINUOUS TRACKING OVER 2-6 HOURS
  11,000+ Live Flights + AIS Vessels             Unbounded V8 Heap Growth
 ┌──────────────────────────────────────┐      ┌─────────────────────────┐
 │ Coordinates sampled every 15-30s     │      │ 400MB                   │
 │                                      │ ───► │ 850MB                   │
 │  40.71278, -74.00594 → 40.71302...   │      │ 1.4GB ──► [OOM CRASH!]  │
 │  (Tens of thousands of new keys)     │      │ (No Eviction / Pruning) │
 └──────────────────────────────────────┘      └─────────────────────────┘
```

### Manifestation & Symptoms
1. **Unbounded Memory Leak:** When tracking live global aircraft (11,000+ contacts) or maritime vessels over multiple polling cycles, each entity position creates a rounded 5-decimal key string (e.g. `"30.19452,-97.66987"`). These keys and their resolved metadata objects are inserted into an internal `Map` that is **never purged**.
2. **GC Thrashing & Frame Stutter:** As the map size exceeds 50,000–100,000 keys, V8 Major Garbage Collection cycles take over 120ms, causing noticeable frame stuttering on Cesium globe camera movements.
3. **Architectural Contrast:** Other streaming modules in the codebase enforce strict capacity ceilings (e.g. `routes.js` bounds route cache to 200 items, and `cctvCards.js` implements explicit `planFrameCachePrune`). `terrainHeights.js` completely omits a cache limit.

---

### Flaw: Indefinite Accumulation in Module-Scoped Map
In `src/data/terrainHeights.js:L47-L58`:

```javascript
// src/data/terrainHeights.js:47
/**
 * In-memory cache: "lat.toFixed(5),lon.toFixed(5)" -> {ellipsoid, source}.
 * Module-scoped (not exported) — the only reads are through
 * cachedEllipsoidalGround and the internal lookup in resolveEllipsoidalGround.
 * @type {Map<string, {ellipsoid: number, source: 'reearth'|'geoid-fallback', retryAt?: number}>}
 */
const cache = new Map();
```

During resolution in `resolveEllipsoidalGround`:
```javascript
// src/data/terrainHeights.js:213-231
if (Number.isFinite(ellipsoid)) {
  cache.set(item.key, { ellipsoid, source: 'reearth' });
}
...
cache.set(item.key, {
  ellipsoid,
  source: 'geoid-fallback',
  retryAt: Date.now() + GEOID_FALLBACK_COOLDOWN_MS,
});
```

- Every unique 5-decimal coordinate visited by any tracked object or camera frustum is permanently retained.
- There is **no eviction policy, no time-to-live (TTL) expiration, and no size check**.
- Because the module is long-lived across the entire browser lifecycle, the `Map` instance lives indefinitely in the root scope.

---

### Pull Request Metadata
```yaml
PR Title: "fix(terrain): implement O(1) LRU capacity bound for terrain height cache"
Branch: fix/terrain-heights-lru-cache
Base: main
Type: Bug Fix / Memory Optimization
Performance: Clamps cache memory to ≤ 10,000 entries (~3.2 MB heap footprint)
```

### Complete Code Implementation (Unified Diff)

```diff
--- a/src/data/terrainHeights.js
+++ b/src/data/terrainHeights.js
@@ -36,6 +36,12 @@ import { ensureGeoidReady, geoidHeight } from './geoid.js';
 const CHUNK_SIZE = 200;
 
+/**
+ * Maximum entries retained in the in-memory terrain height cache.
+ * 10,000 entries provide extensive spatial reuse while capping memory footprint under 4 MB.
+ */
+const TERRAIN_CACHE_CAPACITY = 10_000;
+
 /** Avoid repeatedly hitting a known-failing proxy from warm fallback reads. */
 const GEOID_FALLBACK_COOLDOWN_MS = 60_000;
 
@@ -48,6 +54,23 @@
 const cache = new Map();
 
+/**
+ * Sets a cache entry adhering to strict LRU semantics.
+ * If capacity is exceeded, evicts the oldest (least recently accessed) entry.
+ * @param {string} key
+ * @param {{ellipsoid: number, source: 'reearth'|'geoid-fallback', retryAt?: number}} value
+ */
+function setCacheEntry(key, value) {
+  if (cache.has(key)) {
+    cache.delete(key);
+  } else if (cache.size >= TERRAIN_CACHE_CAPACITY) {
+    const oldestKey = cache.keys().next().value;
+    cache.delete(oldestKey);
+  }
+  cache.set(key, value);
+}
+
 /**
  * Builds the rounded cache key shared between the in-memory cache and the
  * `points` string sent to the proxy (see file header, point 2).
@@ -69,6 +92,11 @@
 export function cachedEllipsoidalGround(lat, lon) {
-  const entry = cache.get(cacheKey(lat, lon));
+  const key = cacheKey(lat, lon);
+  const entry = cache.get(key);
+  if (entry) {
+    // Refresh recency in LRU order
+    cache.delete(key);
+    cache.set(key, entry);
+  }
   return entry ? entry.ellipsoid : null;
 }
 
@@ -86,6 +114,11 @@
 export function cachedRealEllipsoidalGround(lat, lon) {
-  const entry = cache.get(cacheKey(lat, lon));
+  const key = cacheKey(lat, lon);
+  const entry = cache.get(key);
+  if (entry && entry.source === 'reearth') {
+    cache.delete(key);
+    cache.set(key, entry);
+    return entry.ellipsoid;
+  }
   return null;
 }
@@ -213,7 +246,7 @@ export async function resolveEllipsoidalGround(coords) {
         if (Number.isFinite(ellipsoid)) {
-          cache.set(item.key, { ellipsoid, source: 'reearth' });
+          setCacheEntry(item.key, { ellipsoid, source: 'reearth' });
         }
       }
     } catch {
@@ -225,7 +258,7 @@ export async function resolveEllipsoidalGround(coords) {
         const ellipsoid = geoidFallback(item.lat, item.lon, item.sourceOrthometricM);
-        cache.set(item.key, {
+        setCacheEntry(item.key, {
           ellipsoid,
           source: 'geoid-fallback',
           retryAt: Date.now() + GEOID_FALLBACK_COOLDOWN_MS,
         });
```

---

### Test Suite Execution
```bash
# 1. Run unit test suite
node scripts/run-unit-tests.mjs

# 2. Verify all terrain datum tests pass
node --test src/data/terrainHeights.test.mjs
```

### Automated LRU Unit Test (`src/data/terrainHeights.lru.test.mjs`)
```javascript
import { test } from 'node:test';
import assert from 'node:assert/strict';
import { resolveEllipsoidalGround, cachedEllipsoidalGround } from './terrainHeights.js';

test('terrainHeights cache evicts oldest entries when exceeding capacity', async () => {
  // Generate 10,005 coordinate points
  const points = [];
  for (let i = 0; i < 10005; i++) {
    points.push({
      lat: 10 + (i * 0.001),
      lon: 20 + (i * 0.001),
      sourceOrthometricM: 50
    });
  }

  // Resolve batch
  await resolveEllipsoidalGround(points);

  // The first point (index 0) should have been evicted to make room for point 10,004
  const evicted = cachedEllipsoidalGround(10.0, 20.0);
  assert.equal(evicted, null, 'Oldest point must be evicted when capacity is reached');

  // The latest point should remain warm in cache
  const latest = cachedEllipsoidalGround(10 + (10004 * 0.001), 20 + (10004 * 0.001));
  assert.notEqual(latest, null, 'Latest point must exist in cache');
});
```

> [!TIP]
> **Performance Benchmark:** Under continuous simulation of 500,000 coordinate resolutions, memory consumption remains strictly plateaued at **~3.5 MB**, compared to **~180 MB** on unpatched `main`.

daikaginza:
## Changes

`createTerrainHeights` (`src/services/terrainHeights.js`) keeps a `Map` keyed by the 5-decimal `"lat,lon"` of every point it resolves. Nothing ever removed a key, so the map grew for as long as the tab stayed open.

- New `cacheMaxEntries` option, defaulting to **50,000**, with oldest-first eviction.
- Trimming runs **once per batch, after the response has been assembled** — never from the write itself.
- New behaviour gate `src/terrainHeightsClientCache.test.mjs` (4 cases), registered in `scripts/format-scope.json`.
- `docs/CURRENT-STATE.md` + `CHANGELOG.md` updated, per CONTRIBUTING's rule for runtime behaviour changes.

### Why 50,000 and not a guess

The server side of this exact traffic is measurable: `.gev-cache/terrain-heights.json` receives precisely the points this module asks for, and it held **6,189 entries** after ordinary use. 50,000 is eight times that working set — about 5 MB at roughly 100 bytes per entry — and matches the ceiling the proxy applies to its own copy. A normal session is never evicted; a session that pans for hours is bounded.

### Why the trim must come after assembly

`resolveEllipsoidalGround` builds its answer by reading each key back out of the cache. Evicting on write turns points the batch had just resolved into `{ellipsoid: null, source: 'unresolved'}` whenever a single batch exceeds the ceiling — a silent loss of heights, not merely a slower path. I hit exactly this while writing the fix; the third test in the new file is the gate that caught it and keeps it caught.

## How it was verified

- `npm test` — 4165 tests, **4164 pass / 0 fail / 1 skip**, including the 13 allocation microbenchmarks under `GEV_REQUIRE_ALLOCATION_GATE=1`.
- Mutation-verified, not source-scanned: removing `pruneCache()` turns one test red; moving it inside the write path turns a different one red; restoring gives 4/4.

## Notes for review

Three corrections to the issue text, offered as findings rather than objections:

1. **The cited location is not the code.** #327 points at `src/data/terrainHeights.js:L47-L88`; that file is an 11-line re-export shim with no `Map` in it. The cache actually lives in `src/services/terrainHeights.js:55`.
2. **The 1.2 GB / OOM figure is not reproducible.** I instrumented a real browser session (headless Chromium, the same flags `scripts/track-regression.mjs` uses) across five cities: the client cache held **0 entries** and heap peaked at **441 MB** — the terrain path simply did not populate under that scenario. At ~100 bytes per entry, even 100k keys is ~10 MB, so this map is very unlikely to be anyone's 1.2 GB. I am fixing it because unbounded-by-construction is still worth closing, not because I reproduced the stated severity.
3. **The fix sketched in the issue (evict on insert) is unsafe here** for the assembly reason above.

This is the client-side half. The server-side map in `server/providers/terrain.js` is #260, handled separately in #686.

Fixes #327


daikaginza:

On the root cause, because it is shared with several of the other open reports:

Every one of these caches was written with care about the **upstream** — batch the points, resolve each one once, never ask twice for ground that does not move. What none of them bounded was **growth**, because the key comes from the session rather than from a fixed set. `-97.74000,30.26000` and `-97.74001,30.26000` are two entries. A map is exactly the right structure for "never ask twice"; it is just not a structure that has an opinion about when to stop.

The client half is the same shape as the server half in #260, one process further out.

Two things I would flag honestly, because they cut against the issue as filed:

**The cited location is not the code.** The issue points at `src/data/terrainHeights.js:L47-L88`; that file is an 11-line re-export shim with no `Map` in it. The cache is in `src/services/terrainHeights.js:55`. I mention it only so the next reader does not lose the time I did.

**The 1.2 GB / OOM figure did not reproduce.** I instrumented a real browser session — headless Chromium with the same flags `scripts/track-regression.mjs` uses — across five cities: the client cache held **0 entries** and heap peaked at **441 MB**; that scenario never drove the terrain path at all. At roughly 100 bytes per entry even 100k keys is ~10 MB, so this map is unlikely to be anyone's 1.2 GB. I am proposing the fix because unbounded-by-construction is worth closing on its own, not because I reproduced the stated severity — and I would rather say that plainly than let a number stand that I could not verify.

**The fix the issue sketches is unsafe here, and my own first attempt had the bug.** `resolveEllipsoidalGround` assembles its answer by reading each key back out of the cache. Evicting on write therefore turns points the batch has *just resolved* into `{ellipsoid: null, source: 'unresolved'}` whenever a single batch exceeds the ceiling — a silent loss of heights, not a slower path. Trimming has to happen once per batch, after assembly. The third case in the new test file is the gate that caught it and keeps it caught.

The ceiling (50,000) is taken from the server side of the same traffic: `.gev-cache/terrain-heights.json` receives exactly the points this module asks for and held 6,189 entries after ordinary use, so it is eight times a real working set and matches what the proxy applies to its own copy.

Full gates on the rebased tree: `npm test` 4184 tests / 0 fail, 13 allocation microbenchmarks under `GEV_REQUIRE_ALLOCATION_GATE=1`; `npm run test:track` 109 / 0; doctor, format:check, check:boundaries, build all clean. Mutation-verified rather than source-scanned: remove `pruneCache()` → 1 red; move it into the write path → a different 1 red; restore → 4/4.

PR #689, opened 2026-09-21
fix(adsbdb): rate-limit /api/adsbdb and cap its enrichment caches

### LINKED ISSUE #261: /api/adsbdb: no rate limiter, unbounded persisted cache via keyspace enumeration

daikaginza:
## Changes

`/api/adsbdb` accepts two caller-supplied keyspaces — a callsign is 2-8 of `[A-Z0-9]`, a hex is six of `[0-9a-f]` — and cached every distinct key that reached upstream, 404s included, with nothing ever removing one.

**Rate limit.** 360 requests/minute per client address, global backstop 3×, refused with `429` + `Retry-After` **before any cache, disk or upstream work**. The number is not a guess: the browser cannot exceed 300/min by construction, because `ENRICH_DISPATCH_GAP_MS` (`src/layers/flights/policy.js:226`) is 200 ms and the enrichment queue drips at most five lookups a second. 360 is that unreachable ceiling plus headroom.

**Cache ceiling.** 20,000 entries per store, dropping least-recently-written first; entries past the existing 24-hour TTL are discarded when the file is read back, so a restart no longer carries dead keys forward. A cache file above 16 MB is skipped rather than parsed. Sized from a real session's own `.gev-cache/adsbdb.json`: **397 aircraft entries (~100 B each) and 3 route entries (~231 B each)** — so the ceiling is roughly fifty times an ordinary working set and bounds the file at about 7 MB.

**Maps instead of plain objects.** `size` is O(1), so the ceiling can be enforced on every write rather than on a timer, and iteration is true insertion order. The object form was subtly wrong for this: an all-digit callsign like `"1234"` is an integer-like key and JavaScript sorts those ahead of every string key, so oldest-first eviction over an object would have evicted the wrong entries. The on-disk JSON shape is unchanged.

Also: `docs/CURRENT-STATE.md` and `CHANGELOG.md` per CONTRIBUTING's runtime-behaviour rule; `src/adsbdbCacheGovernor.test.mjs` registered in `scripts/format-scope.json`; `server/providers/common/rate-limit.js` + `src/sources/rateLimit.js` declared on the `live-providers` module in `scripts/package-boundaries.json`.

## How it was verified

- `npm test` — **4186 tests, 4185 pass / 0 fail / 1 skip**, including the 13 allocation microbenchmarks under `GEV_REQUIRE_ALLOCATION_GATE=1`.
- New behaviour gate `src/adsbdbCacheGovernor.test.mjs` (5 cases) drives the real exported route handler with `fetch` stubbed per test and restored after, so a case that reached the network would fail loudly.
- Mutation-verified rather than source-scanned:
  - neuter `prune()` → 1 red
  - make the limiter always admit → 2 red
  - collapse the route and aircraft stores into one → 1 red
  - restore → 5/5 green

The limiter test asserts both directions: at least 300 requests are served before refusal (a real session must never be throttled), and the number of upstream calls equals the number of served requests (a refused request never reaches api.adsbdb.com).

## Notes for review

- The issue cites `vite.config.js:2439-2499`; that code now lives in `server/providers/aircraft/enrichment.js`. Same handler, same `lookup()`.
- `.gev-cache/adsbdb.json` written by an older build is still read correctly — `toStore()` just drops what has expired.
- This touches two lines that #685 and #686 also touch (the `live-providers` module list, and the `deps` map in `src/proxyErrorResponses.test.mjs`). Whichever lands first, the rebase is a one-line take-both; happy to rebase on request.

Fixes #261


daikaginza:
On the root cause, because it turned out to be shared with a few of the other open reports rather than specific to this route:

These proxies were each written with real care about the **upstream** — validate the target, cap the body, cache the answer, negative-cache the misses. What none of them bounded was **the caller**. And once a cache key comes from the caller, the cache stops being protection and becomes amplification: every distinct key is a guaranteed miss, so the cache bounds nothing that matters. Here it is doubly so, because the miss is also *written down* — a permanent line in `.gev-cache/adsbdb.json`.

The "cached aggressively" comment at the top of this module is honest about the intent and is exactly what made the gap easy to miss: it reads like a protection, and for the real client it is. It only inverts when the key is attacker-chosen, and `^[A-Z0-9]{2,8}$` is a very large space to be remembering forever.

Two things I would call out in the diff:

**The stores became `Map`s, and that is load-bearing, not tidying.** `size` is O(1), so the ceiling can be enforced on every write instead of on a timer. More importantly, iteration order: an all-digit callsign like `"1234"` is an integer-like object key, and JavaScript sorts those *ahead* of every string key. Oldest-first eviction over a plain object would therefore have evicted the wrong entries — quietly, and only for numeric callsigns. The on-disk JSON shape is unchanged either way.

**The ceiling is measured, not picked.** A real session's own `.gev-cache/adsbdb.json` held 397 aircraft entries (~100 B each) and 3 route entries (~231 B each). 20,000 per store is about fifty times that, and bounds the file near 7 MB.

Checked at the real door, not only in the suite — dev server, actual HTTP:

```
cold  /api/adsbdb/type/abc123 -> 200   (one call to api.adsbdb.com)
warm  same key                -> 200   in 0.5 ms, no upstream
360 requests                  -> 200
request #361                  -> 429 + Retry-After: 10
.gev-cache/adsbdb.json        -> written, 404 negative-cached as expected
```

The 360/min ceiling is above anything the browser can produce: `ENRICH_DISPATCH_GAP_MS` is 200 ms, so the enrichment queue drips at most five lookups a second.

Full gates: `npm test` 4186 / 4185 pass / 0 fail / 1 skip with the 13 allocation microbenchmarks under `GEV_REQUIRE_ALLOCATION_GATE=1`; `npm run test:track` 109 / 0; doctor, format:check (919 files), check:boundaries, build all clean.

Issue #31, opened 2026-08-25
Enforce GBFS response limits while streaming bytes


jsawyerdev:
Reviewed commit: 880a672b5e16ad3e41d318801d3a5203f9201923

Problem:
- The GBFS proxy reads the entire response body before enforcing the configured maximum.
- A missing or false `Content-Length` can still force memory growth before the cap is checked.

Evidence:
- `vite.config.js:3317-3330` checks `Content-Length`, then calls `await upstream.text()`.
- The size check happens after the whole decoded body has been materialized.

Suggested fix:
- Stream the response and count bytes as they arrive.
- Abort once the configured maximum is exceeded.
- Preserve the existing content-type validation before parsing.

This was found during an adversarial repository review; impact depends on how the dev/preview server is exposed.

dlucas489:
Took this one, along with #30, as a single gbfs hardening #180

Lob26:
Verified against `main` at `a8125ad`: this looks already fixed.

`server/providers/gbfs.js:14` declares the cap and the fetch path reads through the shared streaming reader:

```js
export const GBFS_MAX_BODY_BYTES = 5 * 1024 * 1024; // 5 MB
...
const body = await readResponseTextCapped(res, maxBytes, ...);   // :85
```

`readResponseTextCapped` (`src/sources/httpBody.js:6`) rejects an oversized declared `Content-Length` up front, keeps a running total while streaming, and cancels the body on the first chunk past the cap — which is the "while streaming" part of the request rather than a post-hoc length check. The cap is also injectable per call (`options.maxBytes`, `:43`). Closeable.

bilawalsidhu:
Thanks for the report. Checked against public main `2746ec7`: Current GBFS reader enforces a streaming response cap; focused tests pass.

[Current implementation](https://github.com/bilawalsidhu/gods-eye-view/blob/2746ec753ce4d6ebafee881a212c3f327907c673/src/data/gbfsProxy.test.mjs). Closing the original report as resolved/superseded. If the same behavior reproduces on current main, please reopen with the commit, environment and reproduction.

dlucas489:
@Lob26 @bilawalsidhu this was me! 😁 #180. Sorry. I didn't repeat keyword "closes" after 31. Just after 30. That's why it didn't auto-close. Sorry!

Issue #30, opened 2026-08-25
Disable or re-validate redirects in the GBFS proxy


jsawyerdev:
Reviewed commit: 880a672b5e16ad3e41d318801d3a5203f9201923

Problem:
- The GBFS proxy validates the original URL host/path, then uses default `fetch()` redirect behavior.
- If an allowed GBFS host redirects elsewhere, the final destination is not revalidated.

Evidence:
- `vite.config.js:3194-3207` defines the allowed GBFS host rules.
- `vite.config.js:3215-3217` validates the original GBFS path.
- `vite.config.js:3245-3349` performs the proxy fetch without `redirect: "manual"`.

Suggested fix:
- Set `redirect: "manual"` and reject redirects, or re-run the full allowlist validation on every redirect target.
- Add a redirect-specific regression test.

This was found during an adversarial repository review; impact depends on how the dev/preview server is exposed.

dlucas489:
Took this, along with #31, as a single gbfs hardening #180

PR #690, opened 2026-09-21
fix(gbfs): rate-limit /api/gbfs

### LINKED ISSUE #264: GBFS proxy (/api/gbfs) has no rate limiter

daikaginza:
## Changes

`/api/gbfs` already restricted **where** it would go: https only, an allowlisted host, a `station_information`/`station_status` path, `redirect: 'manual'`, a 12-second deadline and a 5 MB streaming body cap. Nothing restricted **how often**, so the dev server would relay an unbounded stream of requests at a third-party bikeshare operator on any caller's behalf — one outbound fetch and one timeout timer each.

- 120 requests/minute per client address, global backstop 3×, refused with `429` + `Retry-After` + `Cache-Control: no-store` **before the target is even parsed**.
- The existing method, host, path, redirect, deadline and size guards are untouched, and still run in the same order relative to each other.
- New behaviour gate `src/data/gbfsProxyRateLimit.test.mjs` (4 cases), registered in `scripts/format-scope.json`; `server/providers/common/rate-limit.js` + `src/sources/rateLimit.js` declared on the `gbfs-provider` module.
- `docs/CURRENT-STATE.md` + `CHANGELOG.md` per CONTRIBUTING's runtime-behaviour rule.

### Why 120

Taken from the layer's own demand rather than picked: `src/layers/bikeshare/registry.js` catalogues **15 systems**, station information is cached per city after its first fetch (`_stationInfoCache`), and station status refreshes once per city every `STATUS_POLL_MS` (60 s) with in-flight requests deduplicated. Even with every catalogued system in camera range simultaneously that is about **15 requests a minute**, so 120 is eight times a maximal real session. The first test asserts that directly: 120 consecutive requests must all be served.

## How it was verified

- `npm test` — **4185 tests, 4184 pass / 0 fail / 1 skip**, including the 13 allocation microbenchmarks under `GEV_REQUIRE_ALLOCATION_GATE=1`.
- Mutation-verified rather than source-scanned:
  - make the limiter always admit → 2 red
  - key the limiter globally instead of per client → 1 red
  - move the limiter after `fetchGbfsUpstream` → 1 red (a refusal would then still cost an outbound fetch)
  - restore → 4/4 green

## Notes for review

- The issue cites `vite.config.js:3348-3454`; that code now lives in `server/providers/gbfs.js`. Same handler.
- `ratePerMinute` is an option only so the ceiling can be exercised as behaviour; production uses the exported `GBFS_RATE_PER_MIN`.
- This is the frequency gap only, and is independent of #30 (redirect revalidation) and #31 (streaming byte enforcement) — both of which the current code already handles.
- Touches the `scripts/package-boundaries.json` file that #685/#686/#689 also touch, but a different module entry, so it should merge cleanly alongside them.

Fixes #264


daikaginza:
On the root cause, since it is shared with a few of the other open reports rather than specific to this route:

This proxy is actually the *most* carefully guarded of the family. It already answers **where**: https only, an allowlisted host, a `station_information`/`station_status` path, `redirect: 'manual'` so an allowlisted operator cannot steer it somewhere the allowlist never saw, a 12-second deadline that includes reading the body, and a 5 MB cap enforced while the body streams. Reading it, it is easy to come away satisfied — I did, on the first pass.

What none of it answers is **how often**. Every one of those guards is a property of a single request; not one of them is a property of a *sequence* of requests. So the dev server would relay an unlimited number of perfectly valid, perfectly allowlisted requests at a third-party bikeshare operator on any caller's behalf — one outbound fetch and one timeout timer each. The failure mode is not a breach, it is the operator's own IP getting rate-limited or blocked by a GBFS provider for traffic they never made.

That is the same shape as the other reports in this group: careful about the upstream, silent about the caller.

The ceiling is taken from the layer's own demand rather than picked: `src/layers/bikeshare/registry.js` catalogues 15 systems, station information is cached per city after its first fetch, and station status refreshes once per city every `STATUS_POLL_MS` (60 s) with in-flight requests deduplicated. Even with every system in camera range at once that is ~15 requests a minute, so 120 is eight times a maximal session.

Checked at the real door rather than only in the suite. Because the limiter runs *before* the host check, an off-allowlist target proves the limit through real HTTP with **zero outbound traffic** — no GBFS operator was touched to produce this:

```
requests 1-120 -> 403  (host refused, as before)
request  121   -> 429 + Retry-After: 10, Cache-Control: no-store
histogram      -> 403=120  429=15
```

`ratePerMinute` is an option purely so the ceiling can be exercised as behaviour; production reads the exported `GBFS_RATE_PER_MIN`.

Full gates: `npm test` 4185 / 4184 pass / 0 fail / 1 skip with the 13 allocation microbenchmarks under `GEV_REQUIRE_ALLOCATION_GATE=1`; `npm run test:track` 109 / 0; doctor, format:check (919 files), check:boundaries, build all clean. Mutation-verified: always-admit → 2 red, global key instead of per-client → 1 red, limiter moved after `fetchGbfsUpstream` → 1 red.

